PR #18, opened 2026-05-12
feat(workspace): hot-reload plugin agent layer (rebased)

hachej:
## Summary

Rebased replacement for stale PR #10. No force-push was used.

- Adds hot-reloadable boring-ui plugin assets and /reload integration.
- Preserves macro live plugin editing and mounted pane hot-swap fixes.
- Keeps old branch recoverable at feat/plugin-agent-layer and backup ref stale/plugin-agent-layer-pre-rebase.

## Verification

- pnpm --filter @hachej/boring-workspace typecheck
- pnpm --filter @boring/macro exec tsc --noEmit
- pnpm --filter @hachej/boring-workspace exec vitest run src/front/registry/__tests__/registry.test.tsx src/front/dock/__tests__/dock.test.tsx src/front/agentPlugins/__tests__/registerAgentPlugin.test.tsx src/shared/plugins/__tests__/frontFactory.test.tsx src/server/__tests__/agentPlugins.test.ts src/server/__tests__/createWorkspaceAgentServer.test.ts src/server/__tests__/workspaceContextPrompt.test.ts src/plugins/filesystemPlugin/front/__tests__/filesystemPlugin.test.ts

## Links

Closes #32

PR #134, opened 2026-05-29
fix(agent): stop CLI-mode slow workspace load from bin-symlink realpath probe

hachej:
## Problem

Opening a workspace in **CLI/local (bwrap) mode** (e.g. `boring-ui workspaces --mode local`) is slow — the UI sits on the *"Waking sandbox… / Preparing files…"* overlay for a long time. Static assets are instant (HTML 7ms, JS bundle 40ms); all the wall-clock goes into backend **runtime provisioning**, which was **crash-looping**.

Service logs for the affected workspace showed provisioning throwing repeatedly with retry/backoff (2s → 3s → 5s):

```
[DIAG path-escape] {"workspaceRoot":".../boring-ui-factory",
 "absPath":".../.boring-agent/node/node_modules/.bin/boring-ui",
 "realCandidate":"/home/ubuntu/.npm-global/.../@hachej/boring-ui-cli/dist/index.js"}
  at assertRealPathWithinWorkspace → shouldInstallNodeRuntime → ensureNodeRuntime → provisionWorkspaceRuntime
```

## Root cause

`shouldInstallNodeRuntime` probes whether expected node-runtime outputs already exist (to **skip reinstalling**). One expected output is the bin shim `.boring-agent/node/node_modules/.bin/boring-ui`, which npm creates pointing at the globally-linked `@hachej/boring-ui-cli` — so `realpath` resolves it **outside** the workspace root.

`workspaceFs.exists()` ran `assertRealPathWithinWorkspace` (follows the final symlink) and threw `symlink-escape` (a non-`ENOENT` error, re-thrown rather than treated as "missing"). Consequences:

1. The fingerprint skip-vs-reinstall short-circuit could **never** succeed.
2. The throw aborted the provisioning phase → **retry-with-backoff loop** on every boot.
3. The frontend kept polling `/api/v1/ready-status` every 500ms behind the boot overlay.

Direct mode already needed a special-case (`enforceSymlinkBoundary: false`) to dodge this exact false-positive; local/vercel-sandbox modes kept the strict check and hit the bug.

## Fix

Make `exists()` a **presence-only** check: `lstat` the entry itself (never follow the final symlink). A boolean "is something here" exposes no out-of-sandbox content, the lexical `validatePath()` still rejects `../` escapes, and content operations (`readText` / `rm` / `writeText`) keep the strict realpath boundary. This also unifies local/direct behavior and removes the need for the direct-mode special-case rationale.

## Verification

- New regression test: local adapter `exists()` on a workspace-internal symlink resolving outside the workspace returns `true` (and still returns `false` for missing, still rejects `../`).
- Full `provisioning` + `runtime/modes` suites: **68 passed**, no type errors, invariants pass.
- Validated against the **real** factory workspace symlink: old realpath check throws `symlink-escape`; new lstat probe returns `true` → fingerprint short-circuit can skip reinstall.

### DIFF
--- a/packages/agent/src/server/runtime/modes/__tests__/provisioningAdapter.test.ts
+++ b/packages/agent/src/server/runtime/modes/__tests__/provisioningAdapter.test.ts
@@ -103,6 +103,50 @@ test('direct adapter resolveInstallSource returns runtime-visible local paths an
   expect(adapter.getRuntimeCacheRoot()).toBe(paths.cache)
 })
 
+test('local adapter exists() treats an out-of-workspace bin symlink as present (no realpath-escape throw)', async () => {
+  // Repro of the CLI-mode slow-load bug: provisioning's skip-vs-reinstall
+  // probe (shouldInstallNodeRuntime) calls workspaceFs.exists() on expected
+  // outputs, one of which is .bin/boring-ui — an npm-created shim that
+  // realpath-resolves to the host's global @hachej/boring-ui-cli install,
+  // outside the workspace. In local (bwrap) mode the realpath guard used to
+  // throw symlink-escape here, breaking the fingerprint short-circuit and
+  // crash-looping provisioning on every boot. exists() must report present.
+  const workspaceRoot = await tempRoot('boring-local-exists-')
+  const outsideRoot = await tempRoot('boring-local-outside-')
+  await writeFile(join(outsideRoot, 'cli.js'), 'module.exports = {}\n')
+  await mkdir(join(workspaceRoot, '.boring-agent', 'node', 'node_modules', '.bin'), { recursive: true })
+  await symlink(
+    join(outsideRoot, 'cli.js'),
+    join(workspaceRoot, '.boring-agent', 'node', 'node_modules', '.bin', 'boring-ui'),
+  )
+  const adapter = createLocalProvisioningAdapter(getBoringAgentRuntimePaths(workspaceRoot))
+
+  await expect(
+    adapter.workspaceFs.exists('.boring-agent/node/node_modules/.bin/boring-ui'),
+  ).resolves.toBe(true)
+  await expect(adapter.workspaceFs.exists('.boring-agent/node/node_modules/.bin/missing')).resolves.toBe(false)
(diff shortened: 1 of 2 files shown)


hachej:
## Update after Gemini 3.1 Pro review

Cross-model review (asked specifically: *is the direct-mode `enforceSymlinkBoundary: false` special-case still needed?*) surfaced two things:

**1. The direct-mode special-case stays — it belongs to `readText`, not `exists`.**
Direct mode is host-native with no sandbox boundary and *relies* on `readText` reading content through escaping symlinks (there's an explicit test). This PR doesn't touch that path, so the special-case is correct as-is. `exists()` no longer routes through that helper at all.

**2. Fixed a real edge case in the original patch: `lstat` → `stat`.**
`lstat` reports a *dangling* symlink as present. If the host's global `@hachej/boring-ui-cli` is ever moved/uninstalled, `.bin/boring-ui` becomes a broken link → `exists()` would return `true` → reinstall skipped forever → workspace permanently bricked. Switched to `stat()`, which follows the link and returns `ENOENT` when the target is gone, so the probe reinstalls and **self-heals**. `stat()` does no boundary check, so it does not reintroduce the crash-loop. Added a dangling-symlink regression test.

**No security regression** (confirmed): `validatePath()` still rejects `../`/absolute/null-byte inputs, and a boolean reachability check on an attacker-planted in-workspace symlink reveals nothing they didn't already know.

Tests: **69 passed**, typecheck clean, invariants pass.

Deferred (out of scope for a bugfix): Gemini suggested replacing the `enforceSymlinkBoundary` flag threaded through every method with a closure-bound intent resolver (`getSafePath(relPath, 'read' | 'write' | 'stat')`). Reasonable cleanup, but a larger refactor — leaving for a follow-up.

hachej:
Folded into #136 so the whole runtime-symlink story is one reviewable PR: tolerate (this exists() hardening), eliminate (sandbox pack→extract), and document (DECISIONS.md #18). Note #136 also *eliminates* the escaping symlink this fix tolerated, so the exists() change is now defensive (dangling-shim self-heal) rather than load-bearing. Closing; branch retired.

PR #135, opened 2026-05-29
docs(decisions): workspace runtime install — symlink-to-global vs self-contained

hachej:
Follow-up to #134. Records the architectural decision behind the symlink-tolerance fix so it isn't re-litigated, and captures the direct-vs-sandbox runtime strategy.

## The decision (DECISIONS.md #18)

The workspace node runtime installs the CLI as a `file:` dependency pointing at the host's **global** install:

```jsonc
// .boring-agent/node/package.json
"dependencies": { "@hachej/boring-ui-cli": "file:…/.npm-global/lib/node_modules/@hachej/boring-ui-cli" }
```

npm materializes a `file:` *directory* dep as a **symlink** (like `npm link`) and auto-generates a `.bin/boring-ui` shim that `realpath`-resolves **outside** the workspace. That's the root cause of the #134 crash-loop.

### Direct mode (no OS sandbox) → embrace the global install out of the box ✅

Per @-feedback: in direct mode the agent already runs on the host with full privileges, the link resolves fine, and reusing the one global `node_modules` is the **fastest, lowest-disk** way to make `boring-ui` + its large dep tree available. A realpath boundary there is pure ceremony. So this is the *correct* default for direct mode — not a wart.

### Sandbox modes (local/bwrap, vercel) → need a self-contained runtime

The symlink target lives **outside the sandbox bind mount**, so inside the sandbox the link is invisible/dangling. These modes need a bundled tarball / preprovisioned base — already designed in [`docs/plans/preprovisioned-base-runtime-acceleration-plan.md`](../blob/main/docs/plans/preprovisioned-base-runtime-acceleration-plan.md). Until that ships, serving runs on the host (where the link resolves), and #134 stops the realpath guard from crash-looping.

### Why not just copy (`npm install --install-links`)?

A copy pulls the CLI's **entire** dependency tree (`vite`, `tailwind`, `react`, `tsup`, `vitest`, …) into every workspace — slow, fat boots. The global-install link exists precisely to avoid that. Making the runtime self-contained *without* paying that cost is the preprovisioning effort, not a one-liner.

## Re-evaluate trigger

Once self-contained sandbox runtimes ship, the symlink-to-global becomes direct-mode-only by design and the symlink-tolerance in #134 can be scoped to direct mode.

Docs-only; no code change.


hachej:
Folded into #136 to keep this work in a single reviewable PR. The decision record (DECISIONS.md #18) now lives with the implementation that realizes it — per the DECISIONS.md Process ("any PR that changes a locked decision must update this document") — and is corrected: #136 makes sandbox installs self-contained now (pack→extract), so the entry scopes remaining future work to the full dep tree + boot speed rather than to symlink elimination. Closing; branch retired.

PR #136, opened 2026-05-30
feat: faster workspace load — unify sandbox runtime materialization + default to direct mode

hachej:
Builds on `work/phase1-curl-uv-bootstrap`. Makes the **local (bwrap)** and **vercel-sandbox** modes provision a workspace's runtime via the **same process**, so neither leaves an escaping symlink — the root-cause shape behind the CLI-mode slow load (see #134, decision #135).

## Before — two different materialization shapes

| mode | external source → | npm/uv result | escaping symlink? |
|---|---|---|---|
| direct | path as-is | symlink → host global dir | yes (fine: no sandbox, host-resolved) |
| **local (bwrap)** | **dir-copy** into `.boring-agent/tmp/…-source` | `npm install <dir>` → **symlink hop** to that dir | no, but link-based |
| **vercel** | **`pnpm pack`** → upload `.tgz` | `npm install <.tgz>` → **extracted copy** | no |

local and vercel both avoided an *escaping* symlink, but by **different mechanisms**. Only vercel produced a true self-contained copy.

## After — one shared pack→extract path

Both sandbox modes now pack the external source into a self-contained tarball **inside the workspace**, then the installer **extracts a real copy**:

- node → `pnpm pack` → `.tgz` → `npm install <.tgz>`
- python → `tar` → `.tar.gz` → `uv pip install <.tar.gz>`

No directory symlink in either mode; the `.bin/boring-ui` shim that tripped the realpath guard in bwrap no longer exists there.

### Changes
- **New** `provisioning/packArtifact.ts` — provider-neutral `packProvisioningArtifact()` + content-addressed `provisioningArtifactName()`, lifted out of the vercel mode so both providers share one path (per the plan's *"reuse the serve-time runtime path, do not build a parallel one"* constraint).
- `vercel-sandbox.ts` — drop the local `prepareVercelProvisioningArtifact` (and now-unused `execFile`/`promisify`/`rename`/`isAbsolute` imports); use the shared helper.
- vercel provisioning adapter — use the shared name helper.
- local adapter — `copyExternalSourceIntoWorkspace` (dir copy → symlink) → `packExternalSourceIntoWorkspace` (tarball → extracted copy). In-workspace sources still fast-path to their `/workspace` path (already visible in the bwrap mount; no packing needed).
- **direct mode unchanged** — no sandbox, so the cheap host-resolved global symlink stays correct there (decision #135).

### Scope note
[...]

PR #137, opened 2026-05-30
refactor(agent): one shared artifact-resolution algorithm for both sandbox adapters

hachej:
Thermo-nuclear code-quality review follow-up to #136, stacked on it.

## The finding

After #136 unified both sandbox modes on pack→extract, the **local** and **vercel** adapters still carried the *same algorithm* in their `resolveInstallSource`:

> name the artifact → if it already exists in the workspace, skip → else pack to a host temp + `copyFromHost` into the workspace → return `${tmpDir}/${name}`

That's duplicated **control flow**, not just a duplicated pack step. #136 only shared the leaf (`packProvisioningArtifact`); the orchestration around it was copy-pasted across two files with subtly different existence checks and error handling.

## The move

Extract one `resolveArtifactInstallSource()` into the provider-neutral `packArtifact` module. Both adapters delegate to it. The only per-mode inputs are:
- the injected `prepareArtifact` packer,
- the mode's `workspaceFs`,
- the sandbox-visible `.boring-agent/tmp` prefix (`runtimeLayout.tmp` vs `/workspace/.boring-agent/tmp`).

Everything else — naming, the exists-skip, temp pack, copy-in, error wrapping — is now **one implementation**.

### What this deletes / fixes
- **Deletes** the bespoke local `packExternalSourceIntoWorkspace` **and** its one-off `pathExists` helper.
- **Shrinks** vercel's `resolveInstallSource` from a 25-line body to a 6-line delegation.
- **Removes a boundary inconsistency**: local now materializes through `workspaceFs.copyFromHost` (the canonical fs boundary) instead of re-implementing `assertRealPathWithinWorkspace` inline, and inherits the same `PROVISIONING_ARTIFACT_FAILED` error taxonomy vercel already had (local previously threw raw errors).
- `VercelProvisioningArtifactRequest` (a public export in `index.ts`) is preserved as a `@deprecated` alias of the canonical `ProvisioningArtifactRequest`.

## Before / after — vercel `resolveInstallSource`

```ts
// before: 25 lines duplicating the local adapter's algorithm
const name = provisioningArtifactName(...)
const workspaceRel = `.boring-agent/tmp/${name}`
if (!(await options.workspaceFs.exists(workspaceRel))) {
  const artifactDir = await mkdtemp(...)
  try { await options.prepareArtifact({...}); await options.workspaceFs.copyFromHost(...) }
  catch (error) { throw toProvisioningError(...) }
}
return `${options.runtimeLayout.tmp}/${name}`

// after
return await resolveArtifactInstallSource({
  workspaceFs: options.workspaceFs,
  prepareArtifact: options.prepareArtifact,
  runtimeTmpDir: options.runtimeLayout.tmp,
  source, opts,
})
```

## Verification
Behavior-preserving. Full agent suite green: **1046 server + 250 front**, typecheck clean, invariants pass. The vercel adapter's caching / `prepareArtifact`-injection / temp-path tests pass unchanged (only the internal `mkdtemp` prefix is now mode-neutral).

> Base is `feat/unify-bwrap-vercel-runtime` (#136); collapses once that merges.

### DIFF
--- a/packages/agent/src/server/runtime/modes/__tests__/provisioningAdapter.test.ts
+++ b/packages/agent/src/server/runtime/modes/__tests__/provisioningAdapter.test.ts
@@ -1,4 +1,4 @@
-import { mkdir, mkdtemp, readFile, symlink, writeFile } from 'node:fs/promises'
+import { mkdir, mkdtemp, readFile, stat, symlink, writeFile } from 'node:fs/promises'
 import { join } from 'node:path'
 import { tmpdir } from 'node:os'
 import { expect, test } from 'vitest'
@@ -103,7 +103,7 @@ test('direct adapter resolveInstallSource returns runtime-visible local paths an
   expect(adapter.getRuntimeCacheRoot()).toBe(paths.cache)
 })
 
-test('local adapter maps workspace-contained package roots to /workspace and copies external roots into writable workspace tmp', async () => {
+test('local adapter maps workspace-contained package roots to /workspace and packs external roots into an in-workspace tarball', async () => {
   const workspaceRoot = await tempRoot('boring-local-workspace-')
   const externalRoot = await tempRoot('boring-local-external-')
   await mkdir(join(workspaceRoot, 'packages', 'plugin'), { recursive: true })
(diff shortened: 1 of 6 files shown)


hachej:
Folded into #136 — the review fixup isn't an independent change, so I squashed `resolveArtifactInstallSource` into the unify commit. #136 now contains the final, deduplicated implementation (one shared artifact-resolution algorithm). Closing to keep the stack at 3 PRs (#134 fix, #135 decision, #136 implementation). Branch `refactor/shared-artifact-resolve` retired.

PR #153, opened 2026-06-02
perf(workspace): skip plugin-authoring runtime install on remote sandboxes

hachej:
## Problem
Provisioning a workspace on a **remote sandbox** (vercel-sandbox, hosted full app) was multi-minute because `createWorkspaceAgentServer` installed the **plugin-authoring runtime** into *every* workspace — `@hachej/boring-ui-cli` + `@hachej/boring-workspace` + the `@hachej/boring-pi` authoring skill — which pulls the CLI's **full dependency tree (~hundreds of MB)** via `npm install` inside the sandbox. But **plugin creation is a local-dev feature**; the hosted agent never runs `boring-ui scaffold-plugin`.

(Verified live: the full app in `vercel-sandbox` mode created a real sandbox fine, but the cold provision dragged on the ~493 MB remote install.)

## Fix
Gate the plugin-authoring tooling on the runtime being **local** — `workspaceFsCapability === "strong"` (direct / local-bwrap), which the code already uses to distinguish local from remote:

| mode | fs capability | authoring runtime + `scaffold-plugin` prompt |
|---|---|---|
| direct, local (bwrap) | `strong` | **installed** (unchanged) |
| vercel-sandbox (+ future remote) | `best-effort` | **skipped** — no install, no pi skill, no prompt |

Remote sandboxes now provision in **seconds** (the ~hundreds-of-MB install is simply not done), and the agent isn't told to use a `boring-ui` bin that isn't there. Local plugin authoring is unchanged.

## Scope note
This is the smaller, surgical win that came out of investigating the slim-runtime plan: since remote sandboxes don't author plugins, we don't need to *slim* the CLI there — we just don't install it. A bundled `@hachej/boring-plugin-tools` for the local-bwrap case remains a possible future optimization (lower priority — the default local mode is `direct`, which symlinks instantly).

## Tests
- `createWorkspaceAgentServer.test.ts`: the `test.each(["direct","local","vercel-sandbox"])` now asserts per-mode — local provisions the CLI + emits the prompt; vercel does neither.
- `workspaceContextPrompt.test.ts`: vercel test updated — the workspace appendix is now empty (no context, no authoring guidance).
- Typecheck clean; the change is behavior-only behind a new optional `installPluginAuthoring` flag (default true, backward-compatible).

### DIFF
--- a/packages/workspace/src/app/server/__tests__/createWorkspaceAgentServer.test.ts
+++ b/packages/workspace/src/app/server/__tests__/createWorkspaceAgentServer.test.ts
@@ -172,7 +172,7 @@ describe("default boring-ui CLI provisioning", () => {
   })
 
   test.each(["direct", "local", "vercel-sandbox"] as const)(
-    "mode %s receives the default CLI provisioning and prompt commands",
+    "mode %s provisions plugin-authoring tooling only for local runtimes (remote sandboxes skip it)",
     async (mode) => {
       const workspaceRoot = await makeTempDir(`boring-cli-${mode}-`)
       let capturedPrompt: string | undefined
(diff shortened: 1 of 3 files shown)


hachej:
Superseded by #162. I rebased #162 onto current main and carried the plugin-authoring runtime gating forward with the slim @hachej/boring-ui-plugin-cli path: full-app/core defaults plugin authoring off, remote sandbox can opt in via installPluginAuthoring / BORING_PLUGIN_AUTHORING=1, and normal workspaces do not provision the authoring CLI.

PR #234, opened 2026-06-09
fix: workspace URL cold-start + legacy session bloat compaction

hachej:
Two load-timing bugs that both manifested as "it won't load until I switch workspaces / reopen differently."

## 1. Workspace won't load on a direct `/workspace/<id>` URL until switching
`packages/cli/src/front/App.tsx`

On a direct hit to `/workspace/<id>`, the selection logic filtered on `available` and, if the workspace was still cold-starting (`available:false`), **silently fell back to the first available workspace**, persisted it to localStorage, and rewrote the URL. From then on `current` outranked the URL id, so auto-refreshes never reconsidered it — only a manual switch recovered.

**Fix:**
- An explicit `/workspace/<id>` is now authoritative — never silently swapped on first load.
- A 1.5s poll self-heals while the workspace initializes; bounded fallback (10 attempts) if it never appears (stale link to a deleted workspace).
- A "Loading workspace…" state replaces the misleading "No available workspaces" screen during cold start.

## 2. Session stuck in "working" mode on first open (legacy bloat)
`packages/agent/src/server/harness/pi-coding-agent/sessions.ts`

Pre-#227 sessions accumulated a full `ui_snapshot` of the entire transcript on every `saveMessages()` call. A 428-message session could reach **90 MB across 60 snapshots**. `resolveSessionTranscript()` (called by every cold-load via `loadEntries()`) read the whole file on every session open, stalling the Node event loop and keeping the UI in a "working" state with the composer disabled.

The #227 rewrite stopped writing new snapshots, but existing bloated files were still read in full.

**Fix:** Repair-on-read compaction — when `resolveSessionTranscript()` encounters `ui_snapshot` entries it strips them out atomically (temp-file + rename) before returning. The first open of a legacy session pays the parse cost once; every subsequent open reads the compacted file. The snapshot entries are never used by the new architecture (`loadEntries` uses `message` entries; `load()` uses `session_info` for title/turnCount).

## Tests
- `App.test.tsx`: URL-targeted workspace is not redirected away while cold-starting.
- `sessions.load.test.ts`: legacy session with 10 `ui_snapshot` records is compacted on first `loadEntries()` call; messages survive intact; file shrinks.

All affected packages typecheck clean; agent/cli test suites pass.

### DIFF
--- a/packages/agent/src/server/harness/pi-coding-agent/__tests__/sessions.load.test.ts
+++ b/packages/agent/src/server/harness/pi-coding-agent/__tests__/sessions.load.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
-import { mkdtemp, rm, writeFile } from "node:fs/promises";
+import { mkdtemp, rm, writeFile, readFile, stat } from "node:fs/promises";
 import { join } from "node:path";
 import { tmpdir } from "node:os";
 import { buildPiChatHistory } from "../../../pi-chat/piChatHistory";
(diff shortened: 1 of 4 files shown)


hachej:
**Branch updated (rebased onto main).**

The #227 baseline rewrite landed on main since this PR was opened and touched both changed files:

**`sessions.ts`** — `saveMessages()` was removed from the call path (#227 replaced `chat.ts` with `piChat.ts`, which doesn't call `saveMessages`). The compaction fix is dropped from this PR; the bug can no longer manifest in the new architecture, and the types it depended on (`UIMessage`, `sanitizeUiMessages`) were deleted in the rewrite.

**`App.tsx`** — The workspace URL fix is fully intact and still needed. The bug (direct `/workspace/<id>` URL silently falling back to the first available workspace on cold start) is confirmed present on main.

The PR now contains exactly one commit with:
- `App.tsx`: URL-authoritative workspace selection + 1.5s cold-start poll + "Loading workspace…" loading state
- `App.test.tsx`: two new tests for the cold-start URL behaviour

CI was green on the old branch; updated CI should confirm the same.

PR #194, opened 2026-06-05
fix(cli): include ask-user by default

hachej:
## Summary
- include ask-user in the CLI frontend plugin list
- register ask-user as a default package in folder mode
- wire per-workspace ask-user runtime/tool/question route in workspaces mode

## Verification
- pnpm --filter @hachej/boring-ui-cli run typecheck
- pnpm --filter @hachej/boring-ui-cli run test src/__tests__/workspacesModeRuntimePlugins.test.ts src/__tests__/folderModeRuntimePlugins.test.ts src/front/App.test.tsx src/__tests__/cli.integration.test.ts

Split from #191 so the media/scroller fixes can release first.

PR #264, opened 2026-06-11
fix(cli): stop Vite optimizer churn + rescan plugin sources on /reload

hachej:
## Context

Diagnosed live on the workspaces hub (global CLI 0.1.35, `boring-ui-workspaces.service`): no plugin panels loaded in the browser despite the server listing and replaying them.

## Fix 1 — Vite dep-optimizer churn breaks plugin front loading

**Symptom:** plugins stuck at `boring.plugin.front-pending`, console shows `ERR_FILE_NOT_FOUND_IN_OPTIMIZED_DEP_DIR` under the global install's `node_modules/.vite/deps`, dynamic imports of plugin modules hang forever inside the app page.

**Cause:** the plugin front runtime's embedded Vite server runs with dep discovery enabled. As plugin imports surface new deps mid-session, Vite re-optimizes, rewriting `.vite/deps` and bumping the `browserHash` — invalidating chunk URLs the browser already holds and stalling in-flight transforms. (Caught it live: `.vite/deps` was being rewritten while the page was broken.)

**Fix:** `optimizeDeps: { entries: [], noDiscovery: true }`. The runtime host serves dependencies through its own proxy/singleton routes, so pre-bundling buys nothing. Verified on the live instance (hot-patched dist): transforms serve in ~3ms with zero `.vite/deps` references; only the small stable react set from `@vitejs/plugin-react`'s `include` is pre-bundled once at boot.

Same change already exists on the PR #194 branch; this PR ports it to main standalone so it can ship as a patch release.

## Fix 2 — plugin sources added after boot are invisible until restart

**Symptom:** restoring package sources to `.pi/settings.json` + `/reload` did nothing; only a process restart picked them up.

**Cause:** workspaces mode resolves discovery roots once when a workspace runtime is created; `/reload` rescans the same frozen root list.

**Fix:** `BoringPluginAssetManager.setPluginDirs()` (small additive API) + `beforeReload` re-resolves roots via `resolveCliBoringPluginDirs` before rescanning. Covered by a new test simulating `boring-ui-plugin install` → `/reload`.

## Test plan

- New test: package source added to `.pi/settings.json` after boot appears after `/reload`
- Existing workspaces-mode suite: 7/7
- Workspace manager suite: 32/32, typecheck clean (cli + workspace)
- Live verification of fix 1 on the running hub

### DIFF
--- a/packages/cli/src/__tests__/pluginFrontRuntime.test.ts
+++ b/packages/cli/src/__tests__/pluginFrontRuntime.test.ts
@@ -492,6 +492,40 @@ describe("pluginFrontRuntime", () => {
     }
   }, 20_000)
 
+  test("maps installed-layout @hachej/boring-workspace imports to the host singleton", async () => {
+    // In an installed CLI the workspace package resolves under
+    // node_modules/@hachej/boring-workspace (not packages/workspace). The
+    // root import must hit the host singleton — a proxied second copy reads
+    // the wrong React context and drags un-interop'd app-level CJS deps.
+    const pluginRoot = await makeTempDir("plugin-front-runtime-installed-ws-")
+    const plugin = await writeRuntimePlugin(pluginRoot, {
+      "front/index.tsx": 'import { useApiBaseUrl } from "@hachej/boring-workspace"\nexport const hook = useApiBaseUrl\n',
+      "node_modules/@hachej/boring-workspace/package.json": JSON.stringify({
+        name: "@hachej/boring-workspace",
+        version: "0.0.0",
+        exports: { ".": "./dist/workspace.js" },
+      }),
+      "node_modules/@hachej/boring-workspace/dist/workspace.js": "export const useApiBaseUrl = () => { throw new Error('proxied copy must not load') }\n",
+    })
+
+    const host = await createPluginFrontRuntimeHost()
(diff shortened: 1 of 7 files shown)


hachej:
Two more production bugs found while end-to-end testing the plugin system on the live hub (with the github-pr-tracker and ccusage-dashboard demo plugins) — both fixed here:

- **`f2030e26`** — the optimizer-chunk filename heuristic singleton-rewrote *any* module named `react.js`, including dockview's own `dist/esm/react.js` (`ReactPart` export). One bad rewrite kills the entire importing plugin module graph with a named-export SyntaxError — silently. Now gated to `/.vite/deps/` paths.
- **`6b32d299`** — the workspace singleton path table only knew the monorepo layout (`/packages/workspace/dist/...`). In an installed CLI, a plugin importing the `@hachej/boring-workspace` root got a proxied second copy of the whole workspace bundle: wrong React context for its hooks, plus un-interop'd CJS deps (`hoist-non-react-statics`) failing the graph. Added installed-layout suffixes (covers npm + pnpm store paths).

Verified live: all three demo plugins now register and render (panels, left-rail chips, hot reload via `/reload`), with the ccusage dashboard fully functional.

PR #291, opened 2026-06-12
fix: tame file-watcher storms — synthetic renames, targeted invalidation, workspace size guard

hachej:
## Problem 1: large moves freeze the file tree

The mv itself is one atomic `fs.rename` — instant. But chokidar has no rename primitive, so a moved directory of N files came back as ~2N `unlink`/`add` events streamed raw over SSE while chokidar re-stats the whole subtree. Client-side, every event invalidated the **entire** `tree` query prefix (every mounted listing) and `FileTreeView` refetched expanded dirs per event with no debounce. Net effect: one rename syscall → multi-second refetch storm, O(N × expanded dirs).

## Problem 2: huge workspace roots kill the server

`watch()` recursively enumerates the entire workspace root with only a short hardcoded ignore list. Started from a giant umbrella folder (`~`, photo libraries, data dumps) that means a multi-minute initial scan, heap growth until OOM ("dies when I start a new chat"), and on macOS without fsevents a silent EMFILE storm — watcher errors were swallowed with `on('error', () => {})`.

## Fixes

1. **Synthetic rename events** — `workspace.rename()` now emits ONE `{op: 'rename', path, oldPath}` event (the op existed on the wire and in the client relay; it was just never produced) and the watcher absorbs the chokidar unlink/add echo for both subtrees via a sliding 1s idle window with a 30s hard cap. Unrelated paths are unaffected.

2. **Targeted tree invalidation** — `useFileEventInvalidation` now invalidates `[base, ws, "tree", parentDir(path)]` instead of the whole `tree` prefix, so a change deep in one folder no longer refetches every mounted listing per 25ms batch.

3. **Workspace size guard** — before starting chokidar, the watcher counts entries with a readdir-only early-exit walk (same ignore set, no fds held, no stats). Past 50k entries (`BORING_MAX_WATCHED_ENTRIES` to override) it refuses: clear server log telling the user to start boring-ui in a subfolder, and the fs-events SSE route relays `unsupported {reason: 'workspace_too_large', message}` via the new optional `WorkspaceWatcher.whenReady()` probe so the client falls back cleanly (it now logs the reason instead of dying silently). First watcher error per instance is also logged now instead of swallowed.

## Tests

- agent: 5 new watcher tests — synthetic rename + echo absorption, unrelated paths unaffected, whenReady ok, cap refusal, ignored dirs excluded from the count (13 pass); fsEvents/file/broadcaster/workspace suites pass (47)
- workspace: invalidation tests updated to targeted keys incl. burst coalescing (22 pass); file-tree suite passes (83)
- typecheck + build green on agent, typecheck green on workspace


hachej:
Applied the thermo-nuclear review findings in 812f6eb6:

1. **Op-aware echo suppression** — the filter now absorbs exactly the rename's predicted echo (`unlink`/`unlinkDir` under `from`, `add`/`addDir` under `to`) instead of muting both subtrees wholesale. A genuine delete or edit inside the moved folder during the window reaches subscribers — pinned by a new test. The filter matches raw chokidar event names because `add` and `change` both collapse to `op:'write'` on the wire.
2. **Canonical helpers** — `parentDir` imported from `treeModel` (was a byte-identical local copy backing cache keys); one `toPosixRel` instead of two duplicate rel formatters.
3. **Honest readiness** — `whenReady` returns `{ok:false, reason:'closed'}` if the watcher closed mid-count instead of reporting a dead watcher ready.
4. **Race documented** — the post-await broadcaster re-check now has a comment explaining the concurrent first-connect race it guards.
5. **Type noise** — readiness closure explicitly typed `Promise<WorkspaceWatcherReadiness>`; three `as const` casts deleted.

Tests: 14 watcher tests (incl. new genuine-events-survive-window), fsEvents/file routes, invalidation suite — all green; agent typecheck+build and workspace typecheck pass.

hachej:
Round-2 thermo-nuclear review applied in 00277cb1:

1. **Regression fix — directory moves stranded descendant caches.** The synthetic rename only carries the top-level `from`/`to`, but file/stat/tree queries are keyed by full path; the per-child cleanup the old echo storm performed was gone with it. An editor open on a file under a moved folder kept stale content on a dead key until window refocus. The `moved` handler now also invalidates everything cached under the old prefix via a predicate filter — only active descendant queries (open editors) refetch; file moves match nothing. Pinned by a test covering descendants, the new prefix, other workspaces, and segment-aware prefix matching (`src-other` ≠ `src/`).
2. **Decomposition** — the watcher subsystem (echo suppression, size guard, chokidar wiring) was ~60% of `createNodeWorkspace.ts`; extracted to `nodeWatcher.ts` (245 lines), leaving a 154-line fs implementation. No behavior change.
3. Last `as const` in `fsEvents.ts` replaced with an explicit `WorkspaceWatcherReadiness` annotation.

56 agent tests + 8 invalidation tests green; agent typecheck/build and workspace front typecheck pass.

hachej:
Round-3 review applied in 9dd1fa61:

1. **No suppression without chokidar** — `emitRename` armed echo-absorption windows even when no chokidar instance was running (no SSE client yet, or the size guard refused the workspace). No echo exists then: the entries leaked (pruned only on chokidar events) and would wrongly mute genuine `add`s if chokidar started inside the window (`ignoreInitial` means the post-start scan never replays the rename). Suppressions now register only while chokidar runs; the synthetic rename fanout stays unconditional. Pinned by a new test renaming before the first subscriber.
2. **Moved dir's own listing** — the descendant predicate matched strictly *under* `from`, leaving `[tree, from]` live for panes mounted at that rootDir. Now matches at-or-under.

15 watcher tests + 8 invalidation tests green, both typechecks pass. Declaring convergence — further rounds would be inventing work.

hachej:
Superseded by combined PR #293, which merges this branch with #292 and includes the shared core bundle-size baseline fix needed for green CI.

PR #292, opened 2026-06-12
feat(agent): canonical pi harness defaults; ambient skills on in the CLI

hachej:
## Why

The pi resource-discovery policy (`noContextFiles`/`noSkills`) was inlined in two parallel harness factories (`createAgentApp`, `registerAgentRoutes`) and implicitly inverted in the skills route. Changing one site silently forked behavior between CLI folder mode, the workspaces hub, and embedders — and the slash-command picker could disagree with the harness about whether ambient skills were on.

## What

- **One definition**: `withPiHarnessDefaults()` in `createHarness.ts` is now the single place the defaults live (`noContextFiles: true`, `noSkills: true`), returning `ResolvedPiHarnessOptions` so consumers get definite booleans enforced by the compiler.
- **One chokepoint per host**: `registerAgentRoutes` resolves + defaults a scope's pi in `resolveScopePi()`, deduplicating the `getPi ? … : opts.pi` resolution that `resolveRuntimeScope` and `resolveSkillScope` each reimplemented. `scope.pi` is now required and pre-defaulted (`scope.pi?.` → `scope.pi.`).
- **Safe-by-construction route**: `skillsRoutes` resolves an unset `noSkills` through the same helper, so a bare registration can't silently flip ambient discovery on, and the picker always matches the agent.
- **CLI = config**: standalone CLI passes `pi: { noSkills: false }` explicitly in both folder and workspaces modes — it runs on the user's own machine, so workspace + user-global `~/.pi` skills are discovered there. Hosted/embed defaults stay off.

## Behavior change for embedders

`registerAgentRoutes`/`createAgentApp` hosts that relied on the skills endpoint listing pi-resolved skills with no config must now pass `pi: { noSkills: false }`.

## Testing

- Full agent suite: 1422 passed / 0 failed (incl. `check-invariants.sh`)
- CLI suite: 52 passed; typecheck clean across `agent`, `cli`, `workspace`
- New e2e test: `.agents/skills` discovery through `registerAgentRoutes` with the CLI's config

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,11 +213,11 @@ Near-term priorities:
 ### Plugins
 
 
-| Plugin                         | What it adds                                                                            | README                                                                   |
-| ------------------------------ | --------------------------------------------------------------------------------------- | ------------------------------------------------------------------------ |
-| `@hachej/boring-ask-user`      | Agent-to-user question/answer surface and `ask_user` tool                               | [plugins/ask-user](plugins/ask-user/README.md)                           |
-| `@hachej/boring-data-explorer` | Searchable, faceted data tables — the primitive for explorer-style panels               | [plugins/data-explorer](plugins/data-explorer/README.md)                 |
-| `@hachej/boring-data-catalog`  | Configurable catalog tab built on `data-explorer`                                       | [plugins/data-catalog](plugins/data-catalog/README.md)                   |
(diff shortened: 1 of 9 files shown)


hachej:
Superseded by combined PR #293, which merges this branch with #291 and includes the shared core bundle-size baseline fix needed for green CI.

PR #316, opened 2026-06-16
fix(workspace): open public landing workspace by default

hachej:
## Summary
- opens the public landing workspace surface by default so the landing page panel is visible

## Verification
- pnpm --filter full-app run typecheck


hachej:
Superseded. The landing copy here is already on main (via #315), and the core change (`defaultSurfaceOpen: true` for the public landing) is reversed by #319, which intentionally starts the no-auth landing with the workspace surface closed. Closing.

PR #392, opened 2026-06-25
Guard surface layout hydration during plugin registration

hachej:
## Context

PR #389 tried to preserve disappearing surface tabs by replacing the Dockview-backed `SurfaceShell`. Thermo review rejected that approach because it broke the real `PaneProps`/Dockview API contract and required unrelated bundle workarounds.

This PR restarts from latest `main` and keeps the existing owner/layer: `ArtifactSurfacePane` remains the only owner of Dockview layout persistence.

## Bug / root cause

A saved Dockview layout can reference app/plugin panels before those panels are present in `allowedPanels` during startup/hot registration. Before this PR, `ArtifactSurfacePane` treated that as an invalid layout and mounted Dockview empty. If Dockview then emitted an empty layout change, the empty layout overwrote the saved one, so tabs were lost.

## Changes

- Split stored-layout reads into explicit states:
  - `ready`
  - `blocked-by-allowed-panels`
  - `empty` / `invalid`
- Preserve valid stored layouts that are only temporarily blocked by `allowedPanels`.
- Skip empty layout writes while a stored layout is blocked by not-yet-registered allowed panels.
- Still allow non-empty writes to replace the blocked layout, so user-created layouts are not suppressed.
- Add regression tests for blocked hydration and non-empty replacement.

## Proof this fixes the bug

Applied the new regression test to unmodified `origin/main`; it failed because the empty layout overwrote the saved layout:

```text
ArtifactSurfacePane — layout persistence > defers layout hydration when a panel is not allowed yet
expected localStorage to remain the saved layout
received {"v":1,"layout":{"grid":{},"panels":{}}}
```

On this branch, the same test passes and then hydrates the saved layout once the previously missing panel is allowed.

## Verification

- `pnpm --filter @hachej/boring-workspace exec vitest run src/front/chrome/artifact-surface/__tests__/ArtifactSurfacePane.persistence.test.tsx`
- `pnpm --filter @hachej/boring-workspace exec vitest run src/front/chrome/artifact-surface/__tests__/SurfaceShell.test.tsx`
- `pnpm --filter @hachej/boring-workspace run typecheck`
- `pnpm --filter @hachej/boring-workspace build`

## Thermo review

- Plan thermo-review: no blockers.
- Implementation loop: no SurfaceShell rewrite, no fake Dockview API, no Vite config change, no new persistence owner.

### DIFF
--- a/packages/workspace/src/front/chrome/artifact-surface/ArtifactSurfacePane.tsx
+++ b/packages/workspace/src/front/chrome/artifact-surface/ArtifactSurfacePane.tsx
@@ -23,43 +23,52 @@ interface StoredEnvelope {
   layout: SerializedLayout
 }
 
-// Read + validate a persisted layout. Returns undefined on any failure
-// (missing, parse error, version mismatch, panel referencing an unknown
-// component) so the caller falls back to a fresh layout. Filtering out
-// individual unknown panels would orphan the grid tree, so we drop the
-// whole layout if any one is invalid.
-function readStoredLayout(
+type StoredLayoutState =
+  | { status: "ready"; layout: SerializedLayout }
+  | { status: "blocked-by-allowed-panels" }
+  | { status: "empty" | "invalid" }
+
+// Read + validate a persisted layout. Malformed/stale payloads are invalid and
+// ignored. Layouts whose panel components are merely not registered/allowed yet
+// are treated as temporarily blocked so an early empty Dockview mount cannot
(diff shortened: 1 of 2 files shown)


hachej:
Manual proof from Julien: workspace playground on this branch (`fix/surface-layout-hydration-guard`, commit `a4a076b9`) was tested at `http://213.32.19.186:5380/` and confirmed working.

CI remains green: 18 passed, 0 failed.

PR #441, opened 2026-06-29
feat(core): add workspace inbox API

hachej:
Slice 3 of #380, stacked on #439.\n\nAdds the core workspace Inbox persistence/API route and database migration.\n\nSize: 20 files, +3047/-1. Most additions are the generated Drizzle snapshot.\n\nValidation on top of the stack:\n- TMPDIR=/home/ubuntu/tmp/pi-tests pnpm --filter @hachej/boring-core exec vitest run src/server/routes/__tests__/inbox.test.ts src/app/server/__tests__/createCoreWorkspaceAgentServer.workspaceBridge.test.ts\n- TMPDIR=/home/ubuntu/tmp/pi-tests pnpm --filter @hachej/boring-core typecheck

PR #438, opened 2026-06-29
feat(workspace): add inbox shell plugin architecture

hachej:
Slice 1 of #380.\n\nAdds the workspace-side Inbox shell/plugin architecture and demo surface on top of the current prerequisite shell stack.\n\nSize: 30 files, +2022/-15.\n\nValidation on top of the stack:\n- TMPDIR=/home/ubuntu/tmp/pi-tests pnpm --filter @hachej/boring-workspace exec vitest run src/shared/plugins/__tests__/frontFactory.test.tsx src/app/front/__tests__/WorkspaceAgentFront.test.tsx\n- TMPDIR=/home/ubuntu/tmp/pi-tests pnpm --filter @hachej/boring-workspace typecheck

Issue #380, opened 2026-06-24
Allow external harnesses to create review/question hooks


hachej:
## Report

Make it possible to create review/question hooks from sessions and harnesses outside boring-ui, such as Claude Code, Codex, or other agent environments.

This would let non-boring-ui coding sessions ask questions, request reviews, or trigger the same owner/agent workflow hooks that boring-ui sessions can use.

## Observed

Review/question hook flow appears coupled to boring-ui sessions and harness context.

## Expected

External sessions and harnesses can create review/question hooks into boring-ui without needing to originate from the boring-ui app itself.

Examples:
- Claude Code session requests owner review/question.
- Codex session requests owner review/question.
- Other agent harnesses can send equivalent structured hooks.

## Context

- Source: user feedback
- Route/panel/plugin: agent/session orchestration, review/question hooks, external harness integration
- Selected item: none provided
- Branch/SHA: `plan/shared-child-app-platform` / `60eef9a8`
- App/browser context: not provided
- Artifacts: none
- Redaction note: feedback was summarized; no secrets, cookies, auth headers, private data, unrelated transcripts, or full local paths included.

### Likely area

Agent/session orchestration APIs and any bridge/protocol used to create review or question hooks from boring-ui sessions.

### Kind

Feature request / integration capability.

### Gate

intake

### Acceptance criteria

1. There is a documented, stable way for non-boring-ui harnesses to create a review hook.
2. There is a documented, stable way for non-boring-ui harnesses to create a question hook.
3. Hooks include enough context to route back to the source harness/session.
4. Hooks are safely redacted and do not require sharing secrets or local-only private paths.
5. Existing boring-ui-originated hook behavior remains unchanged.
6. At least one external-harness smoke path is proven, e.g. Claude Code or Codex.

### Proof path

- Unit/integration coverage for external hook payload validation and routing.
- Manual smoke: trigger a question/review hook from an external harness and verify it appears in boring-ui with source context and can be acted on.
- Regression smoke: trigger the existing boring-ui hook path and verify no behavior change.

hachej:
Kanzen triage

URL: https://github.com/hachej/boring-ui/issues/380
What: External harnesses (Claude Code, Codex, other agent sessions) need a stable way to create review/question hooks that surface in boring-ui for owner input.
Labels: `state:active phase:plan track:owner source:feedback`
Gate: `plan`
Track: `owner`
[...]

PR #444, opened 2026-06-30
docs(human-action): plan target-scoped review actions

hachej:
Follow-up to #441.\n\nPlans generic human-action buttons for any workspace target: artifact, file, HTML, PDF, image, markdown, or custom surface. Inbox keeps only pointers; target renderers own the header buttons.\n\nThis is intentionally a planning PR so implementation can be reviewed separately before wiring cross-surface contracts.

### DIFF
--- a/docs/issues/380/human-action-target-actions-plan.md
+++ b/docs/issues/380/human-action-target-actions-plan.md
@@ -1,0 +1,96 @@
+# #380 Follow-up: generic human-action target actions
+
+## Goal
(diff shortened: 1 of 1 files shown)


hachej:
Closing as obsolete for the current #380 cleanup. This planning-only target-actions PR is not needed for the validated Inbox/ask_user branch (#438), and the target-action work should be replanned/rebased separately if we resume it.

PR #487, opened 2026-07-03
chore(deps): split green production dependency updates

hachej:
Split from Dependabot #478.\n\nThis PR contains the production dependency updates that build/typecheck without source/config migrations.\n\nHeld back for a follow-up PR because they need code/config review:\n- TypeScript 6 / Vite 8\n- ai 7\n- chokidar 5\n- Vercel sandbox 2.3\n- Fastify plugin majors (@fastify/compress, @fastify/rate-limit, fastify-plugin)\n- zxcvbn 4\n- nodemailer 9\n- react-helmet-async 3\n- dockview/dnd majors\n- BI dashboard dependency removals\n\nProof run locally:\n- pnpm install --frozen-lockfile --ignore-scripts\n- pnpm typecheck\n\nDo not merge until CI is green.


hachej:
CI red cause was the core graceful-shutdown test. The green split still resolved `fastify` to 5.9.0 via ranges, and that changes in-flight shutdown behavior (slow request drains immediately -> exit 0 instead of grace timeout exit 1). I moved Fastify to the migration bucket and added a temporary root override pinning `fastify` to 5.8.5 for this green slice.

Local proof after fix:
- `pnpm install --frozen-lockfile --ignore-scripts`
- `pnpm --dir packages/core exec vitest run --no-file-parallelism src/server/app/__tests__/gracefulShutdown.test.ts` ✅

hachej:
Moved better-auth fully out of the green slice and into #488: added a root pnpm override pinning `better-auth` to `1.6.11` for #487 so semver cannot resolve to `1.6.23` under the old `^1.6.11` range.

Also updated #488 on top of the pinned green base so the migration slice explicitly resolves `better-auth@1.6.23` and `fastify@5.9.0` rather than inheriting the green pins.

PR #566, opened 2026-07-08
feat(391-P1): PR A — capability-facts projection (BBP1-007)

hachej:
First slice of the reopened-P1 purity chain (owner ruling: foundation first), per the rewritten P1 pack (#565). **Stacked on #547** — merge the P1 stack first; I retarget after. Facts shape {v:1, runtimeMode diagnostic-only, environments[], tools, skills, mcpServers} exposed through the existing /api/v1/capabilities seam; AgentRouteBindingProfile carries it as plumbing (not a registry); pure mode = honest minimal facts (environments:[], real tool names); direct/local/vercel = documented coarse projection until P2/P3 real bundles. Day-one reader: conformance test consumes the endpoint and asserts honesty for pure + direct, so the surface cannot drift unread. Deliberately excluded: resolver engine, de-mode gating (PR B), core split (PR C), readiness (PR D). Gates: agent typecheck/test (1618), lint:invariants, audit:imports, check:isolation, autoreview clean (0.83). LOC +116 production / +207 tests. Review ~10 min: shared/capabilities.ts, the contributor emission in registerAgentRoutes.ts, the conformance test. 🤖 Generated with Claude Code

PR #568, opened 2026-07-08
feat(391-P1): PR B — de-mode input-asset intake (BBP1-002/007 follow-up)

hachej:
Second purity slice (stacked on #566). Removes runtimeMode FEATURE gating: attachment allow/reject is now decideInputAssetIntake() over environment facts — writable-env-sink | provider-direct-asset | stable-rejection — not config.runtime!==none. runtimeMode stays diagnostic-only (facts/labels). This supersedes BBT2-007 none|direct|workspace: intake is derived, not a declared axis; ERR_NO_FILESYSTEM_FOR_ATTACHMENTS kept for the no-sink case. Excluded: resolver engine, core split (PR C), readiness (PR D). Behavior-frozen — fs modes accept attachments exactly as today; pure rejects because no writable sink. Gates: agent typecheck/test (1621), lint:invariants, audit:imports, check:isolation, autoreview clean (0.85). LOC +93 prod / +85 tests. Review ~10 min: inputAssetIntake.ts + the two call sites in createAgent.ts/harnessPiChatService.ts. **With A+B merged, the mode-label semantics are gone — T2/P3 can safely consume facts.** 🤖 Generated with Claude Code

### DIFF
--- a/packages/agent/src/server/__tests__/createAgent.test.ts
+++ b/packages/agent/src/server/__tests__/createAgent.test.ts
@@ -15,6 +15,7 @@ import type { AgentHarness, AgentHarnessFactoryInput, AgentSendInput, RunContext
 import type { SessionCtx, SessionDetail, SessionStore, SessionSummary } from '../../shared/session'
 import type { PiAgentPromptInput, PiAgentSessionAdapter, PiAgentSessionSnapshot } from '../pi-chat/PiAgentSessionAdapter'
 import { ErrorCode } from '../../shared/error-codes'
+import { decideInputAssetIntake } from '../inputAssetIntake'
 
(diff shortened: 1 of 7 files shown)


hachej:
Plan alignment: this removes mode-label gating with available compatibility facts, but it does not close BBT2-007. E1 must still supply authoritative sinks, ambiguity handling, and provider-direct intake policy before transport intake is complete.

hachej:
Superseded by #642, the current-main P1 readiness-only recut. #642 preserves only the binding-local fail-closed readiness contract and does not cherry-pick this stack. Core relocation and lifecycle/teardown are already landed via #626/#627/#630/#631; pure-mode, input-asset, cache/eviction, provisioning, and old lifecycle changes remain excluded.

PR #575, opened 2026-07-08
feat(391-P1): PR C — relocate runtime-free core to src/core

hachej:
Third purity slice (stacked on #568). Structural move making the Fastify-free core a REAL module the server wraps, not a re-export. src/core/ now contains only createAgent.ts + inputAssetIntake.ts + piChatSessionService.ts + index.ts — no fastify, no server imports, no Pi default harness, no process.env/cwd. The server createAgent.ts injects the runtime adapter + Pi harness + chat service into core and owns non-loaded adapter disposal; routes consume the core PiChatSessionService interface. Behavior-frozen. Gates: agent typecheck/test (1622), boring-bash check:invariants (core Fastify-graph scan: 0 fastify specifiers in 4 dist files), lint:invariants, audit:imports, check:isolation, autoreview clean (0.82). LOC: 668 new core-move lines / 201 ins / 619 del, gross 1488 under cap. Review ~15 min: the src/core boundary + the server injection seam. Next: PR D (readiness honesty). 🤖 Generated with Claude Code

PR #547, opened 2026-07-06
test(391-P1): pr6 — invariants + pure smoke (BBP1-006)

hachej:
**Stacked on #545**. Keep this PR based on `bclaw/391-p1-pr5` until #545 merges; then retarget to `main` without rewriting history.

## Summary

BBP1-006 only: permanent package-boundary/build invariants plus one Node-hosted pure-facade smoke. This PR changes no production source and does not complete P1 or open later lanes.

## Scope

- Replaces the agent-to-boring-bash regex with a TypeScript AST scan over agent source, including `.jsx`. Declaration-level and inline type-only imports/exports are allowed; static values, side-effect imports, dynamic imports, and `require()` edges to `@hachej/boring-bash` fail. Exact TS and JSX detector fixtures guard the classification.
- Keeps `boring-bash check:invariants` responsible for invoking a fresh `@hachej/boring-agent` dependency build. The agent build owns its artifact proof in `assert-build-artifacts.mjs`.
- Resolves the actual published `./core` and `./server` import entries from `packages/agent/package.json`, requires nonempty paths contained under published `dist`, and uses those exact files for artifact parsing and closure analysis.
- Runs both published entries through one esbuild analyzer: `bundle: true`, Node platform, runtime packages bundled, `write: false`, and metafile enabled. The core must contain no reachable `fastify` / `@fastify` inputs or remaining external Fastify specifiers; the legitimate server entry is the same-helper negative proof and must detect Fastify.
- Retains a synthetic metafile fixture covering bundled Fastify inputs plus external `fastify` and `@fastify/*` edges.
- Adds a Vitest Node-environment smoke through `createAgent({ runtime: 'none' })` with a fake harness and one host-registered tool. It proves sealed runtime/session inputs, no workspace/user/cwd leak across every legitimate adapter context, persisted session identity, and facade events from `agent-start` through `message-delta` to terminal `agent-end`. The host tool is registered, not executed.
- Declares esbuild directly in the agent package because its build assertion imports the programmatic API.

## Net Diff

Against final #545 (`origin/bclaw/391-p1-pr5`): **5 files, +578/-45**.

- Production source: 0 files
- Build/invariant tooling: 2 files
- Test: 1 file
- Dependency metadata: `packages/agent/package.json` + `pnpm-lock.yaml`
- All #547 changes to `HANDOFF.md`, `PLAN.md`, `TODO.md`, and `_p1-config-surface.md` were removed; those files exactly match final #545.

## Review Guide (~14 min)

[...]

Issue #391, opened 2026-06-25
Domain-routed agent workspaces, then multi-agent and runtime expansion


hachej:
## Goal

Ship the product in three steps:

```text
Step 1: domain-routed workspace products
Step 2: several agents inside one workspace
Step 3: durable runtime, external A2A, and package extraction
```

## Step 1A — next implementation

```text
exact trusted domain
-> persisted workspaceTypeId
-> authenticated/current-app membership
-> authorized matching workspace
-> exactly one trusted server-only agent behavior
```

- Full-app remains one `default -> primary` compatibility product.
- Seneca proves two domains, workspace types, and agents.
- Typed signup/listing does not create implicitly.
- Explicit typed creation is authenticated, server-stamped, and idempotent.
- No AgentHost, controller, CAS, mutable registry, or compiled deployment authority.

## Later

- **1B:** authenticated external MCP to the same workspace/agent.
- **Step 2:** multiple agents in one workspace, selector/default, native existing subagents.
- **Step 3:** durable events/replay/approvals/recovery, external A2A, `boring-sandbox`/`boring-bash`, custom sandbox tools, channels.
- **Later:** contracted agents in their own workspace/sandbox with governed readonly input and returned artifacts.

## Canonical planning

- Plan: `docs/issues/391/plan.md`
- Decision: `docs/DECISIONS.md` §26
- Consumption modes: `docs/issues/391/AGENT-CONSUMPTION-MODES.md`
- Prebuilt WP alignment: `docs/issues/391/ROADMAP-ALIGNMENT.md`
- Proof: `docs/issues/391/proof.md`
- Plan PR: #812

## Tracker

Active graph: `wt-391-forward-o0b.11`–`.27`.

After #812 merges and planning bead `.11` closes, tomorrow's implementation starts with:

```text
/exec wt-391-forward-o0b.12
```

That first slice only persists `workspaceTypeId` safely; it does not add domain or agent routing yet.

hachej:
## Summary

Refactor boring-ui so `@hachej/boring-agent` can run as a pure/headless agent with **no filesystem at all**, and extract the files + bash working environment into a new optional package: `@hachej/boring-bash`.

Plan source: `docs/plans/boring-bash-agent-runtime-refactor.md`

## Target model

- `@hachej/boring-agent`: model loop, sessions, runner API, tool registry. No implicit files/cwd/bash.
[...]

PR #616, opened 2026-07-10
#391: recut core boundary invariants

hachej:
## Summary

- recuts the valid tooling-only boundary from #547 on current main
- proves the published `@hachej/boring-agent/core` closure is Fastify-free, with the server export as a negative control
- rejects agent value imports from both `@hachej/boring-bash` and `@hachej/boring-sandbox` using a parse-failing AST scan
- adds only the direct esbuild dev dependency and lockfile importer entry needed by the artifact proof

## Scope boundary

No production source or pure/no-environment behavior is changed. The later mechanical core source relocation remains #575 work after the workspace-first production correction.

## Proof

- `git diff --check`
- two independent static reviews plus final structured autoreview: clean
- local tests/build/typecheck intentionally not run per owner direction; GitHub CI owns executable proof

## Review focus

1. published-export resolution and Fastify closure/negative-control semantics
2. type-only allowance versus value-import rejection across TS/JS/JSX forms
3. fresh-build ordering before artifact inspection

Estimated review time: 20-30 minutes.

Issue: #391
Supersedes the obsolete stacked implementation in #547.

### DIFF
--- a/packages/agent/package.json
+++ b/packages/agent/package.json
@@ -116,6 +116,7 @@
     "@types/react-dom": "^19.1.3",
     "@vitejs/plugin-react": "^4.7.0",
     "ajv": "^8.20.0",
+    "esbuild": "^0.25.12",
     "postcss": "^8.5.15",
     "react": "^19.2.7",
     "react-dom": "^19.2.7",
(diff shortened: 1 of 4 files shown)


hachej:
Stack update: the production workspace/runtime correction is now #620. #616 stays boundary/invariant-only and independent; both #616 and #620 must land before the recut #575 core move.

PR #695, opened 2026-07-12
feat(391): activate D1 host collection

hachej:
## Summary

Complete D1-004a4b: strict D1 process wiring, one shared active reader, exact loopback health/readiness bypass, redacted readiness, and production activation before other host effects.

## Issue / Slice

#391 — D1-004a4b, following merged a4a #694 and binding split #693.

## What Changed

- validates D1 host ID, owner UID, process `10001:10001`, and exact `.250/32` one-hop proxy policy before plugin/DB effects
- reads `BORING_D1_OWNER_UID` through the existing process-level `core.env`; uses fixed `/var/lib/boring/d1/<hostId>`
- constructs one exact-DAC reader reused by host scope, landing, and readiness
- permits scope bypass only for unforwarded raw `127.0.0.1` exact GET `/health` or `/internal/d1/readiness`
- fails query/encoded/deep-encoded/malformed/prefix/suffix/method/IPv6/remote/Caddy/spoof variants closed
- switches Compose health to literal `127.0.0.1` and proves health works before an active pointer while readiness is generic 503
- exposes only sorted binding readiness, active revision, and desired digest locally
- preserves generic mode when `BORING_D1_HOST_ID` is absent

## Proof

- `pnpm run build:packages` — pass; core/workspace/agent/plugin artifact assertions pass (existing workspace declaration-bundler warning only)
- `pnpm --filter full-app run typecheck` — pass on fresh recut/current package links
- `pnpm --filter @hachej/boring-core run typecheck` — pass
- full-app focused tests — 3 files / 80 tests pass
- core scope-hook tests — 16/16 pass
- total focused proof — 96/96 pass
- `git diff --check origin/main HEAD` — pass

## Review

- independent security/spec review found and fixed two P1s: malformed/deep-encoded control paths and owner UID values outside Linux `uid_t`
- decoding is now fixed-depth/fail-closed; UID range is canonical `0..4294967294`, excluding app UID
- re-review: CLEAN
- structured review: `/home/ubuntu/.agents/skills/coding-autoreview/scripts/autoreview --mode branch --base origin/main` — clean, no actionable findings
- original runtime patch ID: `176249b99a21448bfadf66c8441dab91b6a5348d`
- follow-up test-narrowing patch ID: `8627fbe86028648bb9ea223526b45b8f0b98ab94`
- combined reviewed branch patch ID: `34060f0b03d63f8a004acb16aedb62c1cb075d0c`
[...]

PR #698, opened 2026-07-13
#391 D1-004b1: fence workspace and signup authority

hachej:
## Summary
- bind workspace list/detail/create/delete authority to the trusted D1 request workspace
- enforce membership before workspace lookup and return redacted 421 for foreign scope
- strip caller-supplied auth scope and install only the canonical host-resolved workspace
- make post-signup accept invites only for the exact bound workspace and suppress scoped default provisioning
- preserve bound rename and generic-host behavior

## Boundaries
- no schema/store change
- no owner/account mutation policy (D1-004b2)
- no public invite resolve/accept fencing (D1-004c)
- no runsc/provider decision
- plan truth-sync and exact b1/b2 contracts are in sibling PR #697; no code dependency

## Proof
- recut patch id: `322f89403b78c1ce4b495e691be149041833b5a9`
- 136 focused b1 tests passed
- 62 #695 host-scope/activation interaction tests passed
- core typecheck, focused ESLint, import audit, core build/assertions, and diff check passed
- independent spec/security review clean
- structured autoreview completed; deferred public invite selectors are explicitly owned by D1-004c


hachej:
CI typecheck fix pushed as `62baddafa0d40b80d2f76d9321d5d98958a1b6e2` (remote-verified). Root cause: the zero-argument Vitest mock inferred `mock.calls` as empty tuples, so the request assertion indexed an impossible element. The test now declares its `Request` parameter and asserts the captured value before checking the internal header. No production change. Local proof: core typecheck; telemetry test 10/10; diff check; structured autoreview clean.

PR #700, opened 2026-07-13
#391 D1-004b2a: make managed member mutations atomic

hachej:
## Summary
- add scoped atomic create-if-absent so concurrent owner creation cannot be overwritten
- evaluate scoped owner demotion/removal inside the same Local/Postgres store mutation seam
- preserve generic store/route behavior and allowed owner promotion/editor-viewer mutation
- add controlled PostgreSQL interleaving proof for both linearization orders

## Contract
Implements merged #699 D1-004b2a. Route-level role reads are not authority; no mutex, route SQL, schema/migration, fault hook, or policy framework.

## Proof
- remote-verified head `3b0396fcf0842917bafdf65fcb0a540801b64921`
- patch ID `6a897fbbdac82e07ce76247cd9a8875a0191caf6`
- 83 local member/Local-store/conformance tests passed
- core typecheck, import audit, core build/assert, diff check passed
- independent atomic/spec review clean
- structured autoreview clean (0.94)

## CI-required proof
The controlled Postgres race tests are committed but could not execute locally because the default `ubuntu:test` database credential is unavailable. CI with its Postgres service must pass before merge.

### DIFF
--- a/packages/core/src/server/app/types.ts
+++ b/packages/core/src/server/app/types.ts
@@ -51,8 +51,9 @@ export interface WorkspaceStore {
   getMemberRole(workspaceId: string, userId: string): Promise<MemberRole | null>
   listMembers(workspaceId: string): Promise<Array<WorkspaceMember & { user: Pick<User, 'id' | 'email' | 'name' | 'image'> }>>
   upsertMember(workspaceId: string, userId: string, role: MemberRole): Promise<WorkspaceMember>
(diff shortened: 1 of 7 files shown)


hachej:
CI hang fix pushed as `174dcf86d7cb2cec77a2ae862033ddb6c4dd8add` (remote-verified). The production lock behavior was correct; the test incorrectly expected both queued operations to be directly blocked by the row holder. PostgreSQL queued waiter 2 behind waiter 1 (`holder <- promotion <- protected`). The harness now captures waiter 1 behind the holder and proves waiter 2 is behind waiter 1. Isolated PostgreSQL 16 proof: update/remove races 2/2 passed; local 83/83; core typecheck/diff; structured and independent review clean.

PR #756, opened 2026-07-14
feat(d1): load deployed agent runtime recipe

hachej:
## Bead

wt-391-forward-vup.2.1.1 - D1-005c A0 deployed-agent runtime recipe

## What changed

- resolves exactly one workspace binding from the active D1 collection and its immutable agent artifact
- exposes frozen runtime identity and recipe resolvers through D1 server wiring
- contributes the selected instructions and resolved digest to the core/agent runtime scope
- memoizes the runtime contribution per request so identity and prompt content cannot drift during one admitted request
- preserves binding-lifecycle admission and cleanup across the new request-scoped contribution

## Decision 25 boundary

This is load-bearing core only. It does not consume or extend RuntimeIsolationEvidenceV1, VerifiedD1HostExecution, approved-host-release, approvedHostArtifactEvidence, hostSecurityConfig, Caddyfile/core-env authority, d1_proof, or any host-artifact attestation ceremony. It does not implement the durable active-pointer publication/recovery half of vup.2.

## Proof

- 37 active-collection reader/runtime-recipe tests passed
- 56 agent register/lifecycle tests passed
- packages/core typecheck passed
- apps/full-app typecheck passed
- packages/agent build passed
- packages/core build passed
- git diff --check passed
- patch size: 358 additions, 17 deletions, 341 net

hostSurface.test.ts / d1Activation.test.ts could not start in this sparse worktree because Vitest module resolution failed on react/jsx-dev-runtime while importing core mail templates; 0 tests executed, so this is recorded as a local environment gap and CI is the gate. NODE_PATH/symlink retry did not change resolver behavior.

## Review findings resolved

- active revision is rechecked after artifact loading to reject mixed-revision reads
- workspace selection fails closed unless exactly one active binding matches
- runtime identity includes the resolved digest and is frozen into request scope
- request-level memoization prevents repeated contribution resolution from changing identity or prompt within one request
- lifecycle tests prove admitted creation and cleanup still use the same scoped runtime

## Risk and rollback

Primary risk is binding a request to stale or mismatched collection data. The resolver fails closed on absent, duplicate, changed-revision, or digest-invalid artifacts, and request-scoped identity prevents mid-request drift. Rollback is a normal revert of this commit; no on-disk publication format or active pointer is changed by this slice.

## Next

A1: add the disposable runtime preloader on top of this stable deployed-agent recipe seam.

## Merge gate

Do not enable auto-merge. This PR requires the mandatory Opus security review before merge.

### DIFF
--- a/apps/full-app/src/server/__tests__/boringMcp.test.ts
+++ b/apps/full-app/src/server/__tests__/boringMcp.test.ts
@@ -39,6 +39,7 @@ const authenticated = { authorization: 'Bearer test' }
 const scopedRequest = {
   bindingId: 'binding-1', workspaceId: actor.workspaceId,
   defaultDeploymentId: 'deployment-1', activeRevision: 'revision-1',
+  resolvedDigest: `sha256:${'a'.repeat(64)}`,
 } as const
 const routeTestConfig: CoreConfig = {
   appId: 'full-app-test', appName: 'Test', appLogo: null, port: 0, host: '127.0.0.1', staticDir: null,
--- a/apps/full-app/src/server/__tests__/managedAgentMcp.test.ts
+++ b/apps/full-app/src/server/__tests__/managedAgentMcp.test.ts
@@ -376,6 +376,7 @@ function scopedRequest(workspaceId = WORKSPACE_ID): CoreRequestScope {
     workspaceId,
     defaultDeploymentId: DEFAULT_DEPLOYMENT_ID,
     activeRevision: 'revision-1',
+    resolvedDigest: `sha256:${'a'.repeat(64)}`,
   })
 }
 
--- a/apps/full-app/src/server/deployment/__tests__/activeCollectionReader.test.ts
+++ b/apps/full-app/src/server/deployment/__tests__/activeCollectionReader.test.ts
@@ -20,7 +20,7 @@ const OWNER_UID = process.geteuid!()
 const APP_GID = process.getegid!() || 10001
 const run = promisify(execFile)
 
-async function desired(): Promise<D1DesiredSnapshotV1> {
+async function desired(content = 'Compare policies.'): Promise<D1DesiredSnapshotV1> {
   const snapshot = canonicalizeWorkspaceCompositionSnapshot({
     schemaVersion: 1, domain: 'boring-workspace-composition:v1', workspaceId: 'workspace:insurance',
     runtimeProfile: {
@@ -32,7 +32,7 @@ async function desired(): Promise<D1DesiredSnapshotV1> {
     policies: { externalPlugins: false, pluginAuthoring: false },
   })
   const compositionDigest = await createAgentAssetDigest(JSON.stringify(snapshot))
-  const instructions = { path: 'instructions.md', content: 'Compare policies.', digest: await createAgentAssetDigest('Compare policies.') }
+  const instructions = { path: 'instructions.md', content, digest: await createAgentAssetDigest(content) }
   const compiledDefinition = { schemaVersion: 1 as const, definitionId: 'definition:insurance', version: '1.0.0', instructionsRef: instructions.path }
   const definition = { definitionId: compiledDefinition.definitionId, version: compiledDefinition.version, instructionsRef: compiledDefinition.instructionsRef,
     digest: await createAgentDefinitionDigest({ definition: compiledDefinition, assets: [instructions] }) }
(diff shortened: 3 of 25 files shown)


hachej:
## CI red at `workspacesModeRuntimePlugins.test.ts:469` — root cause + fix

**Symptom (confirmed from actual CI logs, run 29338638234):** after `DELETE /api/v1/local-workspaces/:id`, `GET` on the evicted plugin's `frontTarget.entryUrl` returned **200**, expected **404**.

**Investigated and ruled out:** the plausible-sounding "new `getRuntimeScope` WeakMap memoization in `registerAgentRoutes.ts` leaks the runtime target past eviction" theory. That code path is structurally disjoint from this test — `/api/v1/agent-plugins`, `/api/v1/agent-plugins/events` (SSE), `DELETE /api/v1/local-workspaces/:id`, and the plugin front-runtime `GET` route all live in `packages/cli/src/server/{modeApps,pluginFrontRuntime}.ts` and never call into `registerAgentRoutes`'s dispatcher/`getRuntimeScope`/`bindingLifecycle`. `packages/cli` is untouched by this PR's diff (confirmed via `git diff <merge-base>..HEAD --stat`). I also reverted `registerAgentRoutes.ts` to `main`'s version locally, rebuilt, and re-ran this exact test — same behavior either way. The new `registerAgentRoutes` lifecycle tests (`registerAgentRoutes.lifecycle.test.ts`, incl. the new "one request pins one runtime recipe..." test added by this PR) all pass and correctly verify the memoization's eviction/disposal semantics on `app.close()`.

**Actual root cause (verified against real CI logs):** a pre-existing TOCTOU race in `packages/cli/src/server/pluginFrontRuntime.ts`, unrelated to this PR's changes, that CI's slower/more-contended runners hit more reliably than this sandbox's idle machine:

1. `getOrCreatePluginRuntime()` (`modeApps.ts`) fires a **fire-and-forget** `runtimeHost.warmupWorkspace(workspaceId)` right after a plugin list loads ("Fire-and-forget: pre-transform ...").
2. That warmup calls `serve()`, which `validateRequest()`s (confirms the plugin is tracked) and then unconditionally does `transformCache.set(...)`.
3. `disposeWorkspace()` (triggered by the `DELETE`) does a **point-in-time** cleanup: delete `trackedWorkspaces`/`trackedPluginRevisions`, then sweep `transformCache` for entries matching the workspace.
4. If the warmup's `validateRequest()` already passed but its `transformCache.set()` hasn't run yet when step 3's sweep executes, the sweep finds nothing to clean — and the warmup then plants a **new, orphaned** cache entry for a workspace that was just evicted. Nothing ever cleans it up, so a later `GET` for that `entryUrl` is served straight from the stale cache → 200 instead of 404.

**Fix** (`packages/cli/src/server/pluginFrontRuntime.ts`): track a per-workspace disposal epoch (`workspaceDisposalEpoch`). `disposeWorkspace()` bumps it synchronously before its cleanup. `serve()` captures the epoch right after `validateRequest()` succeeds and re-checks it immediately before `transformCache.set()`; if the epoch moved (workspace evicted mid-flight), it discards the result and throws the same 404 the disposal sweep would have produced, instead of caching a result for an evicted workspace.

`registerAgentRoutes.ts` / the D1 recipe wiring are unchanged — this PR's shared-dispatcher and D1 logic were not implicated.

**Verification:**
- `packages/cli` workspaces-mode suite (`workspacesModeRuntimePlugins.test.ts`), incl. the eviction test at line 469: green, run repeatedly (20+ runs, no flake).
- `packages/agent` `registerAgentRoutes.test.ts` + `registerAgentRoutes.lifecycle.test.ts`: 56/56 pass.
- `packages/cli` build + `tsc --noEmit`: clean.
- `packages/agent` build: clean.
- `pnpm run lint:invariants` (repo-wide): pass.

Force-pushed with lease. Bead vup.2.1.1 stays `in_review` — not arming auto-merge, flagging for Fable re-review given the root cause differs from the original diagnosis.

hachej:
Fable final word: APPROVED. Opus review cleared the D1 recipe logic (well-fenced, TOCTOU-guarded, fails-closed, Decision-25-clean, sibling-isolated) but mis-diagnosed the CI red as a registerAgentRoutes memoization leak. The Sonnet fixer DISPROVED that (reverted the file, test failed identically) and root-caused the real bug: a pre-existing TOCTOU race in pluginFrontRuntime.ts where a fire-and-forget warmup cache-write lands after disposeWorkspace()'s sweep, leaving an evicted workspace's runtime target answering 200. Fixed with a per-workspace disposal epoch (bumped on eviction, re-checked before the cache write; eviction test untouched, now green). Good catch — the fix hardens a real tenant-cleanup race. Closes bead wt-391-forward-vup.2.1.1.

Issue #808, opened 2026-07-17
#391 child: Sandbox provider extraction and S3/FUSE mounts


hachej:
Parent: #391

## Ownership
Sandbox-provider extraction and native S3/FUSE mounts: P2 and X1.

See the canonical issue plan at `docs/issues/808/plan.md`.

## Current state
The physical plan redistribution from #391 is complete. This issue's canonical work-package documents now live under `docs/issues/808/runtime-refactor/`; former #391 Markdown paths are direct redirect stubs. Decision 25 does not cancel this independently owned work, which remains outside #391's static P0→N1 critical path.

## First blocker
Retain the issue's existing named-consumer/design/Bead gate before dispatching implementation. The completed path move does not authorize runtime work.

## Scope boundary
This was a path-only documentation move. No runtime source, package behavior, release, migration, or deployment behavior changed.

PR #823, opened 2026-07-18
#808 plan: recut boring-sandbox extraction remaining work (Decision 26)

hachej:
## Summary

- Recut the remaining boring-sandbox provider extraction against current main and Decision 26.
- Separate verified Today state from the remaining direct, bwrap, and Vercel provider delta.
- Define one versioned Workspace + Sandbox pair contract, explicit lifecycle gates, and a small proposed Bead chain.

## Proof

- Docs-only diff: docs/issues/808/plan.md
- No product code or .beads edits
- git diff --check passes
- Remote branch SHA verified with git ls-remote
- Tier-1 fresh-eyes and tier-2 architecture review findings incorporated; final convergence pass clean

## Boundaries

This is a plan-only PR. It does not merge or enable implementation: execution remains gated by Decision 26 Step 2 and owner activation. X1 S3/FUSE mounts, attestation, and new provider features remain out of scope. This PR does not close the implementation issue.

### DIFF
--- a/docs/issues/808/plan.md
+++ b/docs/issues/808/plan.md
@@ -2,7 +2,7 @@
 github: https://github.com/hachej/boring-ui/issues/808
 issue: 808
 state: ready-for-human
-updated: 2026-07-18
+updated: 2026-07-19
 flag: not-needed
 track: owner
 ---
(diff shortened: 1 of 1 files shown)


hachej:
Owner direction 2026-07-19: SaaS will run on our own cloud; vercel-sandbox is a bridge, not a permanent dependency. The `SandboxProviderV1` extraction in this plan is therefore load-bearing for the future remote-worker+runsc provider (bead SBX1, `wt-391-forward-6gd`, deferred, gated on this extraction). No scope change requested to this PR — extraction stays exactly as planned; runsc stays frozen here.

Owner directive on sequencing — extraction must be copy-verify-swap, not move-in-place, and this should be reflected in the plan's step/bead structure if it isn't already:
1. **Copy** the existing provider implementations (bwrap, vercel-sandbox, direct) from `packages/agent` into `packages/boring-sandbox` behind `SandboxProviderV1`, leaving `packages/agent`'s live path completely untouched.
2. Run `boring-sandbox` **standalone** first: its own conformance/parity test suite proves the copied providers behave identically (the same suite should be runnable against both the old in-agent path and the new package to demonstrate parity).
3. Only once standalone is proven, a separate near-mechanical **swap PR** rewires `packages/agent` to consume `boring-sandbox` (imports flipped, zero behavior delta) and deletes the old copy.
4. **Freeze rule** during the copy window: no changes to `packages/agent`'s sandbox code once copying starts — divergence between the two copies is the failure mode.

hachej:
Update: the copy-verify-swap directive above is now encoded in the plan itself as of c21e2a9d8 — new "Execution phasing" section (Phase 1 COPY / Phase 2 STANDALONE VERIFY with dual-target parity gate / Phase 3 SWAP), explicit freeze rule on packages/agent sandbox code with divergence as the named failure mode + stop condition 10, and the bead chain restructured to one bead per phase. Please review the latest revision rather than the earlier comment.

Issue #900, opened 2026-07-22
Add a thin full-catalog Composio mode to boring-mcp


hachej:
## Named consumer

Generic Seneca: https://github.com/hachej/seneca/issues/35

This is **MCP egress/consumption by a Seneca Agent**. It is not #806's opposite-direction managed-Agent MCP ingress and does not broaden Constellation's curated tenant configuration.

## Owner scope decision

Ship the fastest sellable product as a **thin wrapper around one Composio integration**:

- Composio remains the source of truth for providers/apps, tools, schemas, managed auth, connected accounts, and execution.
- Every valid app-native tool exposed by Composio MCP for every provider is searchable, describable, and callable.
- Use one full-catalog Composio Session per isolated Seneca user/workspace scope; omit the one-toolkit filter.
- Do not mirror the provider catalog, generate one Boring template per app, or build another credential/marketplace system.
- Boring adds only identity isolation, server-only secret custody, bounded/redacted transport, and the execution approval boundary.
- For the sellable launch, **every call requires one exact Ask User approval**. Verified direct reads are optional post-launch work, not a blocker.

## Explicitly out of scope

- Official MCP Registry.
- User-supplied/custom MCP URLs.
- `stdio`, remote workbench, remote bash, or arbitrary package execution.
- Per-server credential profiles or a second secret store.
- Loading thousands of generated direct tools into model context.

## Required reusable capability

1. Create a Composio Session with `mcp: true`, no `toolkits.enable` filter, connection management enabled, and sandbox/workbench disabled.
2. Map Boring's bounded search/describe/connect/call surface shallowly to Composio Connect's own meta-tool flow.
3. Pin the exact selected connected account using Composio `connectedAccounts`/execute account selection; never silently use recent-account fallback.
4. Keep raw execution/control-plane meta-tools unreachable so they cannot bypass approval.
5. Reuse the exact shared `@hachej/boring-ask-user` runtime at the provider-call boundary.
6. Bind one-use approval to actor, Agent session, source/account, tool/version, schema/policy/source revisions, and canonical arguments.
7. Never automatically retry execution; return an explicit unknown outcome when commit status is ambiguous.
8. Preserve curated consumers unless they explicitly select the full-catalog mode.

## Fast sequencing

- First run an isolated real-Composio capability spike proving unfiltered catalog behavior, sandbox disablement, meta-tool control, bounded query/schema, and exact pinned-account execution.
[...]

PR #941, opened 2026-07-25
fix(boring-mcp): preserve injected provider methods

hachej:
## Summary
- replace object spread of injected `ManagedConnectorProvider` with explicit receiver-preserving delegation
- retain optional abort/revoke behavior and catalog cleanup
- add regression coverage using a class-instance provider through connect, refresh, and disconnect routes

## Validation
- `pnpm --filter @hachej/boring-mcp typecheck`
- `pnpm --filter @hachej/boring-mcp test` (116 passed)


hachej:
Superseded by the owner-approved full revert in #946. We will redesign and re-land issue #900 in smaller lifecycle-safe slices after the release.

Issue #909, opened 2026-07-22
AgentGateway v0: freeze session gateway contract + canonical createAgentHost() composition


hachej:
## Goal

The grunt-work spine that unlocks everything else in the #905 direction: freeze one server-side \`AgentGateway\` contract, standardize the three divergent agent-construction paths behind one canonical \`createAgentHost()\` factory, and migrate the three consumers (workspace app server, core server, CLI) onto it. After this lands, all remaining #905 work (streaming activation, definitions/catalog revival, #861, v2 remote host) becomes independent parallel lanes gated by a conformance suite.

## Why now

Code verification (2026-07-22) confirmed:
- There is no interface today — Workspace *constructs* agents. Three consumers build the agent three different ways: \`packages/workspace/src/app/server/createWorkspaceAgentServer.ts\` (calls \`createAgentApp\`, picks sandbox providers), \`packages/core/src/app/server/createCoreWorkspaceAgentServer.ts\` (hand-assembles \`registerAgentRoutes\`), \`packages/cli/src/server/modeApps.ts\` (reaches into Workspace for Agent's own runtime adapter).
- \`createAgentApp.ts\` (~478 lines) and \`registerAgentRoutes.ts\` (~1474 lines) are parallel duplicated constructors that each independently build harness + bridge + tools.
- \`apps/agent-playground\` imports Workspace backwards.
- No \`AgentGateway\`/\`AgentHost\` symbols exist yet; naming collision with core's \`BORING_AGENT_HOST_ID\` must be resolved.

## Scope (v0, embedded only)

1. **G1-lite** — five-method session \`AgentGateway\` contract (listAgents / listSessions / createSession / connectSession / send+control) + session state snapshot read + conformance suite with two levels (bounded-replay+rehydrate now, durable-offset later). Types/DTOs only; no files/pluginAssets/egress surfaces in v0.
2. **AH0** — canonical \`createAgentHost(options)\` factory funneling the duplicated constructors; \`EmbeddedAgentGateway\` facade; runtime-mode adapter ownership moves Workspace→Agent.
3. **Consumer migrations** — workspace, core, CLI each swap bespoke construction for the factory + injected gateway; agent-playground backward import removed.

Front: unchanged wire (\`PiChatEvent\` HTTP/NDJSON projection stays byte-compatible); only multi-agent addressing (agentTypeId in session refs/routes) is added.

[...]

PR #811, opened 2026-07-17
#775: adopt and rename one native Pi session

hachej:
Closes #775.

## Summary

- keeps unsent New chat explicitly ephemeral and browser-only
- creates and adopts exactly one native Pi session on first trusted direct/local send
- reconciles response loss/gateway ambiguity with the same idempotency key
- preserves native identity through follow-ups, deletion, reset, navigation, and detached views
- enables assistant-gated native rename without message-time reordering
- adds compact Copy ID / Rename / Delete actions while keeping Pin/Open direct

## Explicit exclusions

No legacy wrapper migration, credential-rotation framework, generic attachment recovery, hosted durability, task binding, or broad activity framework.

## Proof

See `docs/issues/775/proof.md`.

- Agent full suite: 1,737 passed, 6 skipped
- Workspace full suite: 1,738 passed, 10 skipped
- builds, typechecks, invariants, and diff check passed
- final structured Auto Review clean (`/tmp/775-final-autoreview-f125.txt`)
- native Pi interoperability/export passed with ID `019f7fe5-88e1-7edf-a8ea-cb0a0b085d16`

## Review size

- 57 files, +4,496 / -271
- production/docs: +1,968 / -255
- tests: +2,528 / -16

The reduction pass removed duplicated tests and dead code, then corrected concrete lifecycle defects found by independent review. The final diff is only modestly smaller than the previous lean head because those correctness fixes offset most line-count savings; no behavior was deleted solely to meet a numeric target.

Issue #775, opened 2026-07-15
Native Pi session creation and rename


hachej:
## Goal

One materialized Boring chat equals one native Pi session. Boring and standalone `pi /resume` show the same transcript/title once the user has sent a message.

## Scope

- New chat is browser-only before send; no rename and no durable empty session.
- First send creates exactly one native Pi JSONL and no Boring wrapper / `pi_session_file`.
- After first assistant response, rename is enabled and writes native Pi title through Pi API.
- Boring list/load for newly-created sessions reads that native transcript.

## Out of scope

Legacy wrapper migration, metadata index, activity, hosted semantics, event-stream migration, task bindings.

## Proof

Manual: New chat → send → first reply → rename → `pi /resume` shows exactly one titled session.

PR #968, opened 2026-07-27
#775 fix: carry session identity across the addressed wire

hachej:
Follow-up to #811. Fixes four regressions reported in the workspace playground after #811 merged.

## Symptoms

- Creating a session left the chat area blank until the user clicked somewhere.
- Switching sessions did nothing — every click snapped back to the first session.
- The first message in a session got no response; only the second did.
- `Rename` never appeared in the session row menu (Copy/Delete did).

## Root cause

One defect, not four. #909 switched the playground to **addressed** mode (`agentTypeId="default"`), but the addressed AgentGateway summary dropped `agentTypeId`, `nativeSessionId` and `hasAssistantReply`, and hardcoded `turnCount: 0`. The legacy mapper beside it preserves all of them (`usePiSessions.ts:689-691`).

Losing `agentTypeId` put every workspace key derivation in a different key domain than `chatSessionKey`, so the pane-sync effect never matched the active session — it fought the newly created pane (blank area), snapped every switch back to `sessions[0]` (switching broken), and remounted the pane mid-stream so the first prompt's SSE was lost (first message ignored). Losing the other two fields meant `renameAvailable` could never pass.

This surfaced only once #811's key-domain work and #909's addressed transport met in a merge; neither side was wrong alone.

## Changes

- Carry `turnCount` / `nativeSessionId` / `hasAssistantReply` through `AgentSessionSummary` and `summaryFromLegacy`.
- Map them in the client addressed mapper, including `agentTypeId` from the session ref.
- Fall back to the host `agentTypeId` in `useDefaultWorkspacePiSessions` so all key derivations stay in one domain.

## Testing gap this closes

CI was green through all of this because the existing addressed test (`usePiSessions.addressed.test.tsx`) only asserted `agentTypeId` appeared in request **URLs** — it never checked that session **fields survived the response mapping**.

The added regression test asserts they do, and **it fails without the mapper fix** (verified by stashing only that change), so it genuinely guards the behavior.

## Proof

- agent: 1841 passed, 13 skipped
- workspace: 914 passed
- `tsc --noEmit` clean for agent and workspace front

Browser verification of the four symptoms is still pending.


hachej:
**Coordination note from the #909 follow-up backlog (2026-07-27).** This PR overlaps two tracked beads — please sequence deliberately so we don't repeat the #811 outcome (23 green CI checks on a commit that broke first-send, session switching, and initial chat render; caught only by manual smoke, reverted same day by #967):

1. `wt-391-forward-0jpy.27` (front addressing) rewrites the same files (`usePiSessions`, `remotePiSession`, `piChatStream`) to make the reload-reconnect stream path addressed-aware and add dynamic agent selection. If this PR lands first, .27 rebases on it — fine — but check you're not re-introducing legacy-wire assumptions .27 removes.
2. `wt-391-forward-0jpy.28` (the #811 re-land) requires three regression tests as **workspace-playground E2E specs that actually run in CI**: (a) create session → FIRST message → assert streamed response; (b) two sessions → switch → correct transcript; (c) create session → chat renders without user interaction. The E2E suite currently runs in NO CI job (`pnpm e2e` only covers packages/agent) and can't run from a clean checkout (untracked fixtures) — bead `.26` fixes that.

Bottom line: green CI on this PR does not currently cover the failure modes that sank #811. Before merge, please run the three checks above manually in workspace-playground (or land .26 first and add them as specs). Happy to share the A/B repro method used to isolate the #811 regressions.

hachej:
**Supersession notice (2026-07-28, orchestrator).** The three #775/#811 regressions this PR targets (first message unanswered, session switch broken, blank chat) have been re-fixed on `integration/wave-1` by PR #982, built on the addressed front (#976) and multi-agent console (#981), with the three regressions as CI-run E2E specs. #982 deliberately does NOT touch the frozen `shared/gateway/types.ts`, which this PR modifies. The two fixes share a file set (`usePiSessions`, `remotePiSession`, `nativeFirstSendTransactions`, `embeddedGateway`) with different transaction models — merging both would conflict at best and produce two live first-send mechanisms at worst. **Please do not merge this to main before a reconciliation pass against `integration/wave-1`.** If the summary-field additions here (`turnCount`/`nativeSessionId`/`hasAssistantReply`) are still wanted, they should be re-proposed as a small additive PR against the integration branch — the owner decides. cc @hachej

hachej:
Closing as superseded.

This PR fixes four regressions that #811 introduced on the **legacy** pi-chat wire. Wave 1 took a different route: #811 was reverted wholesale, and the same native session adoption + rename behaviour was re-landed on the **addressed** gateway wire in #982, which is part of #1008.

The regressions this PR targets cannot recur on that path — the addressed wire is covered by `sessionIsolation.test.ts` and the session-identity work in #976/#982. Nothing here is lost; the destination changed.

If any specific fix in this diff turns out to have no equivalent in #1008, reopen and say which — that would be a real gap.

hachej:
Verified superseded: merged #1044 is the stated semantic port of this PR's session-correctness invariants onto the addressed AgentHost architecture, and all distinctive fields (hasAssistantReply, nativeSessionId, agentTypeId, turnCount, renameAvailable gating) are confirmed present on main.

Issue #1012, opened 2026-07-31
Lane: sandbox isolation for untrusted code


hachej:
Lane issue for **sandbox isolation** (DIRECTION Wave 4). Replaces closed #808 as the tracking issue; PR #916 currently points at a closed issue.

## Today (verified against `origin/main`, 2026-07-31)
Both packages are real and published at 0.1.93: `@hachej/boring-sandbox`, `@hachej/boring-bash`. (`boring-bash`, despite the name, is filesystem/shell *binding* code — pi tool wiring and workspace file/git/search/watch routes — not a sandbox.)

Interface `SandboxProviderV1` (`packages/boring-sandbox/src/shared/providerV1.ts`), capability facts in `providerMatrix.ts`, wiring in `providers/static.ts`.

**Backends that actually execute:**
- `direct` — raw `child_process.spawn`, `hardening: "none"`. Zero isolation, by design.
- `bwrap` — genuine: `--unshare-all`, `--die-with-parent`, `--new-session`, tmpfs root, ro-binds, optional `--cap-drop ALL`.
- `vercel-sandbox` — real remote provider with snapshot/bake/circuit-breaker.

Strongest real path today is full-app's agent worker (`apps/full-app/src/server/agent-worker.ts` + `packages/agent/src/server/worker/*`): bwrap with `dropAllCapabilities`, internal-token auth, `ExecSemaphore`, and `buildExecEnv` scrubbing `*_API_KEY`/`*_TOKEN`/`*_SECRET`/`DATABASE_URL`.

## Two defaults that are true right now, not Wave-4 problems
1. **The CLI defaults to `direct` — no isolation — on every platform**, including Linux with bwrap present. bwrap is opt-in behind `--mode local-sandbox` for boot-cost reasons (`packages/cli/src/server/cli.ts:571-583`). Defensible for a local dev tool the user controls; wrong the moment anything less-trusted runs.
2. **Neither real backend isolates network egress.** bwrap defaults to `--share-net`; the matrix records `networkIsolation: "none"` for both `direct` and `bwrap`.

## Dormant
- **runsc/gVisor does not execute on main.** `providers/runsc/` is preflight + evidence + admission only (`preflight.ts`, `isolationEvidence.ts`, `fleetAdmission.ts`, `qualificationBundle.ts`), with a constant literally named `RUNSC_UNPROVEN_SECURITY_FACTS`.
[...]

Issue #1011, opened 2026-07-31
Lane: external MCP — user-registered servers


hachej:
Lane issue for **external MCP** (DIRECTION Wave 3). Umbrella for the egress side; #806 stays the ingress issue, #900 stays the full-catalog slice.

## Today (verified against `origin/main`, 2026-07-31) — this genuinely works
Plugin `@hachej/boring-mcp` (`plugins/boring-mcp/`) is a real, tested implementation (~15 test files):
- Real MCP client over the official SDK: `src/server/mcpSdkTransport.ts` — `@modelcontextprotocol/sdk` `Client` + `StreamableHTTPClientTransport`, per-source endpoint + headers. Generic enough for any Streamable-HTTP endpoint.
- Composio-managed connectors: `composioManagedConnector.ts` (session + hosted OAuth connect URL), `managedConnectorAdapter.ts`, `managedConnectorPreflight.ts`, `hardening.ts` (rate budgets, redaction canaries, origin pinning).
- **7 agent tools live in the model's toolset**: `mcp_servers_list`, `mcp_server_status`, `mcp_server_doctor`, `mcp_server_probe`, `mcp_tools_search`, `mcp_tool_describe`, `mcp_readonly_call` — wired via `agentBridge.ts` → `agentTools.ts` → `appServerBinding.ts` → `apps/full-app/src/server/main.ts:69` `getExtraTools`, consumed by `registerAgentRoutes.ts:808`. Note this is the app's extra-tools seam, **not** `buildAgentComposition`.
- Read-only by construction: deny-before-allow `classifyMcpTool`; only `mcp_readonly_call` executes.
- Front overlay `BoringMcpSourcesOverlay` mounted at `apps/full-app/src/front/main.tsx:136` (prod-only flag).
- Flags `BORING_MCP_ENABLED` / `BORING_MCP_PROD_ENABLED`; one server-env `COMPOSIO_API_KEY`.

Ingress (exposing Boring as an MCP server) also exists — `packages/agent/src/server/mcp/managedAgentMcpServer.ts`, app glue at `/mcp/managed-agent` — but is **hard-disabled in every released host**: `hostSecurityConfig.ts:163` freezes `managedAgentMcp: {enabled: false}`.

## The actual limit
[...]

Issue #1010, opened 2026-07-31
Lane: BYOK — per-workspace model credentials


hachej:
Lane issue for **BYOK / per-workspace credentials** (DIRECTION Wave 3). Replaces closed #820 as the tracking issue; PR #917 currently points at a closed issue.

## Today (verified against `origin/main`, 2026-07-31)
- Encrypted generic settings store: table `workspace_settings` (`packages/core/src/server/db/schema.ts:106-119`), pgcrypto `pgp_sym_encrypt/decrypt` with the app-wide `WORKSPACE_SETTINGS_ENCRYPTION_KEY` (`PostgresWorkspaceStore.ts`). `GET` returns masked `configured` booleans only.
- **Provider API keys for model calls do not come from Boring at all.** `createHarness.ts:583-588` calls Pi's `AuthStorage.create()` + `ModelRegistry.create()`; the comment states credentials are Pi-owned. Extra providers register from env (`server/models/modelConfig.ts`).

## Dormant — the contract is built and unreachable
`packages/agent/src/shared/credentials/` is a complete frozen BYOK contract: `registry.ts` (provider definitions incl. `llm`), `bindings.ts` (delivery `host-only|sandbox-pipe|sandbox-tmpfs`), `ref.ts`, `authority.ts` (opaque scope, Core-owned verifier), `lease.ts` (resolver + lease with `dispose()`), `sandboxDelivery.ts` (ships a deliberate not-implemented resolver). Server side: `server/credentials/hostResolver.ts`, `withResolvedCredential.ts`.
**Its only reference outside the module is the barrel re-export** (`server/index.ts:234-236`). Nothing constructs a registry or resolver. The harness never touches these types.

## Delta to "a user pastes their key and their agents use it"
1. **UI** — none exists. The only key-entry surfaces in the repo are Stripe/LemonSqueezy for credits.
2. **Store backend** — no `CredentialStoreBackendV1` implementation on main.
3. **Authority verifier** — no Core implementation of `WorkspaceCredentialAuthorityVerifierV1` (only a fake in tests).
4. **Injection seam** — the hard one. The harness has nowhere to accept an invocation-scoped credential; it delegates wholesale to Pi's env-only path.
5. **Precedence policy** (workspace key vs instance fallback) is documented (`docs/DECISIONS.md:450`) with no code.
6. `sandbox-pipe`/`sandbox-tmpfs` delivery is a stub.

## Parked PR #917 does not close this
[...]

Issue #1009, opened 2026-07-31
Lane: chat streaming durability (Level B → Level D)


hachej:
Lane issue for **chat streaming durability** (DIRECTION Wave 2). Supersedes nothing; complements #807.

## Today (verified against `origin/main`, 2026-07-31)
Streaming is in-memory and per-process, and it works correctly at Conformance Level B:
- `PiChatReplayBuffer` (`packages/agent/src/server/pi-chat/piChatReplayBuffer.ts`) — bounded ring, default 1000 events, monotonic `seq`, subscribe-before-replay so nothing is lost in the handoff.
- NDJSON (not SSE) on both wires: legacy `/api/v1/agent/pi-chat/:sessionId/events?cursor=` and addressed `/api/v1/agents/:agentTypeId/sessions/:sessionId/events?cursor=`.
- Client (`front/chat/pi/remotePiSession.ts`, `piChatStream.ts`) hydrates `/state`, tracks cursor continuity, reconnects with jittered backoff.
- Reload mid-stream degrades cleanly: `PI_CHAT_REPLAY_GAP` / `PI_CHAT_CURSOR_AHEAD` → full rehydrate. **Never a silent gap.**

## Dormant
`SqliteEventStreamStore` (`packages/agent/src/server/events/eventStreamStore.ts`) is fully written — schema, idempotent append, offsets, subscribe. The *consumer* is written too: `HarnessPiChatService` accepts an optional `eventStore`, appends durably, resumes `seq` across restarts, and has a poison-on-failed-append policy. **Zero production instantiation** — it appears only in tests. The single production caller, `buildAgentComposition.ts:218-226`, builds the service without `eventStore`.

## Delta to durable (Level D)
1. Wire a `SqlStorage`/`SqliteEventStreamStore` into `buildAgentComposition` → `bridgeOptions.service.eventStore`; add DB path config, lifecycle/close, retention/pruning.
2. **Read path** — the gap that isn't just wiring. Both routes read exclusively from the in-memory buffer, so an old cursor rehydrates even if durable rows exist. `HarnessPiChatService.subscribe` and `EmbeddedAgentGateway.connectSession` must fall back to the store on a buffer miss.
3. Cross-process fan-out: `EventStreamStore.subscribe` is in-process listeners only.
4. Level D conformance suite; `closeStream`/compaction have no production caller.

## Blocker to resolve first
[...]

PR #1017, opened 2026-07-31
docs(direction): record verified lane reality for waves 2-4

hachej:
Stacked on #1008 — `docs/DIRECTION.md` does not exist on `main` yet (it arrived via #974 into the integration branch), so this targets `integration/wave-1-final` and lands with the wave. **Retarget to `main` if #1008 merges first.**

Records what the code actually does for the four remaining lanes, verified against `origin/main` on 2026-07-31 by four independent explorations, so the next orchestrator dispatches from verified state rather than from four separate plan folders.

## What it adds

A **Lane reality** section: a today/delta table per lane, plus the findings that change sequencing —

- **The shared substrate.** BYOK, MCP and sandbox converge on one missing component (per-workspace credentials + registration). Wave 3 now names it as a deliverable instead of letting whichever lane fires first improvise one.
- **Streaming is wiring, not a build** — but the agent-keying omission must be settled before durable rows bake it in.
- **Parked PRs #916/#917 do not close their lanes** — both defer the load-bearing parts by their own bodies.
- **Two sandbox defaults are true today** and belong to no wave: CLI defaults to zero isolation; neither real backend isolates network egress.
- **Two inert surfaces that read as working support**: `mcpServerRefs` is never resolved; MCP ingress is hardcoded off in every host.
- **Bead `0jpy.15` has a false premise** and should be rewritten before anyone works it.

## Also

Threads lane issues #1009/#1010/#1011/#1012 into the wave sections, and records that #820 and #808 are closed — their parked PRs were pointing at closed issues.

Docs only. No code changes.

Refs #1009, #1010, #1011, #1012

### DIFF
--- a/docs/DIRECTION.md
+++ b/docs/DIRECTION.md
@@ -54,9 +54,13 @@ slice absorbed by .27), `.28` (re-land native sessions + rename menu, after
 
 ## Wave 2 — trigger: Wave 1 demo works
 
-- **Chat streaming durability** (`0jpy.8` + `26v`, ONE lane): activate the
-  dormant SqliteEventStreamStore. Trigger is explicit: `.31`'s concurrent
-  streams + Seneca production chat. This is where Seneca hardening lives.
+- **Chat streaming durability** (`0jpy.8` + `26v`, ONE lane; issue #1009):
+  activate the dormant SqliteEventStreamStore. Trigger is explicit: `.31`'s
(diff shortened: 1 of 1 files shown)


hachej:
Lane-reality snapshot superseded by the factory direction docs (docs/factory/) and the per-lane tracking issues (#806, #807, #1081, #1082).

PR #1020, opened 2026-07-31
#909 fix(agent): stabilize runtime scope identity

hachej:
## Summary

- replace path/package/front-build-sensitive runtime pins with a semantic v2 identity over normalized server/runtime declarations and executable bytes
- keep physical placement in the binding/lease identity, separate from persisted session compatibility
- add exact, scoped, fail-closed v1→v2 migration authorization with generation CAS and byte-preserving atomic transcript-header updates
- prepare target bindings before durable migration and gate both public and legacy session mutations before effect admission
- retain fail-closed stale-lock behavior under the documented single-Host writer per session namespace invariant

## Why this blocks Wave 1

Equivalent rebuilds/restarts currently change runtime identity and orphan writable chats. Genuine model/tool/plugin/runtime changes must still remain read-only unless an exact migration is authorized.

## Proof

- Agent runtime identity: 27/27
- Agent Host/session suites: 149 passed, 12 skipped
- Workspace identity/manifest tests: 42/42
- Agent + Workspace typechecks
- Agent invariants
- independent standards review: CLEAN
- independent thermo re-review, including legacy mutation bypass reproduction/regression: CLEAN

## Safety

- migrations are exact agent/workspace/namespace/from/to matches and fail closed
- transcript tails are preserved byte-for-byte; only the authoritative header is changed
- target binding is proven before irreversible pin CAS
- no session effect is admitted before migration succeeds
- `33293674ddb7f24bcc036f4b5bedbf2457ac3a639e2969353ccb0175d385d7fe` remains intentionally unauthorized pending exact historical reconstruction

## Residual operational constraint

Online migration assumes one Host writer per session namespace. Crash-stale migration locks are not broken automatically; they require operator investigation.

### DIFF
--- a/packages/agent/src/server/agent-host/__tests__/legacyAdmissionCompatibility.test.ts
+++ b/packages/agent/src/server/agent-host/__tests__/legacyAdmissionCompatibility.test.ts
@@ -84,6 +84,7 @@ function createFixture(options: {
   readonly failPromptSynchronously?: boolean
   readonly distinctPreboundService?: boolean
   readonly preboundRuntimeScopeIdentity?: string
+  readonly missingSession?: boolean
   readonly migration?: {
     readonly fromIdentity: string
     readonly toIdentity: string
@@ -189,7 +190,7 @@ function createFixture(options: {
     ledger,
     compiledById: new Map([['default', { agentTypeId: 'default', legacyDefault: true }]]),
     options: { resolveRuntimeScope: async () => runtimeScope },
-    resolveSessionRuntime: async () => ({
+    resolveSessionRuntime: async () => options.missingSession ? null : ({
       runtimeScope,
       runtimeScopeIdentity: persistedPin,
       migrateRuntimeScopeIdentity: async () => {
@@ -265,6 +266,15 @@ describe('legacy admitEffect Level-B compatibility', () => {
       'mutation:session.prompt',
       'complete:session.prompt',
     ])
+  })
+
+  it('preserves the frozen legacy session-not-found code before mutation admission', async () => {
+    const fixture = createFixture({ missingSession: true })
+
+    await expect(fixture.compatibility.deleteSession!(context('legacy-missing-http'), 'missing')).rejects.toMatchObject({
+      code: ErrorCode.enum.SESSION_NOT_FOUND,
+    })
+    expect(fixture.events).toEqual([])
   })
 
   it('reuses a pre-resolved compatibility service only when its identity matches the verified binding', async () => {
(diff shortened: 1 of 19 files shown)


hachej:
## Historical `332936…` reconstruction audit

No migration authorization was added.

Evidence recovered:

- Affected session `019fb719-65dd-7551-a5ee-0796158566ba` was created at `2026-07-31T07:35:22.845Z` with stored v1 identity `33293674ddb7f24bcc036f4b5bedbf2457ac3a639e2969353ccb0175d385d7fe`.
- Its native transcript records `openai-codex:gpt-5.3-codex-spark`.
- The original playground log shows the serving process started at 05:55 UTC in local mode with workspace root `.worktrees/smoke-main/apps/workspace-playground/workspace`.
- The `smoke-main` reflog proves that checkout was exactly `150fb3031b53eb7ca7de9a5e57f10823a430dfad` from 05:41:58 until 07:59:22 UTC.
- I rebuilt the complete playground dependency chain from detached commit `150fb3031`, ran the historical fleet/model configuration, and mounted that tree at the original absolute `smoke-main` path for v1 path-sensitive identity assembly.

Result: the reconstructed Researcher identity was `aa1350e788a072e148f6ec1bb204a64c4611058739739600d92aee2c8561ffec`, **not** `332936…`. Other path/build candidates also failed exact equality.

The original server log begins directly at Vite startup rather than a dependency build. The most likely unrecoverable input is the exact heterogeneous set of ignored `dist/**` bytes already present when that long-lived process started; later builds overwrote those artifacts, and v1 incorrectly included such outputs in its digest.

Conclusion: exact original identity has not been reproduced. Keep the session read-only and fail closed; do not authorize `332936…` based on a guessed predecessor.

hachej:
## CI follow-up

The first CI run exposed a frozen legacy-wrapper integration gap: newly created in-memory compatibility sessions were verified through the migration gate, but the verified target binding did not retain affinity to the pre-resolved legacy service, producing 404s in wrapper tests and broad Agent E2E fallout.

Fixed in `2e10d28bd`:

- records the semantic identity proven with each compatibility service binding
- after `bindingForSession` verifies/migrates the pin, reuses that service only when its identity exactly equals the verified target identity
- otherwise executes against the service returned by the verified binding (preserving migration safety)
- records the new session pin after successful legacy creation so subsequent mutations cannot bypass or fail before verification

Proof:

- failing wrapper + compatibility tests: 91/91 passed
- Agent Host/session suite: 150 passed, 12 skipped
- Agent typecheck passed
- full invariants passed
- added explicit regression for matching-identity binding affinity; the migration regression still proves an identity-mismatched prebound service is not used.

hachej:
Second CI follow-up `35951e1a6`: the remaining E2E failure was the frozen legacy delete route receiving addressed `AGENT_SESSION_NOT_FOUND` before reaching the legacy service. The compatibility gateway now translates only that pre-admission addressed miss back to stable legacy `SESSION_NOT_FOUND`; no effect is admitted.

Proof:
- legacy compatibility tests: 10/10
- focused wrapper/integration tests remain green (91/91 from prior follow-up)
- `m3a-sessions.spec.ts`: 4/4 browser E2E
- Agent typecheck green

PR #1128, opened 2026-08-07
[gh-1009 slice 1] wire SqliteEventStreamStore into production (flagged)

hachej:
## What

Wires the already-written-and-tested `SqliteEventStreamStore` into the one production composition funnel (`packages/agent/src/server/agent-host/buildAgentComposition.ts`), behind a flag, so `HarnessPiChatService` can resume live event streams across process restarts.

## Why

`SqliteEventStreamStore` (schema, idempotent append, offsets, subscribe) and `HarnessPiChatService`'s `eventStore` support (resume seq across restarts, poison-on-failed-append) were fully written and tested as of 2026-08-06, but `buildAgentComposition` constructed the service without an `eventStore`, so every durability guard was a permanent no-op in prod. This is Wave-2 slice 1: wire it, flagged, with a restart-resume proof.

## Flag

`BORING_CHAT_DURABLE_STREAM=1` (or `true`). Absent (default) = byte-identical to current behavior — no store constructed, in-memory streaming only.

When set, opens `.agent-event-stream.sqlite` under the host session root (`sessionRoot`, falling back to `workspaceRoot`), following the same durable-volume convention as `BORING_AGENT_SESSION_ROOT` (AGENTS.md hard rule 9) and the existing request-ledger path pattern.

## Fail-closed-ish behavior

If the store can't be opened (bad path/permissions), `openDurableEventStore` reports a new stable error code `EVENT_STORE_OPEN_FAILED` via the telemetry sink and returns `undefined` — composition proceeds with in-memory-only streaming rather than crashing boot, and never silently claims durability. Code documented in `packages/agent/docs/ERROR_CODES.md` and added to the canonical `ErrorCode` enum + its test.

## Proof

`packages/agent/src/server/agent-host/__tests__/buildAgentComposition.durableStream.test.ts`:
- **Durable resume across restart**: append events via one store instance, close it (simulated restart), reopen a second instance over the same on-disk sqlite path, and assert `readEvents` replay-from-cursor reproduces the exact sequence with no gap or duplicate.
- Flag parsing (`1`/`true` enabled, absent/other values disabled).
- `sessionRoot` preferred over `workspaceRoot`, and fallback to `workspaceRoot` when absent.
- `EVENT_STORE_OPEN_FAILED` reported via telemetry (not console) and store construction fails soft.

Ran: `pnpm run typecheck` (clean), `pnpm exec vitest run` for `packages/agent` (all green except one pre-existing full-suite-only timeout in `createAgentHost.test.ts` that passes in isolation — unrelated resource contention, not a regression), and the repo's `scripts/check-invariants.sh` (clean — required moving the diagnostic from `console.error` to the telemetry sink and using `ErrorCode.enum.*` instead of a raw string literal).

## Scope

Kept to one reviewable slice: no changes to NDJSON/cursor semantics or the streaming path beyond what wiring + the fail-closed diagnostic required.

https://claude.ai/code/session_01V1HtxZAQUmLXhaZmL3sEH4


hachej:
Fix round 1 (6c64e3cb4) addresses adversarial review findings:

**H1 (overclaim, fixed for real — chose option a):** durability was write-only. `HarnessPiChatService` only recovered the seq high-water-mark on restart; `PiChatReplayBuffer`'s event backlog was never re-hydrated, so any client resuming behind the tip hit `PI_CHAT_REPLAY_GAP`. Added `hydrateDurableReplayBuffer()`: on cold channel construction it replays up to `DEFAULT_REPLAY_BUFFER_MAX_EVENTS` trailing envelopes from the durable store and publishes them into a fresh buffer before serving `subscribe()`. New PROOF test subscribes at cursor 0 after a simulated restart and asserts the replayed events equal the pre-restart live events — mutation-checked (reverting the hydration fails the test with `replay_gap`).

**H2 (host/guest path bug):** switched the fallback from `runtimeBundle.workspace.root` (an in-sandbox guest path under sandbox mode) to `getOptionalRuntimeBundleStorageRoot` — the same host-resolving accessor already used for `bashRuntimeBundle` in this file. If neither `sessionRoot` nor a host storage root resolves, the store now fails closed (reports `EVENT_STORE_OPEN_FAILED`, no store) instead of ever writing to a guest path.

**H3 (vacuous test):** added `existsSync` assertions proving the db lands under `sessionRoot` and not the fallback root. Mutation-checked: collapsing the `??` fallback now fails two tests.

**M1 (busy_timeout + writer model):** `openDatabase` now sets `PRAGMA busy_timeout=5000` alongside WAL, matching `SqliteAgentRequestLedger`. Documented the writer model in a comment: concurrent runtime bindings for the same workspace scope open the same file as distinct in-process connections serialized by SQLite's file lock — not a multi-process/multi-replica writer setup.

**M2 (misleading comment):** the proof test now exercises `sessionRoot` distinctly from a decoy `hostStorageRoot`, and the comment no longer overclaims it mirrors full service construction.

Re-ran typecheck, `scripts/check-invariants.sh`, and the full `packages/agent` vitest suite — all green except one pre-existing full-suite-only timeout in `createAgentHost.test.ts`, confirmed unrelated (passes in isolation).

hachej:
Concurrency fix landed (5bedbed0b) — this was the owner's residual hold item.

**Root cause confirmed:** \`busy_timeout=5000\` does not cover a read-then-write transaction's lock upgrade. \`runTransaction\` issued a plain deferred \`BEGIN\`, and \`appendEventOnce\`/\`appendAgentEvent\`/\`appendEvent\` read (SELECT) then write (INSERT/UPDATE) inside that same transaction. SQLite refuses that upgrade with \`SQLITE_BUSY\` ("database is locked") immediately when another connection holds the write lock, without honoring \`busy_timeout\` — under concurrent appends from two runtime bindings over the same file this reached \`publishChannelEvents\`' poison policy and killed a live channel.

**Fix:** added \`SqlTransactionMode\` (\`'deferred' | 'immediate'\`) to \`runTransaction\`/\`RunTransaction\` in \`sqlStorage.ts\` — \`deferred\` stays the default for any future read-only callers (not a blanket change), and all three of \`SqliteEventStreamStore\`'s write paths (\`appendEvent\`, \`appendEventOnce\`, \`appendAgentEvent\`) now explicitly request \`'immediate'\`, matching \`SqliteAgentRequestLedger\`'s model — acquiring the write lock up front so \`busy_timeout\` applies normally instead of hitting the non-retried upgrade case.

**Proof:** two real \`node:worker_threads\` workers, each with its own \`node:sqlite\` connection to the SAME on-disk file, run the actual \`runTransaction\` primitive with the exact read-then-write shape of \`allocateSeq\`/\`readIdempotencyKey\` — genuine OS thread concurrency, not simulated. With \`'immediate'\`: 120/120 commits succeed across both workers, zero errors surface. With \`'deferred'\` (the bug): reliably reproduces "database is locked" — proving the immediate-mode test is non-vacuous. A second wiring test spies on \`runTransaction\` to assert the three real call sites in \`eventStreamStore.ts\` pass \`'immediate'\`; mutation-checked (reverting those call sites fails it).

**Also fixed while in there (LOW):** investigated the claimed \`hydrateDurableReplayBuffer\` off-by-one. Verified empirically that \`meta.nextOffset\`/\`tailIndex\` is already the tail entry's own seq (not one-past-the-end despite the field's name), so the existing window formula returns exactly 1000 entries — no functional bug existed. Added a permanent boundary regression test plus a clarifying comment rather than changing already-correct behavior (which would have introduced the very bug being described).

Verification: typecheck clean, \`scripts/check-invariants.sh\` clean, full \`packages/agent\` vitest suite green (same pre-existing unrelated full-suite-only timeout as prior rounds).

PR #1132, opened 2026-08-07
[16f.2] BYOK credential vault: KmsBackend abstraction + local-KEK backend

hachej:
Implements bead `wt-391-forward-16f.2` — the keystone slice of the ratified BYOK design in `docs/issues/820/byok-secret-vault-plan.md`. Builds the pluggable `KmsBackend` abstraction, a local-KEK implementation, AAD-bound AES-256-GCM field crypto, and a real `CredentialStoreBackendV1` that makes the previously-unconstructible frozen credential contract usable end to end.

## (a) Design choices

**KEK source.** Per the plan's "Explicit local-KEK/self-host provider" section, the KEK is a 32-byte key generated *outside* the app and stored *outside* Postgres, in an operator-owned sealed file mounted read-only. Selection is explicit, immutable startup config:

| Env var | Meaning |
| --- | --- |
| `BORING_CREDENTIAL_KMS_BACKEND` | must be exactly `local-kek` to select this backend; anything else yields `undefined` (host must fail closed, not substitute a default) |
| `BORING_CREDENTIAL_LOCAL_KEK_FILE` | path to the sealed KEK file — **required**, no default |
| `BORING_CREDENTIAL_LOCAL_KEK_REF` | opaque non-secret key reference (default `default`) |
| `BORING_CREDENTIAL_LOCAL_KEK_VERSION` | KEK version for rotation (default `1`) |

There is deliberately **no plaintext-env key**: the key never travels in an env var, only a path does. `WORKSPACE_SETTINGS_ENCRYPTION_KEY` is never read (a test asserts it is ignored even when present). The file may hold 32 raw bytes, 64 hex chars, or 44 base64 chars; anything else fails closed. `resolveLocalKekProviderConfigV1` takes the env record as an argument — this package forbids `process.env` reads outside `src/server/config/**`.

**File layout.**

- `packages/agent/src/shared/credentials/kmsBackend.ts` — the ratified backend-agnostic contract: `WorkspaceKekProviderV1`, `WrappedWorkspaceDekV1`/`WrappedWorkspaceDekPayloadV1` (with the `vault-transit-ciphertext.v1` and `external-kms-opaque.v1` variants *reserved, unimplemented*), `CredentialEnvelopeV1`, `CredentialFieldAadContextV1`, and the fixed AES-256-GCM size constants. `Uint8Array`-only; no `node:*` import and no Node byte-array type.
- `packages/agent/src/server/credentials/vault/kmsBackend.ts` — local-KEK implementation + env/file config resolution.
- `packages/agent/src/server/credentials/vault/envelopeCrypto.ts` — field AES-256-GCM + canonical AAD encoding + `timingSafeEqual` helper.
- `packages/agent/src/server/credentials/vault/persistence.ts` — the injectable `CredentialVaultPersistenceV1` port (`get/putCredentialRecord`, `get/putWrappedDek`, `get/putField`), shaped one-to-one onto the plan's proposed tables.
- `packages/agent/src/server/credentials/vault/inMemoryPersistence.ts` — a real in-memory implementation of that port for tests/dev.
- `packages/agent/src/server/credentials/vault/vaultStoreBackend.ts` — `createVaultCredentialStoreBackendV1({ kmsBackend, persistence })`, satisfying the existing `CredentialStoreBackendV1`.

**Crypto.** One 256-bit DEK per workspace generation. Field AAD is a canonical, versioned, **length-prefixed** encoding of `workspaceId:credentialId:providerId:fieldId:credentialVersion:dekGeneration` — length prefixing makes it unambiguous, so `("a","bc")` can never collide with `("ab","c")` (asserted by a test). `setAAD()` before encryption, `setAAD()` + `setAuthTag()` before finalization. Fresh `crypto.randomBytes(12)` nonce every encryption; every write mints a new credential version, so no nonce is ever reused across a rewrite. Persisted AAD is recomputed and compared with `crypto.timingSafeEqual` before tag verification. No ECB/CBC, no `createCipher`, no `Math.random`, no `==` on secret state. Plaintext DEKs and field buffers are `fill(0)`'d in `finally` — documented in-code as best-effort only, matching the plan's "Process-memory honesty" section (Node/V8 cannot guarantee erasure).

**Error codes.** Reused `CREDENTIAL_NOT_CONFIGURED`, `CREDENTIAL_SCHEMA_MISMATCH`, `CREDENTIAL_UNREADABLE` (all authentication failures — never "absent"), and `CREDENTIAL_BACKEND_UNAVAILABLE` (readiness not-ready). Added exactly one code: **`CREDENTIAL_KEK_UNAVAILABLE`**, because a missing/unreadable/wrong-length KEK is an operationally distinct, non-retryable fail-closed condition rather than a transient backend outage.

`docs/ERROR_CODES.md` previously had no `CREDENTIAL_*` rows and its parity test bound the whole file to the `ErrorCode` enum, so appending credential rows broke it. Rather than weaken the check, this PR gives credential codes their own **`## Credential registry`** section, documents **all sixteen** `CREDENTIAL_ERROR_CODES` values, and makes the parity test section-scoped with a **second parity assertion** for the credential enum. Both registries are now independently enforced.

## (b) Deliberately deferred

- OVH KMS, Scaleway, Exoscale, and Vault/OpenBao Transit backends. Open Question #1 in `docs/issues/1082/plan.md` (which backend ships as default) is still open; this slice builds only the abstraction plus local-KEK, and the reserved payload variants sit unimplemented behind the union.
- Postgres schema/migration for credential tables. The plan itself notes the crypto-conformance surface alone exceeds one review budget; persistence lands here as an injectable port so the schema pass can drop in behind it. No Drizzle changes.
- Fleet-loader wiring (`loadConfiguredAgentFleet.ts`), onboarding/settings UI, MCP generalization, and the `WORKSPACE_SETTINGS_ENCRYPTION_KEY` migration — all untouched.
- `hostResolver.ts` / `withResolvedCredential.ts` contracts are unchanged; the in-memory fake backend is kept intact for existing tests.
- `external-managed-account` material is not yet vault-persisted (only `field-set` and `none`); it needs its own opaque-reference contract pass.

## (c) Proof

New: `packages/agent/src/server/credentials/vault/__tests__/vault.test.ts` — 35 tests, all passing.

*Round-trip:* DEK wrap/unwrap through local-KEK; field encrypt/decrypt (12-byte nonce, 16-byte tag, ciphertext contains no plaintext); full store-backend write→read round-trip; persistence holds ciphertext only.

*Wrong-KEK:* unwrapping a KEK-A-wrapped DEK under KEK B → `CREDENTIAL_UNREADABLE`, no fallback; the same at store level via a shared persistence with a KEK-B reader; a foreign backend's `external-kms-opaque.v1` envelope is refused rather than opportunistically read.

*Tamper detection:* parameterized over `ciphertext` / `nonce` / `authTag` / `aadContext` (each flipped independently) and over each AAD component swapped independently — `workspaceId`, `credentialId`, `providerId`, `fieldId`, `credentialVersion`, `dekGeneration`. All twelve fail closed with `CREDENTIAL_UNREADABLE`, never "absent". Also: a wrapped DEK replayed into another workspace or another generation, a v1 field ciphertext spliced under the v2 credential identity, and decryption under a different DEK.

*No-secret-in-error:* `expectNoSecretLeak` asserts `.message`, `String(error)`, `.stack`, and a JSON serialization contain none of the plaintext field value, its hex/base64 encodings, or either KEK's hex/base64. It runs on every thrown-error assertion in the file.

*Per-workspace isolation:* workspace A's wrapped DEK, field envelope, and credential record are copied verbatim into workspace B's identity; B's read fails with `CREDENTIAL_UNREADABLE` — an authentication failure, explicitly not a `NOT_CONFIGURED` masking a bypass.

*Readiness:* a missing KEK file reports `{ ready: false, reasonCode: 'local-kek-source-unreadable' }` without throwing and never yields a default key; a wrong-length file reports `local-kek-invalid-length`; credential operations under that provider fail with `CREDENTIAL_KEK_UNAVAILABLE`, and the store-level read fails with `CREDENTIAL_BACKEND_UNAVAILABLE`. Config resolution returns `undefined` when unselected, throws when selected without a file path, and ignores a `WORKSPACE_SETTINGS_ENCRYPTION_KEY` present in env.

*Integration (the key proof):* `frozen credential contract, constructed end to end` builds the real provider registry + consumer binding registry + ref factory + `createHostSideCredentialResolverV1` on top of the vault-backed backend, and round-trips a credential through `withResolvedCredential` — the frozen contract is now actually constructible and usable. A second case shows a different workspace resolving the same binding fails closed.

Also added: `ERROR_CODES.md documents every credential error code` parity test.

**Commands run** (from the worktree):
- `pnpm -C packages/agent test` → **1934 passed, 1 failed, 17 skipped**. The single failure is `createAgentHost.test.ts > durably replays insufficient-credit rejection…` hitting the 5s default timeout under full-suite parallel load; that file passes 12/12 in isolation and is untouched by this PR.
- `pnpm -C packages/agent typecheck` (`tsc --noEmit`) → clean.
- `pnpm -C packages/agent lint:invariants` → all invariants pass (no `node:*`/Node byte types in `src/shared/**`, no `console.*` in `src/server/**`, no `process.env` outside `src/server/config/**`).

## (d) Friction

1. The worktree had no `node_modules`; `pnpm install --frozen-lockfile` plus a `@hachej/boring-ui-kit` build were needed before the front-end test files would transform (23 unrelated failures until then).
2. The `ERROR_CODES.md` parity test bound the entire document to one enum, so registering credential codes required the section-scoping change described above. This touches a test outside the credential surface — flagged for review, but it strengthens rather than relaxes the check.
3. `hostResolver.ts`'s `CredentialStoreBackendV1.read()` receives no credential version, so the backend owns version selection; the vault store therefore keeps an explicit credential record (`credentialId`, `credentialVersion`, `dekGeneration`, `materialKind`) per workspace/provider. `credentialId` is a stable per-record UUID, since the frozen registry has no such identifier and the plan's AAD requires one.
4. No deviations from the ratified design were found; no conflict-stop was triggered.

https://claude.ai/code/session_01V1HtxZAQUmLXhaZmL3sEH4


hachej:
Owner ratified the key-scope decision (PR #1137 merged): per-workspace DEK stands; seat/tier differentiation, if ever needed, ships as credential profiles under the same DEK. Per the gate note in #1137's body, merging this implementation PR alongside.

Issue #1082, opened 2026-08-05
Epic: BYOK tenant keys — salvage 16f.2 vault storage (KMS backend)


hachej:
Post-graduation factory epic. Resurrects the missing middle of the BYOK sequence (#820 closed; 16f.1 + 16f.6 on main). Salvage branch `feat/820-vault-storage` (was PR #917, CI-green pre-conflict) via rebase when scheduled. Absorbs the [lane] BYOK anchor (PR #1014). Needed for Seneca-style tenants bringing their own model keys.

hachej:
Folded into PR #1409 (programs.md). Front door: stacked PR pair #1145 → #1164; merge S1 first, then retarget slice B.

PR #1145, opened 2026-08-07
[BYOK Credential Persistence] Preserve audited lineage after #1539 (gh-1082)

hachej:
Bead: `wt-391-forward-wf9n.8`  
Issue: #1082

## Summary
- persist encrypted BYOK credentials behind workspace authority without plaintext fallback
- preserve distinct V1 compatibility and V2 deferred anchor/atomic persistence contracts
- recover DB→anchor crashes only when durable state exactly matches an authenticated pending intent
- authenticate the complete committed recovery state, including `displayLabel` and `maskedLastFourSuffix`
- preserve KEK fail-closed behavior, rollback detection, lifecycle checks, DEK rotation, and current-main/#1539 reconciliation

## Current status

**READY FOR ORCHESTRATOR GATE; WORKER DID NOT MERGE.** Exact pushed head `62a096f8ad62a935219f0f044e320beec0e50b23` is bound to the proof and artifact below. Exact-head CI and Workflow Invariants are green. Fresh independent review grants explicit **SECURITY PASS** and **CROSS-PACKAGE ABSTRACTION PASS** with no material findings.

## Proof of work

- Current main integrated: `8690fd82ff278ab447585c52c14398316971f8d9` via merge `a305308975383656fc2fd809c82b86ef5127aa30`; `origin/main` has not advanced.
- Exact head: `62a096f8ad62a935219f0f044e320beec0e50b23`.
- Exact-SHA sandbox `a9874ef4-4279-4378-878b-d6534ab9c0f4`: 4 credential files / 118 tests PASS; Agent typecheck PASS; Core typecheck PASS; full `pnpm lint:invariants` PASS.
- Tracker parser at exact head: 647/647 unique issue IDs; 1241/1241 unique nested comment IDs; every nested `issue_id` matches its parent. Duplicate `factory-plugin-z4hs.1` comment IDs `1219`/`1220` were semantically renumbered to `1256`/`1257`; text, authorship, timestamps, parent and history are retained.
- [CI 34371243126 — PASS at exact head](https://github.com/hachej/boring-ui/actions/runs/34371243126)
- [Workflow Invariants 34371243185 — PASS at exact head](https://github.com/hachej/boring-ui/actions/runs/34371243185)
- PR: https://github.com/hachej/boring-ui/pull/1145 — exact head; MERGEABLE/CLEAN read-back required before owner action.

## Independent review

- Reviewer: `boring-reviewer`; model `openai-codex/gpt-5.6-sol`; session `827db93e-4e0c-4c01-910c-56eda0d4f455`.
- Target: `62a096f8ad62a935219f0f044e320beec0e50b23`; mandate digest `sha256:098ee3aea46536fd6b51a119071eb851610295593979be7c7860e663a08bb04c`; round 2; capReached false.
- Verdict: APPROVE; explicit **SECURITY PASS** and **CROSS-PACKAGE ABSTRACTION PASS**; no material findings.
- Round 1 (`8ed140c2-86af-46cc-bccb-fec11987e55c`, digest `sha256:7935033bfac2edbd178a3bc30115c209412934d33490a303a8931b5bfccb9570`) granted both technical passes but requested restoration of three unrelated historical tracker IDs; fixed forward and re-reviewed.

## Security and compatibility guarantees

The recovery digest now uses a version-tagged, fixed-shape JSON tuple containing record identity/version/generation/material kind, lifecycle state, credential type, display label, masked suffix, and sorted/base64 envelope fields. JSON escaping plus fixed object keys and explicit `null` values make boundaries and absence unambiguous. Two adversarial crash tests produce exact-after durable state, tamper only the label or suffix, and prove recovery returns `CREDENTIAL_UNREADABLE` while leaving the pending intent unfinalized. Plaintext credentials and KEKs never enter the anchor.

V1 behavior remains distinct from V2 deferred commit and atomic metadata persistence. Exact-before ambiguity, rollback, malformed KEK, lifecycle mismatch, and arbitrary state mismatch remain fail-stop. The current-main merge did not duplicate #1539 or alter authority boundaries.

## Risk / rollback

Protected triggers: credentials/secrets authority, public versioned contracts, and 861 changed production lines under `packages/` before this focused repair (tests/config/generated excluded). Owner merge approval remains mandatory and belongs to the Orchestrator. No waiver. Roll back only with new revert commits, newest first; never rewrite or force-push. Reverting the recovery lineage restores the known DB-commit/anchor-EIO defect and is not a safe delivery candidate.

## Owner Review

Bead / PR / issue: `wt-391-forward-wf9n.8` / #1145 / #1082  
What changed / why: Complete credential metadata is now authenticated during crash recovery; semantic tracker comment collisions and the revision-bound handoff were repaired.  
Why you are needed: credentials/secrets authority and shared V2 persistence/anchor contracts are protected boundaries.  
Recommendation: approve this exact revision; final independent SECURITY and CROSS-PACKAGE ABSTRACTION passes are recorded above.  
Scope / base / head: current main `8690fd82ff278ab447585c52c14398316971f8d9`; head `62a096f8ad62a935219f0f044e320beec0e50b23`.  
What approval authorizes: merge of this exact revision only; no deploy/publish authority.  
Risk / rollback: fail-closed credential availability on mismatch; revert with new commits only.  
Proof: exact sandbox and clickable exact-head CI links above; presentation `.handoff/pr-1145-handoff.html`.  
Please test: inspect the two exact-after metadata tamper cases and confirm the HTML displays the full head SHA and both exact CI links.  
Decision: approve | request changes | defer | reject

## Show me

```diff
 DB commit -> pending anchor recovery
   fixed-shape versioned canonical tuple
   record + lifecycle + credential type + sorted encrypted fields
+  displayLabel
+  maskedLastFourSuffix
   exact-after match -> finalize anchor
+  label/suffix tamper -> CREDENTIAL_UNREADABLE; pending remains

 semantic tracker
-  nested IDs 1219/1220 reused across two issues
+  original Gateway Retry history retains 1219/1220
+  later factory-plugin comments retain history as 1256/1257
+  647/647 issue IDs and 1241/1241 comment IDs unique

 owner handoff
-  stale artifact without full revision; incomplete body headings
+  full 62a096f8ad62a935219f0f044e320beec0e50b23
+  clickable exact-head CI and Workflow Invariants links
+  Owner Review / Show me / Handover sections
```

```mermaid
sequenceDiagram
  participant DB as Credential persistence V2
  participant Anchor as Authenticated version anchor
  participant Recovery as Startup/read recovery
  DB->>Anchor: pending intent(beforeDigest, afterDigest)
  DB->>DB: atomic record + metadata + envelopes
  Note over DB: crash before final anchor write
  Recovery->>DB: canonical durable digest (includes label + suffix)
  alt exact afterDigest
    Recovery->>Anchor: finalize committed version
  else exact beforeDigest or any tamper/mismatch
    Recovery-->>Recovery: fail closed; do not finalize
  end
```

## Handover

Exact candidate: `62a096f8ad62a935219f0f044e320beec0e50b23`. Worker has pushed normally and has not merged, force-pushed, asked the owner, or granted approval. Final independent exact-SHA security/abstraction review is recorded; Orchestrator owns Gate 2 and the owner decision. Artifact: `.handoff/pr-1145-handoff.html`.

Refs #1082


hachej:
Proof of work — `196d22bd9`

What changed:
- Postgres credential-vault persistence, migrations, deletion/tombstones
- external local-KEK version/state anchor with snapshot-replay rejection
- in-memory + Postgres parameterized conformance and rollback tests

Automated verification:
- `pnpm --filter @hachej/boring-agent run typecheck` ✅
- `pnpm --filter @hachej/boring-agent exec vitest run src/server/credentials/vault/__tests__/vault.test.ts src/server/credentials/vault/__tests__/postgresPersistence.test.ts` ✅ — 57 tests

Screenshot/demo: N/A (server persistence/security slice)

Waiver / known gaps:
- S1 narrows replay exposure; S2 closes exfiltrated-ciphertext exposure with DEK rotation/destruction.
- Cross-store commit/anchor failure deliberately fail-stops rather than accepting attacker-writable DB-ahead state.

hachej:
Proof refresh — `1118a123d`

Final fail-closed classification now treats a missing envelope under an externally authenticated current record as `CREDENTIAL_UNREADABLE`, not `NOT_CONFIGURED`.

Automated verification at this head:
- `pnpm --filter @hachej/boring-agent run typecheck` ✅
- `pnpm --filter @hachej/boring-agent exec vitest run src/server/credentials/vault/__tests__/vault.test.ts src/server/credentials/vault/__tests__/postgresPersistence.test.ts` ✅ — 57 tests

hachej:
Final proof refresh — `02c5c626b`

Automated verification at final head:
- `pnpm --filter @hachej/boring-agent run typecheck` ✅
- `pnpm --filter @hachej/boring-agent exec vitest run src/server/credentials/vault/__tests__/vault.test.ts src/server/credentials/vault/__tests__/postgresPersistence.test.ts` ✅ — 57 tests

Local anchor mutations now serialize with bounded retry and stable fail-closed backend-unavailable behavior under lock contention.

hachej:
Final proof — `ac09c1487`

Automated verification:
- `pnpm --filter @hachej/boring-agent run typecheck` ✅
- `pnpm --filter @hachej/boring-agent exec vitest run src/server/credentials/vault/__tests__/vault.test.ts src/server/credentials/vault/__tests__/postgresPersistence.test.ts` ✅ — 57 tests

Schema proof now includes the ratified credential metadata/state columns, bounded safe-integer generations/versions, field→wrapped-DEK generation FK, and metadata-only tombstone constraints.

hachej:

Not because the work is bad — the crypto hygiene and the Postgres adapter are solid and should survive — but because the version anchor as implemented introduces two unrecoverable failure modes and one crash-persistent outage into the credential path.

**Stack context:** #1164 is stacked on this branch (its base *is* `feat/1082-s1-persistence`) and cannot merge first. GitHub reports #1164 CLEAN only because it's measured against this base, which is itself DIRTY — the classic stacked-PR merge trap. **Main has NOT superseded either PR:** `origin/main:packages/agent/src/server/credentials/` still contains only the 16f.1/16f.2 surface — no `versionAnchor.ts`, no `postgresPersistence.ts`, no `startupComposition.ts`. Conflicts vs current main are just `packages/agent/package.json`.

### Correctness

**C1 (BLOCKER) — a single crash between the DB commit and the anchor write permanently bricks the credential, with no recovery path shipped.** `vaultStoreBackend.ts:252` wraps the write in `versionAnchor.withMutation`. Inside the callback the Postgres transaction commits (`commitCredentialVersion`), and only *after* the callback returns does the anchor persist (`versionAnchor.ts:389 await replaceSealedState(next, options)`). DB-first, anchor-second. If `replaceSealedState` fails — disk full, EIO, SIGKILL between `rename` and `fsync`, container eviction — Postgres is at v2 and the anchor at v1. From that moment **every read fails** (`vaultStoreBackend.ts:177 requireCurrentVersion` → `:143 unreadable('Credential current state failed rollback verification')`) **and every write fails too** (`:265` applies the identical anchored-vs-existing check before it can advance). The credential is unreadable *and* unrepairable through the shipped API; the only fix is hand-editing a HMAC-sealed file, requiring the operator to independently re-derive the MAC with the KEK. No CLI, no repair function, no runbook in this PR. The body frames this as an accepted tradeoff ("does not claim cross-store crash atomicity") — but accepting a fail-stop is only defensible when a recovery path exists. As shipped it's a permanent-data-loss-equivalent bug for the user's key, triggered by an ordinary crash rather than an attack.

*Standard fix:* two-phase anchor — write a *pending* `{provider, nextVersion}` intent **before** the DB commit, commit, then finalize. A crash leaves a marker unambiguously resolvable by reading the DB (pending==DB → finalize; pending>DB → discard), and rollback protection is preserved because a replayed DB can never satisfy a pending-or-committed anchor at a higher version. Minimum acceptable alternative: an explicit KEK-authenticated `reanchorFromDatabase` operator function gated on deliberate intent, plus a runbook.

**C2 (BLOCKER) — the design is single-host-only, and nothing says so, enforces it, or tests it.** The anchor is a **local file** (`versionAnchor.ts:349` docstring: "operator-controlled durable mount beside the KEK"); the persistence it guards is **shared Postgres**. Incompatible the moment more than one host process runs against the same database — which is the entire reason to choose Postgres. Host A writes v2 → shared DB v2, A's anchor v2. Host B's anchor is still v1 → **every** read on B fails closed at `vaultStoreBackend.ts:177`, and every write fails at `:265`. Not degraded — bricked, per C1, since B can never advance past a DB that is already ahead. Sharing one anchor file over NFS/EFS doesn't rescue it: `acquireMutationLock` (`versionAnchor.ts:68-71`) relies on `open(path,'wx')` exclusive-create, whose atomicity is not guaranteed on network filesystems. There is no assertion, no startup check, no doc note, and no test covering two anchors over one database.

**C3 (MAJOR) — the mutation lock is a crash-persistent global write outage.** `acquireMutationLock` (`versionAnchor.ts:68-80`) creates `${anchorFilePath}.lock` with `wx`, retries 100× at 25ms, then throws `BACKEND_UNAVAILABLE`. No PID check, no mtime staleness break, no `flock`. The code's own comment (`:367-369`) states the consequence: "A process crash can leave the lock behind. Future writes then fail closed until operator cleanup." That's a permanent write outage after any ungraceful exit, again with no tooling. `flock`/`fcntl` advisory locks release automatically on process death and are the correct primitive; an mtime-TTL stale-break is the acceptable fallback. Separately, this lock is **global across all workspaces** — the path derives from `anchorFilePath` alone — so one tenant's slow Postgres commit serializes (and after 2.5s errors) every other tenant's credential write. Make it per-workspace; one-line change.

**C4 (MINOR) — wrong error type on a defensive path.** `versionAnchor.ts:230-236`: when `loadKek()` returns a non-`Uint8Array`, the guard runs `material?.fill(0)` *before* `unreadable(...)`. If the value is a string or object, `.fill` isn't a function and a raw `TypeError` escapes instead of `CredentialResolutionError`, breaking the stable `CREDENTIAL_*` code contract the module otherwise honors. Validate, throw, and only zero when it's actually a typed array.

**C5 (MINOR) — `dekGeneration` is threaded into the read key but silently ignored by the adapter.** `vaultStoreBackend.ts:205` passes `dekGeneration: record.dekGeneration` into `persistence.getField`, but the Postgres `getField` SELECT filters only on `(workspace_id, provider_id, credential_version, field_id, deleted_at IS NULL)`. The real defense is the AAD binding (which does cover `dekGeneration`), so this isn't exploitable — it's a parameter that reads as a security check and isn't one. Filter on it or drop it.

**C6 (MINOR) — `putWrappedDek` escapes the commit transaction.** In `writeCredentialFields` the new-DEK path calls `persistence.putWrappedDek(...)` on the *outer* persistence before `commitCredentialVersion` opens its transaction, so a failed commit leaves an orphan `workspace_credential_keys` row. Harmless today (idempotent upsert, FK still satisfiable), but the field→key FK is satisfied by a non-atomic prior write — which will matter for S2's DEK destruction.

**C7 (MINOR)** — `tombstoneCredentialVersionFields` (Postgres) has no `WHERE … AND deleted_at IS NULL` guard, so re-tombstoning rewrites `deleted_at`/`deletion_reason` on already-tombstoned rows. Forensic accuracy only.

**C8 — the migration is never called from any deployment path.** `runCredentialVaultPostgresMigrationsV1` is referenced only by `postgresPersistence.test.ts` and the barrel exports (verified across the whole branch), so the tables do not exist in any real deployment. The repo already has the right seam: `packages/core/src/server/db/migrate.ts` exposes `additionalMigrations: Array<(sql) => Promise<void>>` ("Explicit app-owned migrations run under the same deployment lock") — this function's signature already matches. Register it there before or with the slice that first writes a credential. Worth separately verifying that raw DDL tables absent from `packages/core/src/server/db/schema.ts` are not at risk from `drizzle-kit push` workflows.

### Secret handling — this part is good

Genuinely careful, and not boilerplate praise: KEK material zeroed on every path including the error path (`versionAnchor.ts:238-241` finally-block); derived MAC key zeroed after seal and after verify (`:265`, `:284`); MAC comparison length-checked then `timingSafeEqual` (`:281`); anchor MAC key HKDF-separated from the KEK's encryption use (`ANCHOR_MAC_INFO_V1`), not the raw KEK; the anchor stores only versions and material *kinds* — no ciphertext, no ids, no secrets — canonicalized (sorted keys) before MAC; `plaintextDek?.fill(0)` in a `finally` covering both unwrap and generate branches; tombstones NULL the ciphertext columns and the DB `CHECK` constraint (`workspace_provider_credential_fields_tombstone_check`) makes "tombstoned but still holding ciphertext" *unrepresentable*, with resurrection blocked at the SQL level via `ON CONFLICT … DO UPDATE … WHERE deleted_at IS NULL` plus a `result.count === 0` check. No secret material in any error message, and tests assert it.

### Over-engineering — is the anchor worth it?

Partly. The AAD binding already on main covers `workspaceId, credentialId, providerId, fieldId, credentialVersion, dekGeneration`, so a DB-write adversary already cannot splice one provider's ciphertext onto another, resurrect an old version's ciphertext under the current identity, or roll back `dek_generation`/`credential_id` alone — all fail decryption. The anchor's marginal value is exactly one scenario: a **consistent whole-database rollback to an earlier point in time** by an adversary who can write Postgres but cannot read the KEK-adjacent local file. That's the scenario `postgresPersistence.test.ts` "fails closed after a wholesale old-version database snapshot replay" pins, and it's a legitimate threat (an operator restoring a backup reinstates an API key the tenant revoked because it leaked). But the anchor is MAC'd with the same KEK that wraps the DEKs, so anyone with the KEK can forge it.

Against that narrow gain: a new durable store with its own lock, MAC, fsync and crash semantics; a global write serialization point; an unrecoverable brick mode (C1); an architectural incompatibility with the multi-host deployment Postgres implies (C2); a permanent write outage after any crash (C3); and a per-read disk read + HKDF + HMAC over the *whole* anchor state, in the hot path of every model call, growing O(tenants). **The threat is real and the concept defensible, but the implementation is not proportionate to it** — as it stands the control is more likely to cause a credential outage through ordinary operations than to block an actual attacker.

Secondary, smaller: `postgresMigrations.ts` declares `display_label`, `credential_type`, `credential_schema_version`, `masked_last_four_suffix`, and a six-value `state` CHECK, while `putCredentialRecord` writes `display_label = providerId`, `credential_type = 'field-set.v1'`, and can only ever emit two of the six states — four speculative columns, one actively lying. Similarly `getWrappedDek` implements decode branches for `vault-transit-ciphertext.v1` and `external-kms-opaque.v1` when only `local-kek` exists. Trim to what is written, or name the bead that will fill them.

### BS check

**The PR body is honest.** It explicitly scopes what S1 does *not* claim (cross-store crash atomicity; making exfiltrated ciphertext dead — deferred to S2) rather than overselling. The verification section reports real commands. The Postgres tests genuinely ran against a real Postgres in CI (`ci.yml:232-243` matches the test's hardcoded default) rather than being silently skipped — the 57-test conformance run is real, and the conformance suites really run against both adapters. No fake tests, no weakened assertions.

One caveat: the conformance test named "atomically replaces a version with metadata-only tombstones" asserts only the *end state*, never a failure-injected rollback, so it does not actually test atomicity. The Postgres adapter gets atomicity from a genuine transaction so the property holds — but the name overstates what the test proves, and the in-memory adapter's `commitCredentialVersion` is *not* atomic (awaits interleave between the field writes, the record write, and the tombstone call). Rename, or add failure injection.

### Required before merge

1. **C1** — two-phase anchor (pending intent before DB commit, finalize after), or a shipped, documented, KEK-authenticated re-anchor recovery operation. Test the crash window explicitly.
2. **C2** — decide and enforce: either declare single-host in config validation + deployment docs + a test, or move the anchor to a store both hosts share.
3. **C3** — replace the `wx` lock with `flock`/`fcntl` (auto-released on process death) or add an mtime-based stale-break; make the lock per workspace.
4. **C4** — fix the `material?.fill(0)` ordering so the path throws `CredentialResolutionError`.
5. **C8** — register `runCredentialVaultPostgresMigrationsV1` via core's `additionalMigrations`, or state which bead does it and gate that bead on this one.
6. Rebase onto current main (only `packages/agent/package.json` conflicts).

C5/C6/C7 and the speculative-column trim are follow-ups, not gates.

**Owner decision needed (spans this PR and #1164):** if the BYOK epic is not being actively driven, the honest move is to leave both open with this review attached rather than merge an unreachable, unprovisionable credential vault into main. Neither PR gives an operator any way to *provision* the anchor today (see #1164 D2), so a fully-configured deployment still fails every credential operation.

hachej:
## 🌙 Overnight review — codex Sol cross-check (adversarial second pass)

An independent cross-model pass was run after the Opus review above. It **confirms all three Opus blockers** (DB-first/anchor-second crash window, single-host-only anchor vs shared Postgres, crash-persistent lock) and adds a new one that is arguably worse — a concurrent-first-write DEK race that silently produces an undecryptable credential.

### Blockers

1. **Concurrent first writes can permanently corrupt a credential.**

   Evidence: `vaultStoreBackend.ts:275-290` checks/generates and upserts the wrapped DEK before the credential transaction. `postgresPersistence.ts:251-274` overwrites `(workspace_id, dek_generation)` on conflict. The advisory lock is acquired only later at `postgresPersistence.ts:121-169`.

   Two hosts can generate different DEKs. Host B overwrites the wrapped DEK, while host A wins the credential CAS with fields encrypted under A’s DEK. The committed credential is then undecryptable.

2. **The acknowledged cross-store crash window is an ordinary availability failure, not acceptable rollback protection.**

   Evidence: the DB commit happens inside `versionAnchor.withMutation` at `vaultStoreBackend.ts:316-331`; the anchor is persisted only after the callback returns at `versionAnchor.ts:371-390`. Verification then permanently rejects DB version N against anchor N−1 at `vaultStoreBackend.ts:133-145`.

   A crash, rename failure, or ambiguous directory-fsync result after DB commit disables all reads and subsequent writes for that credential. There is no reconciliation protocol.

3. **The local-file anchor is incompatible with shared Postgres across hosts.**

   Evidence: production local-KEK composition always creates a file anchor at `kmsBackend.ts:479-487`. Serialization is only `<anchor>.lock` on that filesystem at `versionAnchor.ts:364-365`.

   Separate hosts therefore maintain divergent anchors. A write on one host makes another host reject the shared DB state. Concurrent hosts also trigger the DEK race above.

4. **A crashed writer disables credential mutation indefinitely.**

   Evidence: `versionAnchor.ts:68-87` retries an existing lock for roughly 2.5 seconds; it records no owner, lease, or age. `versionAnchor.ts:367-368` explicitly requires operator cleanup after a process crash.

5. **The anchor is authenticated, but not intrinsically rollback-proof.**

   Evidence: `versionAnchor.ts:245-287` uses an HMAC over a rewritable JSON file. An old valid anchor plus the matching old database snapshot passes verification.

   This is sound only for the narrow threat model “attacker can roll back Postgres but cannot read, write, snapshot, or restore the anchor mount.” It is not a monotonic anchor against host-volume rollback.

6. **No deployable provisioning path exists.**

   `initializeLocalFileCredentialVersionAnchorV1`, the Postgres migrations, and the Postgres backend are only defined/exported or invoked by tests. A missing anchor fails at `versionAnchor.ts:264-272`.

### Additional missed issues

- **Global bottleneck/outage domain:** one JSON file and one lock serialize every tenant; every mutation reads, clones, sorts, rewrites, and fsyncs all workspace state (`versionAnchor.ts:106-123,264-315,364-390`).
- **Initialization is not retry-safe:** `versionAnchor.ts:321-337` opens the final path before sealing/writing. Failure can leave an empty or partial file which the `wx` initializer then refuses to replace.
- **Incomplete erasure claim:** `kmsBackend.ts:160-195` copies the `readFile` buffer and creates immutable secret strings for encoded keys, but only later copies are zeroed.

### Exact merge gate

- Serialize first-DEK selection inside the same Postgres advisory transaction as the credential CAS. Wrapped-DEK generations must be immutable—never overwritten on conflict. A losing writer must reload the winning DEK and re-encrypt before committing.
- Replace DB-first mutation with a recoverable protocol:

  1. Insert immutable candidate DEK/envelope/version rows without changing the current pointer.
  2. CAS the external anchor from N to N+1 with an operation ID/digest.
  3. Switch the DB current pointer and tombstone old material.
  4. On startup, finalize anchor-ahead candidates deterministically; DB-ahead state must be impossible.

- Use a deployment-wide strongly consistent CAS/monotonic anchor for shared Postgres. Explicitly restrict the file implementation to single-host deployments.
- Replace persistent sentinel locks with crash-released OS locking or an external lease/CAS.
- Add an operator provisioning command, startup authentication/readiness check, and backup/restore topology documentation. Do not auto-create a missing anchor during normal startup.
- Add failure-injection tests for every persistence phase, two-host first-write concurrency, stale-lock recovery, and old DB+anchor replay.


*Cross-model review (codex Sol). Advisory, not a merge gate.*

hachej:
## Thermonuclear security review — REVISE

Reviewed exact head: 52c7b4ca62323299e1c51e67fa0be322008e1fcb

### Ranked findings

1. **P1 — DB-first / anchor-second failure is permanently unrecoverable.** `vaultStoreBackend.ts:252-333` commits the credential version inside `versionAnchor.withMutation`; only after the callback returns does `versionAnchor.ts:369-390` replace/fsync the external anchor. A crash, disk-full/EIO, or ambiguous fsync after the DB commit leaves DB=N+1 and anchor=N. `vaultStoreBackend.ts:133-145`, `:256-265`, and `:346-355` then reject every read and every future write before any repair can run. The PR body honestly documents fail-stop and lack of cross-store atomicity, but there is no authenticated two-phase intent, reconciliation, or operator recovery protocol. For durable production credentials, documenting permanent bricking is not enough; ship a safe recovery design or explicitly gate this adapter away from production.

2. **P1 — the local-file anchor has no enforced deployment topology and is incompatible with ordinary multi-replica shared Postgres.** The rollback lock is only `${anchorFilePath}.lock` on one filesystem (`versionAnchor.ts:364-394`). Two hosts with distinct local durable mounts but one Postgres DB maintain divergent anchors: after host A writes, host B rejects the new DB state. Worse, because wrapped generation 1 is upserted before the Postgres advisory lock/record CAS (`vaultStoreBackend.ts:275-290`; `postgresPersistence.ts:121-169,251-274`), independently anchored concurrent first writes can overwrite the wrapped DEK and commit fields encrypted under the losing DEK. Enforce/document a single-writer topology, or move coordination/monotonic state to a deployment-wide strongly consistent primitive and make DEK selection immutable/atomic with the record commit.

3. **P1 — KEK rollover and downgrade protection are incomplete.** The anchor MAC key is derived only from the currently configured KEK (`versionAnchor.ts:217-260`), so changing the KEK makes the existing anchor unreadable before it can be resealed. `rewrapDataKey` uses that same configured KEK to both unwrap and rewrap, so it cannot transition material from a genuinely old KEK to a new one. Also, the HMAC anchor prevents DB-only replay only while the KEK-adjacent anchor is outside the attacker's restore scope and an anchor-aware binary is enforced: replaying an old valid DB+anchor pair passes, and a downgraded pre-anchor binary ignores the check. Add a dual-key/reseal transition and state the required external anti-downgrade/deployment controls precisely; do not imply that this file alone is a monotonic rollback anchor.

4. **P1 — a crash leaves a permanent global write lock.** `acquireMutationLock` uses a sentinel file with no lease/PID ownership recovery, and the comment explicitly requires manual cleanup after a crash (`versionAnchor.ts:68-87,364-394`). Every workspace/provider mutation shares this one lock and rewrites/fsyncs the entire all-workspace JSON state. Use crash-released locking or external CAS/leases; this is both a restart availability defect and an unbounded global serialization point.

5. **P2 — KEK best-effort erasure is incomplete.** `kmsBackend.ts:259-283` copies loader material into a Buffer but never zeroes the original `Uint8Array`; invalid-length material is also retained. The rest of the implementation carefully wipes KEK/DEK buffers, so this extra live key copy defeats the stated best-effort handling. Zero loader material in a `finally` after copying/validation. Separately, `versionAnchor.ts:227-237` calls `material?.fill(0)` on an untrusted runtime return; a non-Uint8Array value can escape as raw TypeError instead of the stable credential error contract.

### Verified vs. trusted

Verified directly: PR body/full diff; anchor/KEK/vault/Postgres implementations and transaction ordering; migrations and conformance/rollback tests; DB-only snapshot replay does fail closed; anchor MAC domain separation, timing-safe comparison, AAD binding, metadata-only tombstones, sanitized error messages, and narrowed plaintext lifetime are sound. I found no secret-bearing log/error addition and no weakened preexisting security assertion. Current GitHub checks are 16 passed / 0 failed.

Took on trust: the claimed local test/typecheck executions, Postgres driver's guarantees beyond the visible transaction boundaries, and filesystem locking semantics on deployment filesystems. I did not rerun tests because these findings are protocol/order/topology defects not test-environment dependent.

hachej:
Owner merge approval recorded from Inbox question `9a1c4812-a46d-4166-bcec-e5d80a9bb53f` on 2026-09-09T18:20:01.870Z.

Approved exact revision: `62a096f8ad62a935219f0f044e320beec0e50b23` only.

Revalidated before recording:
- PR head unchanged and `MERGEABLE/CLEAN` against base `8690fd82ff278ab447585c52c14398316971f8d9`
- CI PASS: https://github.com/hachej/boring-ui/actions/runs/34371243126
- Workflow Invariants PASS: https://github.com/hachej/boring-ui/actions/runs/34371243185
- independent final review: SECURITY PASS and CROSS-PACKAGE ABSTRACTION PASS, no material findings
- tracker: 647/647 unique issue IDs; 1241/1241 unique nested comment IDs; 0 parent mismatches

The agent has not merged. Any code-head movement voids this approval and requires a new owner gate.

PR #1164, opened 2026-08-08
[new-feature][byok] slice B: pi-derived startup credential registry + vault resolver composition (1082)

hachej:
## Summary

BYOK slice B (plan r3 S3 server half / onboarding-plan r2 PR-B), **stacked on #1145** (`feat/1082-s1-persistence`): composes the shipped 16f.1 contract (`ProviderRegistryV1`, `createHostSideCredentialResolverV1`, `withResolvedCredential`) with the #1132 vault (`createVaultCredentialStoreBackendV1` + local-KEK KmsBackend + version anchor) at server startup, so a workspace host resolves LLM credentials from the vault.

### Composition design

- **`packages/agent/src/server/credentials/startupComposition.ts`** (new)
  - `derivePiLlmProviderCatalogV1()` — LLM provider catalog derived from pi's own surface: `ModelRegistry.inMemory()` provider set ∪ `AuthStorage.getOAuthProviders()` (auth kinds + display names), egress origins from pi model base URLs. No hand-maintained list; a provider pi adds appears without edits here. Unrepresentable ids are reported in `skippedProviderIds` (fail-visible, non-secret).
  - Per provider: `llm-model-call.v1.<provider>` trusted **host-only** consumer binding over a single secret `api-key` field; registries constructed through the validating 16f.1 factories.
  - `resolveWorkspaceCredentialVaultCompositionFromEnvV1()` — env-driven startup root:
    - `BORING_CREDENTIAL_KMS_BACKEND` unset → composition absent, behavior byte-identical to today.
    - Set to anything but `local-kek` → stable `CREDENTIAL_NOT_CONFIGURED` (a typo never silently disables the vault).
    - Missing KEK/anchor file env → `CREDENTIAL_NOT_CONFIGURED`; unreadable KEK at run time → `CREDENTIAL_BACKEND_UNAVAILABLE` (vault invariants carried through unchanged).
    - Durable persistence (#1145 port) is injected; in-memory needs an explicit `BORING_CREDENTIAL_PERSISTENCE=memory` test/dev opt-in, else fail-closed.
  - Returns `{ providerRegistry, bindingRegistry, catalog, vaultBackend, resolver?, createResolver }`; the resolver pre-binds when the Core-owned authority verifier is supplied, `createResolver` lets the host bind it later.
- **`buildAgentComposition.ts`** — builds the composition before harness assembly (misconfiguration fails composition with a stable `CREDENTIAL_*` code) and exposes it on `BuiltAgentComposition.credentials` for later slices (routes/PR-C, pi AuthStorage backend/PR-D).
- **`CreateAgentHostOptions.credentials`** — `{ env?, vaultPersistence?, authorityVerifier? }` embedding point.

Strictly composition/wiring + tests: no UI, routes, OAuth flows, or pi AuthStorage integration (later slices).

### Tests (10 new, `startupComposition.test.ts`)

- Registry mirrors pi's provider surface both ways (every pi provider represented or explicitly skipped; nothing pi doesn't know).
- Every catalog entry resolvable through the validating 16f.1 registries (host-only, `api-key` field only).
- Env-selection fail-closed matrix (absent / unknown backend / missing files / missing persistence).
- End-to-end: vault write → `withResolvedCredential` lease → plaintext only inside the lease; cross-workspace authority cannot read.
- Missing KEK file → `CREDENTIAL_BACKEND_UNAVAILABLE` on write and resolve.
- No secret material in catalog, registry, or any error surface.

### Verification

- `pnpm -C packages/agent typecheck` clean; `lint:invariants` OK.
- Agent suite: 1784 passed; the 2 remaining failures (`createAgentHost` credit-replay timeout, `PiChatPanel` hydrate timeout) reproduce identically on the clean base branch with this diff stashed — pre-existing on `feat/1082-s1-persistence`, not introduced here.


hachej:
## Gate-2 audit hardening applied (55c609756)

| Finding | Fix |
|---|---|
| **F1 (Medium)** per-binding composition + prod-reachable in-memory vault | (a) `BORING_CREDENTIAL_PERSISTENCE=memory` now refused under `NODE_ENV=production` unless `BORING_CREDENTIAL_ALLOW_MEMORY=1` (`CREDENTIAL_NOT_CONFIGURED` otherwise). (b) Composition hoisted to true once-per-host: resolved in `createAgentHost` startup, threaded into `createRuntime` → every binding's `buildAgentComposition`; bindings share one vault instead of minting fresh empty ones per rebind. |
| **F2 (Low)** startup-vs-binding failure timing docs | Env misconfiguration now genuinely throws at host creation (new `createAgentHost` test asserts `CREDENTIAL_NOT_CONFIGURED` at startup); `buildAgentComposition`/`types.ts` doc comments rewritten to state the actual behavior. |
| **F3 (Low)** raw backend + resolver minting on per-binding surface | Bindings now get a narrowed frozen `WorkspaceCredentialRuntimeViewV1` (registries + pre-bound resolver only); `vaultBackend`/`createResolver` remain host-scope-only. `createFakeAuthorityVerifierV1` moved behind `server/credentials/testing.ts` and dropped from both public index surfaces. |
| **F4 (Info)** dishonest cross-workspace test | Split into two honestly named assertions: authority denial (ungranted scope → `CREDENTIAL_AUTHORITY_INVALID`, before storage) and storage keying (validly-authorized ws-b principal → ws-b's empty rows, `CREDENTIAL_NOT_CONFIGURED`). |

Verification: agent typecheck clean, invariants OK, credentials suites 82/82, full server suite 1014 passed / 0 failed (including the previously-flaky `createAgentHost` suite, 13/13 locally).

hachej:
## 🌙 Overnight review — MERGE-AFTER-FIXES (blocked on #1145)

The code itself is the best-shaped part of this stack: correct layering, real capability narrowing, honest tests, no hand-maintained lists. But it cannot merge before its base, and its base is NEEDS-REWORK.

**Stack note:** this PR's base branch *is* #1145's head branch, and the two diffs are genuinely disjoint (composition/wiring vs persistence/anchor). GitHub reports this PR CLEAN only because it's measured against that stacked base, which is itself DIRTY — the classic stacked-PR merge trap. **True state vs main: conflicting** (`packages/agent/package.json`, `packages/agent/src/server/agent-host/buildAgentComposition.ts`). Main has not superseded any of this work.

### Correctness

**D1 (MAJOR) — nothing consumes it; this is wiring with no consumer.** `buildAgentComposition.ts:287` attaches `credentials: input.credentialComposition?.runtimeView` to `BuiltAgentComposition`, and across the entire branch **no code reads it**. The catalog, the registries, the resolver — all constructed, all unreachable. `derivePiLlmProviderCatalogV1` is called only from `resolveWorkspaceCredentialVaultCompositionFromEnvV1` and tests. That's legitimate for a deliberately-sliced security stack, and slicing is the right instinct here — but the named consumers (PR-C routes, PR-D pi AuthStorage) do not exist as open PRs, the epic has been idle 13 days, and neither this PR nor #1145 gives an operator any way to provision the vault (D2). Merging 881 lines of unreachable composition ahead of a consumer is only worth it if the consumer is imminent.

**D2 (MAJOR, spans both PRs) — a fully-configured deployment still fails every credential operation, because nothing provisions the anchor.** `resolveWorkspaceCredentialVaultCompositionFromEnvV1` (`startupComposition.ts`, the `createLocalCredentialVersionAnchorFromEnvV1` call) constructs a *reader* over `BORING_CREDENTIAL_LOCAL_KEK_ANCHOR_FILE`. It never creates the file and never checks it exists. `initializeLocalFileCredentialVersionAnchorV1` — the only thing that can create it, and it refuses to overwrite (`open(path,'wx')`) — is called from **tests only** across the whole branch. So: set every env var correctly, start the host, it boots clean; then every read and every write dies with `UNREADABLE 'Credential version anchor is unreadable'`. No CLI command, no startup provisioning, no runbook, no docs. Combined with #1145's C3, an operator's first two encounters with this feature are both "mysteriously fails closed forever." Either provision-on-first-use behind an explicit `BORING_CREDENTIAL_INIT_ANCHOR=1`, or add a CLI/migration step — and document it.

**D3 (MINOR) — the "fails at host startup" claim is narrower than advertised.** The test in `createAgentHost.test.ts` and the doc comments assert that env misconfiguration throws at host creation. True only for *env-shape* errors: unknown backend value, missing `BORING_CREDENTIAL_LOCAL_KEK_FILE`/`_ANCHOR_FILE`, missing persistence. The composition root is synchronous and never touches disk, so a KEK path that doesn't exist, an anchor that doesn't exist, or an unreadable/short KEK all boot fine and fail at first use. Reasonable design (lazy KEK load, and the PR does test the `BACKEND_UNAVAILABLE` outcome), but the doc comments at `types.ts:363-370` and `buildAgentComposition.ts:131-136` read as a stronger guarantee than the code gives. Add an optional async readiness probe, or narrow the wording.

**D4 (MINOR) — `createAgentHost` now reads `process.env` by default on every host.** `createAgentHost.ts:724`: `env: options.credentials?.env ?? process.env`. Safe today — an unset `BORING_CREDENTIAL_KMS_BACKEND` returns `undefined` early and behavior is byte-identical. Worth noting only because a stray env var in an unrelated deployment now hard-fails host creation. That's the intended fail-closed semantics and it's the right call; it just needs to be in the release notes.

**D5 (MINOR) — every derived provider is asserted to support `api-key`.** `derivePiLlmProviderCatalogV1` unconditionally emits `authKinds: ['api-key']` (plus `'oauth'` when pi lists an OAuth provider), and `toProviderDefinition` hardcodes `credential.type: 'api-key'` with a single required `api-key` field for *every* provider, including OAuth-only ones. Harmless for the registry's current use, but `authKinds` is documented as "metadata for later slices" — if PR-D drives UI from it, OAuth-only providers get an API-key form. Derive `api-key` support from pi rather than assuming it.

**D6 (MINOR) — providers with no parseable `baseUrl` get empty `egressOrigins` silently.** `toHttpsOrigin` correctly rejects non-HTTPS, credential-bearing, and `{template}`-host URLs — careful work. But a provider whose models yield zero usable origins still enters the catalog with `sandboxEgressOrigins: []`, and that is not surfaced in `skippedProviderIds` (which covers only id-grammar failures). If a later slice builds an allowlist from `sandboxEgressOrigins`, those providers get silently blocked with no diagnostic. Report empty-origin providers alongside skipped ids.

**D7 (MINOR) — pi-version-coupled test assertions.** `startupComposition.test.ts` asserts `catalog.length > 10`, `openai.authKinds` **exactly** `['api-key']`, and `anthropic.egressOrigins` contains `https://api.anthropic.com`. All three break on a pi upgrade that adds an OpenAI OAuth provider, changes a base URL, or drops providers below 11 — a dependency bump becomes a red build in a security package. The union-both-ways mirror assertion is the valuable one and is version-independent; tighten the rest to structural properties.

### Things done well

- Moving `createFakeAuthorityVerifierV1` off the public `server/index.ts` and `credentials/index.ts` barrels onto a dedicated `credentials/testing.ts` subpath is exactly right — a fake authority verifier must not be reachable from production composition. Good gate-2 catch.
- The `runtimeView` narrowing (`vaultBackend` and `createResolver` stay host-scope; bindings get registries + a pre-bound resolver) is real defense in depth, and the test asserts the *absence* of those keys and `Object.isFrozen`, not just their presence.
- Resolving the composition **once** per host rather than per binding (F1b in the second commit) fixed a real bug where each rebind minted a fresh, empty vault.
- The production in-memory-persistence refusal (`NODE_ENV=production` + `BORING_CREDENTIAL_ALLOW_MEMORY=1` override) is the right shape: fail closed, explicit escape hatch, injected durable persistence bypasses the check entirely.
- The cross-workspace test is split *honestly* into authority denial (`AUTHORITY_INVALID`) and storage keying (`NOT_CONFIGURED` against ws-b's empty rows) rather than one test claiming to prove both.
- Deriving the catalog from pi instead of hand-maintaining it is the correct call, and the union-both-ways test actually enforces it.

### BS check — clean

Commit messages describe what changed accurately. The body's claim that the 2 remaining suite failures reproduce on the clean base branch with the diff stashed is the right way to report a pre-existing-failure situation, and it matches the known `createAgentHost` credit-replay flake (fixed on main since, in `24b383b8c`). CI is fully green on both PRs in this stack, including a real Postgres service. No inflated claims except D3's over-broad "fails at host startup" wording.

### Fixes required in this PR

1. **D2** — ship anchor provisioning (explicit init flag, CLI step, or migration hook) and document it. This is the difference between a configurable feature and one that cannot be turned on. May land in #1145 instead; it must land in the stack before the consumer slice.
2. **D3** — narrow the startup-guarantee wording in `types.ts` and `buildAgentComposition.ts`, or add an async readiness probe.
3. **D6** — surface empty-`egressOrigins` providers as a diagnostic.
4. **D7** — de-couple the three brittle pi-version assertions.
5. Rebase after #1145 lands; resolve `packages/agent/package.json` and `buildAgentComposition.ts`.

D4/D5 are follow-ups.

**Owner decision needed (D1):** before merging, confirm PR-C/PR-D are actually next in the queue. If the epic is going back to sleep, hold this rather than land unreachable composition. Recommended sequencing: rework #1145's anchor (C1/C2/C3/C4 — the only genuinely hard work in the batch, and a design change rather than a patch) → land anchor provisioning + migration registration (D2, #1145's C8) somewhere in the stack → rebase and merge #1145 → apply D3/D6/D7, rebase, merge this, **only if** the consumer slice is actually queued.

hachej:
## 🌙 Overnight review — codex Sol cross-check (adversarial second pass)

A second, independent cross-model pass was run on this PR after the Opus review above. It **confirms the "nothing consumes it" blocker and escalates it**: the registry is not merely unwired pending later slices — the runtime path actively bypasses it.

**1. Vault credentials never reach model calls.** `buildAgentComposition.ts:287` returns `credentials`, but production model invocation constructs its own Pi auth independently — `AuthStorage.create()` / `ModelRegistry.create()` at `harness/pi-coding-agent/createHarness.ts:546-557`. Environment/settings auth stays authoritative, so enabling this feature changes nothing about where keys come from.

**2. No production composition root can enable durable credentials.** Core, Workspace, CLI and standalone `createAgentHost` callers inject neither `credentials.vaultPersistence` nor `authorityVerifier`. With `BORING_CREDENTIAL_KMS_BACKEND` set, `startupComposition.ts:317-336` refuses startup for missing persistence; unset, the feature is inert. Both branches are dead ends.

**3. The advertised startup fail-fast is false.** `startupComposition.ts:311-343` only builds closures — it never reads or authenticates the KEK or anchor. `startupComposition.test.ts:336-369` proves it: a missing KEK composes successfully and fails only on first use.

**4. The production durability guard is nominal.** `CredentialVaultPersistenceV1` carries no durability capability, so any injected implementation bypasses the guard — `startupComposition.test.ts:174-190` injects the in-memory backend under `NODE_ENV=production` and expects success.

**5. Enabled composition may silently have no resolver.** `startupComposition.ts:356-365` permits a missing authority verifier and returns `runtimeView.resolver === undefined`; `startupComposition.test.ts:255-264` explicitly blesses that state.

**6. Provider capabilities are fabricated.** `startupComposition.ts:181-217` treats every OAuth provider as API-key-capable and emits an API-key credential schema — misrepresenting OAuth-only providers (`github-copilot`, `openai-codex`). Related: the discovery registry (`startupComposition.ts:142-157`) is a different registry from the one that executes calls (`createHarness.ts:549-553`, `models/modelConfig.ts:231-243`), so custom/Infomaniak providers can be absent while unused built-ins are authorized.

**7. Dependency updates silently expand the trusted set** — new providers appearing in Pi automatically acquire trusted host-only bindings and egress origins with no security review. Inappropriate for an authorization registry.

**Clean areas** (both reviewers agree): AES-GCM/AAD construction, error redaction, transactional record/field/tombstone CAS, and hiding the plaintext write backend from runtime views.

**Merge gate:** resolve #1145's persistence/anchor blockers first; wire persistence + migrations + authority verification + anchor readiness through every supported production composition root; require an authority verifier whenever runtime credentials are enabled; make startup composition async and verify KMS readiness plus anchor MAC before host creation returns; add an enforceable durable-persistence discriminator; use one reviewed registry for selection, authorization and invocation, allowlisting API-key providers explicitly; resolve at the Pi provider-request boundary with invocation-scoped injection, disposed and zeroed in `finally`. Prove it with an integration test showing a real model invocation receives the workspace vault key while the process environment and Pi auth files never contain it.

*Cross-model review (codex Sol). Advisory, not a merge gate.*

hachej:
## Thermonuclear security/composition review — REVISE

Reviewed exact head: 828ed0cac828501cd27b1cfb2a461952efe07842 (stacked on #1145 head 52c7b4ca62323299e1c51e67fa0be322008e1fcb)

### Ranked findings

1. **P1 — production durability is not actually fail-closed.** `startupComposition.ts:317-330` checks the memory policy only when `options.persistence` is absent. Any `CredentialVaultPersistenceV1`, including the exported in-memory implementation, is accepted under `NODE_ENV=production` with no opt-in; the env-created memory store can also be re-enabled in production with `BORING_CREDENTIAL_ALLOW_MEMORY=1`. The tests explicitly bless both bypasses (`startupComposition.test.ts:174-190,255-264`). A production host can acknowledge credential writes and lose the entire encrypted vault on restart, conflicting with this slice's dependency on #1145 durable persistence and with the body calling memory a test/dev opt-in. Encode durability/kind in the port (or accept a separately typed durable port) and reject memory unconditionally in the production composition path; keep test/dev composition separate.

2. **P1 — the derived catalog fabricates API-key support for OAuth-only providers.** `startupComposition.ts:181-217` assigns every pi provider `authKinds` containing `api-key`, a required API-key schema, and an API-key consumer binding; OAuth merely adds a second metadata kind. In the pinned pi surface, `openai-codex` is an OAuth provider with generated models but has no API-key environment mapping. The registry therefore advertises and accepts a secret that cannot authenticate that provider. Derive actual API-key capability from pi's auth surface; until OAuth schemas are implemented, skip OAuth-only providers fail-visibly instead of claiming API-key support. The current tests spot-check Anthropic/OpenAI and miss this case.

3. **P1 — enabled BYOK can start without an authority verifier and has no usable late-binding path.** `startupComposition.ts:356-365` permits `resolver === undefined`, and the test at `:255-264` blesses it. In the actual public host path, `createAgentHost.ts:680-685` retains the full composition only privately, while `buildAgentComposition.ts:322` exposes `runtimeView`, which contains neither `createResolver` nor the raw backend. Thus the body's “createResolver lets the host bind it later” claim is not implemented here: the host starts with BYOK selected, but runtime bindings have no resolver. Require `authorityVerifier` when KMS is selected, or expose and test a real host-scoped late-binding lifecycle.

### Secret-surface result

I found **no secret leak** in the added logs/errors/catalog path: this module adds no logging, errors contain fixed env/backend names rather than values, the runtime view excludes the raw vault backend and resolver factory, and resolver failures remain sanitized. The blockers are durability, capability truthfulness, and fail-closed startup composition—not leakage.

### Verified vs. trusted

Verified directly: PR body/full stacked diff; startup env matrix and host wiring; narrowed runtime surface; pinned pi provider/auth capability behavior; #1145 persistence shapes at the composition boundary; relevant tests; no weakened assertion hiding a secret leak; current GitHub checks are 13 passed / 0 failed.

Took on trust: the PR body's full-suite/typecheck claims, #1145's cryptographic/Postgres guarantees beyond the inspected boundary, and pi runtime behavior outside the pinned catalog/auth metadata. I did not rerun tests because the three findings follow directly from the accepted branches and public composition surface, while CI is already green.

Issue #1123, opened 2026-08-07
Epic: executable environments — per-agent fs + exec grants (configurable multi-root execution)


hachej:
Owner-ratified direction (2026-08-07, from the #970 review conversation): agents should have access to various filesystems AND execution, configurable per agent — lifting the deliberate file-ops-only boundary on named filesystems where safe.

## Design anchors (already ratified elsewhere)
- Grants model from the healio design session: `filesystem.read/write`, `environment.bash.execute`, `environment.mount.manage`, `git.read/write`, `network.egress`, `secret.use`.
- Named filesystems stay honest: only **materialized** (real-dir) filesystems are mountable/executable; virtual/projected bindings remain file-ops-only (capability flag on the binding).
- Security default unchanged: readonly content mounts ro; exec is an explicit per-agent grant, never inferred.

## Today / Delta
| Piece | Today | Delta |
|---|---|---|
| Sandbox multi-mount | bwrap/gVisor mount the primary workspace | bind-mount N additional roots with ro/rw flags |
| Per-agent exec grants | fleet loader + capability projection (merged #1114) expose the policy seam; grants model designed, unbuilt | `environment.bash.execute` per fs in the agent's trusted policy (shares the #1087 seam) |
| Shell routing | #971 ships readonly shell qualification; single-root cwd | multi-root cwd selection + cross-root path qualification |
| Materialization contract | implicit | explicit `mountable` capability on RuntimeFilesystemBinding |

## Sequencing (blockers)
1. #970 + #971 on main (binding substrate + shell qualification)
2. #1107 (defines which filesystems an agent has: knowledge/, package roots)
Then this epic plans (Steward plan + adversarial review + gate 1) before any code.

## Why it matters
Killer demo for the productionized-agency track: an agent with its own knowledge fs AND a scratch exec workspace, grants-configured per client. Also the real answer to 'playground on a worktree from a pane' (#1083 adjacent).

Factory epic: 1 epic = 1 issue = 1 worktree = 1 PR chain; plan-first.

hachej:
**Design input from Cloudflare OS 'Gatekeeper' research (2026-08-07):**

Steal (mapped to our seams):
1. **network.egress must be default-deny at the SANDBOX layer** (bwrap/gVisor), not just the grants table — Cloudflare disables outbound networking independent of any data capability. Correction to this epic's design: a `environment.bash.execute` grant must NOT implicitly buy network.
[...]

PR #971, opened 2026-07-27
#942 feat: readonly primary workspace paths

hachej:
## Summary

Consolidates readonly primary-workspace paths for #942 on top of #970.

- Adds path-sensitive capabilities inside the primary `user` filesystem.
- Composes host and governance policy so readonly remains a ceiling.
- Enforces the same decisions across providers, HTTP routes, agent tools, UI affordances, uploads, moves, deletes, and shell qualification.
- Discovers conventional `.agents/skills/*/SKILL.md` resources from every request-visible filesystem binding, including virtual/remote bindings.
- Requires read access on the skill root, individual folder, and `SKILL.md`; denial removes the skill from the Skills pane/composer.
- Uses the canonical authorized file route to load skill content at invocation time instead of a second `/skills/invoke` protocol.
- Reconciles stale composer registrations and stabilizes request-header identity to avoid refresh loops.
- Keeps Pi/native duplicate-name winners authoritative; filesystem duplicate losers are management-only.
- Retires the legacy absolute readonly-skill bypass.

No separate trusted-skill permission exists: effective read access to the conventional skill path is the authority.

## Simplification pass

Cross-filesystem skill discovery moved out of #970 into this capability-owning PR and was reduced from the earlier custom invocation design:

- canonical `/api/v1/files` authorization is reused;
- custom invocation endpoint and second catalog resolution were deleted;
- request-scoped catalog entries never enter shared native-skill caches;
- equivalent request-header objects no longer trigger registration/refetch loops.

## Consolidates

Supersedes the review surface of #955, #957, #959, #960, #961, and #962.

## Proof

- `pnpm typecheck:changed` passes on the #970 base; focused #942 Agent typecheck passes.
- Focused skill catalog, composer, hook, and route-registration suites pass (75 tests).
- Root alignment, package, and plugin invariants pass.
- Read denial at `.agents/skills`, the skill directory, and `SKILL.md` is explicitly tested.
- CI is running for the latest revision.

## Review order

Review #970 first, then this PR. Do not merge either PR until review is complete.

Closes #938.
Closes #942.

PR #1166, opened 2026-08-08
[new-feature][env-mounts] slice 1: environment mount contract + provider substrate (#1123)

hachej:
## Summary

Slice 1 of the executable-environments epic ([plan r3.1](../blob/main/docs/issues/1123/plan.md), gh-1123): the mount **contract + provider substrate**, behind `BORING_ENV_MOUNTS`. No scope/lease plumbing, no grants derivation, no FUSE bridge — those are slices 2–4.

### What this slice covers (per the plan's own slice boundary)

- **Provider contract**: `SandboxEnvironmentMountV1` (`sourceRoot`/`logicalPath`/`access: ro|rw`), optional `mounts` on `SandboxProviderCreateContextV1`, optional `ProviderCapabilities.mounts` (bwrap `true`, all other providers `false`).
- **Fail closed**: direct, vercel-sandbox, and remote-worker reject a non-empty effective mount list at create with stable `SANDBOX_PROVIDER_MOUNTS_UNSUPPORTED` (new `ErrorCode`, documented in `ERROR_CODES.md`).
- **N-mount bwrap**: ro/rw binds after the primary workspace bind; logical paths must live in the dedicated `/mnt/<fsid>` namespace, never under `/workspace` (bind-shadowing/RO-under-HOME hygiene); duplicates/traversal/relative paths rejected.
- **Resolve-once hygiene**: `resolveEnvironmentMounts` realpath-resolves every `sourceRoot` exactly once at pair create; resolved paths are re-bound verbatim per exec (bwrap re-spawns per command — per-exec re-resolution would reopen the TOCTOU race). Missing/non-dir/workspace-aliasing sources (incl. symlink escapes) fail with stable `SANDBOX_MOUNT_INVALID`. The pre-existing out-of-workspace `.boring-agent` bind is brought under the same resolved-once rule (computed at sandbox init, realpath-resolved).
- **`buildBwrapArgs` dedupe** (plan-mandated): the byte-identical copy in `packages/boring-bash/src/agent/runtime/buildBwrapArgs.ts` is removed; the canonical implementation stays in `@hachej/boring-sandbox/providers/bwrap` and reaches boring-bash through the existing `RuntimeHostOperations` injection seam (no new package edge — a real dependency either direction would close a build cycle through `boring-agent`). `BwrapArgsOptions` stays in boring-bash as the injection contract type.
- **Latent `sandboxRoot` reconciled — honored**: new `sandboxHome` option on the builder; the `local-sandbox` strategy's `sandboxRoot` now selects bind target/chdir/HOME, not only the python-env paths. Default `/workspace` ⇒ zero behavior change.
- **Binding contract**: `RuntimeFilesystemBinding` gains optional `mountKind: 'direct' | 'view'` + server-private `materialization.sourceRoot` (boring-bash canonical shape and the agent `mode.ts` structural duplicate kept in sync). Inert metadata in this slice.
- **Catalog vocabulary**: `'execute'` added to `FILESYSTEM_CATALOG_CAPABILITIES` (primary `user` fs reports `true` — today's implicit exec; bindings report `false` until the exec-grants slice). Parser tolerates pre-1123 payloads (absent ⇒ `false`).
- **Docs**: `PROJECT_ENVIRONMENT_MODEL.md` amended (not repealed) with the gh-1123 mount rules.

### Flag / rollback

`BORING_ENV_MOUNTS` unset ⇒ `resolveContextMounts` yields the empty set for every context and bwrap args are byte-identical to today (existing snapshot unchanged — additions only; plus an explicit `mounts: []` ≡ no-options equality test).

### Tests

- `packages/boring-sandbox`: 446 passed (new: shared mounts/flag/fail-closed suite, `resolveEnvironmentMounts` incl. symlinked-sourceRoot resolution and symlink-escape rejection, bwrap-args mount/namespace/snapshot tests, direct-provider fail-closed create).
- `packages/boring-bash`: 54 passed (new catalog `execute` tests; route test expectations updated).
- `packages/agent` error-codes suite, workspace `filesystemPlugin` suite (414), and both packages' invariant lints pass.

### Sequencing note (important)

The plan gates all slices on #971 / its slim rebuild #1146, which touches these same files (`buildBwrapArgs` both copies, `mode.ts`, `types.ts`). Neither is merged yet; this PR is cut from `main` per instruction and **should land after #1146**, taking a rebase here if #1146 merges first. Do not merge before resolving that ordering.


hachej:
Independent review (pi worker, gemini-2.5-pro): 2 findings — **(MEDIUM)** nested/overlapping or duplicate `sourceRoot`s are not rejected in resolveEnvironmentMounts.ts (permission-confusion risk; fix round dispatched), **(LOW)** residual create→bind TOCTOU window acknowledged, to be documented explicitly for the SBX gates. Hold validation until the fix round pushes.

hachej:
Applied a review fix round (bfe72e81e..HEAD):

1. **MEDIUM — overlapping/duplicate mount sources**: `resolveEnvironmentMounts` now rejects, after realpath resolution, (a) the same resolved `sourceRoot` mounted at two logical paths (including duplicates that only appear post-symlink-resolution) and (b) any pair of mounts whose resolved source roots nest as ancestor/descendant, in either declaration order — all with the stable `SANDBOX_MOUNT_INVALID` code. Tests cover duplicate, symlink-alias duplicate, `/A` before `/A/B`, `/A/B` before `/A`, and the sibling `data` / `data-extra` prefix-string false-positive guard.
2. **LOW — residual TOCTOU documented**: the module doc now states the accepted risk window between create-time realpath and each bwrap bind, and why full elimination would require fd-based binds that bwrap's CLI contract does not offer.

`packages/boring-sandbox`: 450 tests + typecheck pass locally.

hachej:
### Mount resolution: check-then-bind is still a TOCTOU primitive (hardening, not a blocker)

From an adversarial review of the slice-1 mount path, verified against this PR's code **and** bubblewrap's source rather than inferred from the plan.

**The finding.** `resolveEnvironmentMounts` validates a pathname, stores only the resulting *string*, and a later exec asks a different process to bind that pathname. There is no stable host object reference between the two operations. Ordinary bwrap `--bind`/`--ro-bind` **re-resolves the source pathname** when constructing the mount namespace, so a host actor able to replace a checked path component can make a later exec bind a different host directory.

This does not contradict the plan's reasoning — resolve-once is still better than per-exec re-resolution — but it means the race was narrowed, not closed. The plan's wording ("per-exec re-resolution would reopen the TOCTOU race") reads as though resolve-once closes it.

**Exploitability — deliberately scoped.**
- **The agent:** no. It never obtains the host parent directory entry that would have to be renamed; the resolver rejects sources equal to, above, or below the primary workspace root, and named v1 mounts are read-only.
- **Another tenant:** no evidence on the inspected wiring; a tenant's namespace does not expose these host parent directories.
- **Plausible actor:** a co-located host process with the relevant Unix permissions, or a privileged host actor.

**Classification: hardening before feature enablement, not a present cross-tenant blocker.** The slice is flag-gated and dormant, so this does not need to hold up the merge.

**Minimal fix, for whichever slice enables mounts in anger.** Preserve an *object*, not a name: an fd-anchored component walk or `openat2(RESOLVE_NO_SYMLINKS|RESOLVE_BENEATH)` beneath a trusted anchor, returning an `O_PATH|O_DIRECTORY` handle plus stable ancestry metadata for the workspace-overlap test — then bind from the handle. Explicitly **not** `realpath()` followed by an ordinary `open()`, which just moves the race.

Full analysis with the attack sequence, bwrap source references, and a diff sketch is in the session research notes; happy to open a follow-up issue against the slice that turns mounts on.

hachej:
Owner decision: request-changes

I think we need ro revist this at the light of the vrey big exploration we have done in PR #1317... please stress test the approach wit thhis meta plan

hachej:
## 🌙 Overnight review — NEEDS-REWORK (extract the dedupe, re-plan the substrate)

**The owner already requested changes on 2026-08-19** ("I think we need to revisit this in the light of the very big exploration we have done in PR #1317… please stress test the approach with this meta plan"). That settles the verdict; the below is the technical case for what to salvage.

### Correctness

**B1 — the exported type guard now lies (must fix, and main already shows the fix).** `packages/boring-bash/src/shared/catalog.ts` on this branch:

```ts
export function isFilesystemCatalogCapabilities(value: unknown): value is FilesystemCatalogCapabilities {
  return FILESYSTEM_CATALOG_CAPABILITIES.every((capability) => {
    if (capability === "execute" && capabilities[capability] === undefined) return true;
    return typeof capabilities[capability] === "boolean";
  });
}
```

`FilesystemCatalogCapabilities` is `Record<FilesystemCatalogCapabilityKey, boolean>` with `execute` **required**, so the guard now narrows objects that lack `execute` to a type claiming it has one — and this is exported publicly from both `src/index.ts` and `src/shared/index.ts`. Main solved exactly this problem for a different capability, correctly, and this PR is unaware of it: `origin/main:.../catalog.ts:58-70` adds `FilesystemCatalogWireCapabilities = Omit<FilesystemCatalogCapabilities, "upload"> & { upload?: boolean }` with a separate *wire* guard for the tolerant parse path, then coerces `capabilities[capability] ?? false` at `:97` — generically, for any optional key. This PR's `execute` special-case is a strictly worse hard-coded second implementation of a pattern that now exists upstream. On rebase: add `execute` to the `Omit` union and delete the special-case branch. The `capabilities[capability] === true` coercion is likewise redundant against main's `?? false`.

**B2 — `.boring-agent` global-tool binds are now frozen at init: real behavior change, untested.** `packages/boring-sandbox/src/providers/bwrap/createBwrapSandbox.ts` hoists `globalToolMountArgs = await buildGlobalToolMounts(hostWorkspaceRoot)` into `init()` and reuses it in `exec`. But `buildGlobalToolMounts` contains a *negative* check — `if (await pathExists(childRuntimePath)) return false`, i.e. do not bind the parent's read-only copy once the workspace has its own. Caching at init freezes that decision: if the agent creates `.boring-agent/<runtime>` inside its workspace mid-session (a normal outcome of provisioning), every subsequent exec keeps binding the **parent's read-only** copy over it, shadowing the workspace's own writable one. The PR frames this purely as TOCTOU hygiene and adds no test for the mid-session case. **This is the one change in the PR that can affect a live bwrap session today, and it's the one change with no coverage.** Also: the `?? await buildGlobalToolMounts(workspaceRoot)` fallback is unreachable (`exec` throws when `!workspace`, and `workspace` is only set in `init`, which always assigns). Dead branch — delete it or make it an assertion.

**B3 — `resolveEnvironmentMounts` leaks a non-canonical error.** `realpath(mount.sourceRoot)` is wrapped in try → `SANDBOX_MOUNT_INVALID`, but the following `stat(sourceRoot)` is **not**. `realpath` succeeding and `stat` then failing (removal between the two calls, EACCES on a parent) throws a raw `ErrnoException` instead of the documented stable code. Tests only cover the file-not-a-directory case, where `stat` succeeds. Wrap it.

**B4 — `sandboxHome` vs the `/mnt` namespace rule is self-contradictory at the edges.** `buildBwrapArgs.ts` `validateMounts` enforces both `logicalPath.startsWith('/mnt/')` **and** `logicalPath` not under `sandboxHome`. Nothing forbids `sandboxHome` from being `/mnt` or `/mnt/x`, in which case every valid mount is simultaneously required and forbidden. Constrain `sandboxHome` to be outside `ENVIRONMENT_MOUNT_NAMESPACE`, or drop the redundant second check. (Inert today — `sandboxRoot` is `'/workspace'` at every call site, `packages/agent/src/server/runtime/modes/local.ts:19` being the only producer.)

**B5 — nothing populates `mounts`. The entire slice is unreachable.** Verified by grep across the branch: no production code path ever sets `SandboxProviderCreateContextV1.mounts`. Therefore `resolveContextMounts` returns `[]` unconditionally in production; `BORING_ENV_MOUNTS` gates a code path with no possible input; `assertNoEnvironmentMounts` in direct/vercel/remote-worker can never fire; `BwrapArgsOptions.mounts` is never passed by `bashToolOptions.ts`, its only caller; and `RuntimeFilesystemBinding.mountKind`/`materialization` are declared **twice** (`packages/agent/src/server/runtime/mode.ts:100-112` and `packages/boring-bash/src/agent/runtime/types.ts:390-403`) as manually synchronized structural duplicates that the body itself labels "Inert metadata in this slice."

### Over-engineering

This is the textbook shape of speculative plumbing: a contract, a capability flag, a feature flag, a fail-closed guard, a validator, a resolver, an error-code pair, a doc amendment, and two hand-synced duplicate interfaces — with **zero** production callers, behind a flag nobody can turn on usefully.

The mitigating facts are real and worth stating: this is explicitly slice 1 of a plan-gated epic, the slice boundary is honest, and the flag-off byte-identity claim is backed by a genuine equality test (`buildBwrapArgs.test.ts`: `mounts: []` ≡ no-options). This is *disciplined* speculative plumbing. But slices 2–4 never shipped — no other PR references #1123 — so thirteen days later the substrate still has no consumer, which is exactly the outcome the objection predicts.

**The one exception is genuinely valuable and stands on its own:** deleting `packages/boring-bash/src/agent/runtime/buildBwrapArgs.ts` (141 lines, byte-identical duplicate of the canonical `packages/boring-sandbox/src/providers/bwrap/buildBwrapArgs.ts`) removes a duplicated **security-critical argument builder**. The removed `boring-bash` exports (`buildBwrapArgs`, `BWRAP_TIMEOUT_SECONDS`, `KILL_GRACE_SECONDS`, `RO_BIND_DIRS`, `RO_BIND_TRY_DIRS`) have no importers on main — `packages/agent/host/sandbox.ts:1` and `test-host/sandbox.ts` already import from `@hachej/boring-sandbox/providers/bwrap`. Clean, zero-risk deletion, and it has a consumer today, unlike everything else here.

### BS check

Tests are real and reasonably adversarial — the symlink-escape rejection, the resolved-duplicate case, the ancestor/descendant nesting in **both** declaration orders, and the `data` / `data-extra` prefix-string false-positive guard are all substantive. No fake or weakened assertions. The residual-TOCTOU limitation is documented in the module docstring rather than hidden, and the PR responded correctly to two rounds of independent review.

Two claim-overreaches: the doc amendment says mounts live in "the dedicated `/mnt/<fsid>` namespace" implying one segment, but `validateMounts` only checks the `/mnt/` prefix so `/mnt/a/b/c` passes (cosmetic, but the doc is the contract); and the "should land after #1146" sequencing note is technically satisfied (#1146 merged 2026-08-09T19:06, last commit 19:09) but stale after 136 commits of drift.

### CI and staleness

15 passed, 0 failed on `3290ab37` — **136 commits behind main**, `CONFLICTING`/`DIRTY`. No signal. Conflicts: `packages/boring-bash/src/shared/catalog.ts` (main added the `upload` capability and the generic wire-tolerance pattern — see B1: the same problem solved twice, better upstream), `.../routes/filesystems.test.ts`, and `.../file-tree/__tests__/FileTreePane.test.tsx`, all from the same cause. Every capability object literal this PR touches (agent `createStandaloneAgentHostApp.test.ts`, `filesystems.ts` `PRIMARY_FILESYSTEM`, `capabilitiesFor`, `FilesystemRootsBinding.tsx`, three test files) needs `upload` added on rebase.

### Recommended shape

1. **Land now as a standalone PR from current main:** the `packages/boring-bash/src/agent/runtime/buildBwrapArgs.ts` deletion plus the `packages/boring-bash/src/agent/index.ts` export removal. Verified consumer-free. This is the whole real value of the PR and it's being held hostage by 900 lines of unreachable substrate.
2. **Hold the mount substrate** pending the #1317 stress-test the owner asked for. When it returns it must: rebase onto main's `upload`-era catalog and use `FilesystemCatalogWireCapabilities` instead of the `execute` special-case (B1); either drop the init-time `globalToolMountArgs` cache or add a test for the mid-session `.boring-agent` creation case (B2); wrap the `stat` call (B3) and constrain `sandboxHome` (B4); and **arrive with slice 2**, so `context.mounts` has a producer and the flag has a meaning (B5). A substrate with no consumer should not merge a second time.
3. Optionally split out the `execute` catalog vocabulary too — but only bundled with the exec-grants slice that makes it non-inert.

hachej:
Closed per owner triage 2026-08-26. The valuable extraction (bwrap dedupe) merged via #1359. Per the standing request-changes (2026-08-19) and the #1359 review, the mount substrate must return re-derived: rebased on the upload-era catalog, stat TOCTOU wrapped, globalToolMountArgs init-cache resolved, and bundled with slice 2 so context.mounts has a producer. Epic #1123 stays open as the tracker.

PR #1150, opened 2026-08-07
[gh-1107 slice 1] persona discovery via plugin asset manager

hachej:
## Summary
- discover plugin-shaped persona packages through `BoringPluginAssetManager` and inject plain scan descriptors into the agent fleet loader from workspace, core, and CLI boot layers
- keep activation roster-gated, surface unseated packages, and fail closed all duplicate `definitionId` claimants with `AGENT_DEFINITION_ID_CONFLICT`
- move skill identity to persona `pi.skills`, supporting bare shared-library names and contained package-relative paths while retaining fleet digest authority
- derive digest expectations from persona manifests and migrate the five factory persona manifests
- preserve the `BORING_AGENT_FLEET` flag boundary and legacy flag-off fleet

## Tests
- `cd packages/agent && pnpm exec vitest run src/server/agentDefinition/__tests__ src/shared/__tests__/error-codes.test.ts`
- `cd packages/workspace && pnpm exec vitest run src/shared/plugins/__tests__/manifest.test.ts src/server/__tests__/agentPlugins.test.ts`
- workspace boot tests for fleet discovery/conflict (`BORING_AGENT_FLEET` cases)
- `pnpm check:skill-digests`
- `pnpm --filter @hachej/boring-agent typecheck`
- `pnpm --filter @hachej/boring-workspace typecheck`
- `pnpm --filter ./packages/cli typecheck`
- `pnpm --filter @hachej/boring-core typecheck`
- `pnpm --filter @hachej/boring-agent run check:isolation`

Issue #1107, opened 2026-08-06
Epic: agent definition as plugin package (agent = installable, versioned, trusted)


hachej:
Design epic — plan approval (gate 1) required before any code.

Decision context: the healio design session converged on agent-as-package (AGENT.md + skills/ + knowledge/ + manifest) and agent-context filesystems (agent-owned readonly knowledge, versioned with the definition). Rather than a parallel system, an agent definition becomes a **plugin package with an agent contribution type**:

- Packaging/discovery/versioning/registry: existing plugin system (packages/workspace PLUGIN_SYSTEM.md).
- Skills: package-owned skill resources (#938/PR #970 substrate — install paths server-private, {filesystem,path} locators).
- Knowledge: readonly filesystem binding, provenance = agent-definition (server-authoritative, catalog-driven per #1000).
- Trust: same digest pinning as factory personas (content-addressed instructions).
- Inspection: #1100 Agent Details is the read-only UI face.

Absorbs/sequences: #1087 (MCP grants per agent), #1100 (deferred until multi-agent stabilizes), agent-context fs design. Depends on: fleet-loader epic slice 3, PR #970/#971 landing.

Deliverable of the first slice: a Steward plan + visual plan doc for owner gate 1 — schema of the agent contribution, migration of factory personas to the format, and what stays out of scope.

hachej:
**Owner ratifications (2026-08-06, gate-1 pre-work):**
1. Persona/agent discovery goes through the plugin system's BoringPluginAssetManager — one pipeline, no bespoke persona reader (the fleet loader's direct dir read is the temporary bridge).
2. **Skills declared via `package.json#pi.skills` in the agent package** (revises the fleet.yaml-holds-skills split): personas declare skills like every plugin; trust preserved because .agents/personas/** is permanently class-B and digest pinning moves with the declaration (check:skill-digests adapts). fleet.yaml shrinks to the seat roster or dissolves into policy.yaml.
3. `knowledge/` folder = readonly filesystem binding, provenance agent-definition, versioned with the package (healio agent-context design).

Priority raised: plan (gate 1) immediately; execution unblocks when #970/#971 + #1114 are on main.

hachej:
## Gate input: where does an agent definition physically live?

Surfaced while shipping #1176 (Agent details capability inventory). The panel renders a
row per authored file behind an agent — persona instructions, skills, later knowledge —
and each row opens in the workbench. Building it forced the question this epic has left
[...]

Issue #997, opened 2026-07-30
Expose server capabilities for editable skill sources


hachej:
## Problem

The Skills management page always opens skill sources with `mode: "view"`, even when the backing resource is writable for the current actor. This is safe for package and governed skills, but prevents useful authoring workflows for personal, development, generated, and tenant-admin skills.

Editability must not be inferred in the frontend from filesystem names such as `user`. It is a request-scoped server capability determined by the effective filesystem/path policy.

## Desired behavior

Examples:

| Resource | Effective server capability | Skills action |
|---|---:|---|
| Personal `user/.agents/skills/*` without a readonly policy | `write: true` | Edit source |
| Healio `user/.agents/skills/*` under `readonlyWorkspacePaths` | `write: false` | View source |
| Package `agent_resources` skill | `write: false` | View source |
| Readonly `company_context` skill | `write: false` | View source |
| Writable tenant-admin skill | `write: true` | Edit source |

Invocability and editability are independent. A readable duplicate-name loser may be management-only for invocation while still being editable if the server grants write capability.

## Proposed direction

1. Extend the request-scoped skill catalog resource row with effective server capabilities, using the canonical filesystem binding/path access resolver.
2. At minimum expose `read` and `write`; prefer the canonical capability set (`read`, `write`, `create-child`, `delete`, `move-from`) rather than a frontend-derived `editable` boolean.
3. Render **Edit source** only when `write: true`; otherwise render **View source**.
4. Keep every mutation route authoritative and freshly authorized. Catalog capabilities are presentation hints, never grants.
5. Reconcile capabilities on catalog refresh, policy changes, actor/workspace changes, and plugin reload.
6. Fail closed if capability resolution is unavailable.

## Authorization and caching

- Compute capabilities per request; do not put actor/path-specific decisions in the shared native-skill cache.
- Verify read access before returning the skill row.
- A stale edit UI must still receive a server denial after revocation.
- Package and remote filesystem operations must flow through `RuntimeFilesystemBinding`, not host `node:fs`.

## Acceptance criteria

- [ ] Writable user/admin skills expose Edit source.
- [ ] Readonly user, package, and company skills expose View source only.
- [ ] `readonlyWorkspacePaths` changes are reflected after refresh.
- [ ] Duplicate invocability does not determine editability.
- [ ] Revoking write access while an editor is open causes subsequent saves to fail closed.
- [ ] Catalog responses do not expose host paths.
[...]

Issue #938, opened 2026-07-24
Discover and open package-owned skills through readonly resource bindings


hachej:
## Summary

Discover and open package-owned skills through the existing multi-filesystem/resource policy model, without exposing absolute installation paths or routing package files through the primary workspace Files API.

`@hachej/boring-bi-dashboard` is the concrete failing case. Its package manifest declares `skills/bi-dashboard-authoring`, but direct registration through `createBiDashboardServerPlugin(...)` omits the skill. Manually wiring the installation path exposes an absolute `node_modules/.../SKILL.md` path that the guarded Workspace correctly rejects.

## Observed

- The BI package correctly declares `package.json#pi.skills`.
- Direct/no-provision registration contributes bridge behavior and prompt text but not the package skill.
- `/api/v1/agent/skills` can expose an absolute package installation path when manually configured.
- The Skills UI attempts to open that path as a primary workspace file and receives an outside-workspace rejection.
- Downstream copies under `.agents/skills` drift from the package and are not acceptable.

## Required model

Use the existing multi-filesystem binding architecture rather than a skill-specific absolute-path exception:

```ts
// Workspace-owned skill
{ filesystem: "user", path: ".agents/skills/healio-reporting/SKILL.md" }

// Package-owned skill
{
  filesystem: "agent_resources",
  path: "packages/@hachej/boring-bi-dashboard/skills/bi-dashboard-authoring/SKILL.md"
}
```

The server-side package-resource adapter owns real installation paths and confinement. The browser receives only a safe locator and readonly access. Direct/no-provision hosts must work without copying package files into the workspace.

## Acceptance

- [ ] Registering the BI dashboard server plugin contributes its package-declared skill without downstream path wiring.
- [ ] Package skills appear in `/api/v1/agent/skills` with a safe `{ filesystem, path }` or equivalent opaque locator; no absolute `node_modules` path is exposed.
- [ ] The Skills UI opens package and workspace skills through their declared filesystem identity.
- [ ] Package-resource bindings are readonly and reject traversal, ambiguous package identities, symlink escape, and all mutations.
- [ ] Direct/no-provision hosts can discover and read package skills without workspace materialization.
- [ ] Workspace Files APIs continue rejecting arbitrary absolute/outside-workspace paths.
- [ ] Package and workspace skills can coexist with the same display name without identity ambiguity.
- [ ] No downstream copy under `.agents/skills` is required.
[...]

PR #1168, opened 2026-08-08
[gh-1107 slice 2] knowledge/ as agent-scoped readonly fs

hachej:
Implements epic #1107 slice 2 (stacked on #1150 / `feat/1107-s1-discovery`).

## Design

- **`knowledge/` in the definition package** — `compilePersonaPackageDirectory` walks an optional `knowledge/` folder through the same contained-read fence as `instructions.md` (symlink refusal, realpath containment, per-file 256KB / 128-file bounds, UTF-8 fail-closed). Each file becomes a digest asset `knowledge/<rel>`, so **knowledge bytes fold into the computed definition digest**; absent folder = byte-identical digest and no binding.
- **Digest is the identity** — `AuthoredAgentSourceV1`/`ConfiguredAgentHostAgentSpec.definition` carry the computed `definitionDigest` (never authored; derived by the trusted compiler), and `host.describe()` surfaces it per agent. Identity is the digest, not the version string.
- **Agent-scoped readonly binding at composition** — `buildAgentComposition` (the one Agent-owned assembly funnel) mounts the spec's `knowledge.rootDir` via `runtimeHost.createAgentResourceFilesystemBinding` as filesystem id `agent_knowledge` with new `provenance: 'agent-definition'`. Because this lives in the funnel, every host (workspace server, core, CLI hub) gets it with zero per-host wiring; the binding exists only for the owning agent — sibling agents get "No filesystem binding is available for agent_knowledge". A declared-but-unmountable knowledge folder fails that agent's composition closed (fleet exclusion diagnostics unchanged).
- **Provenance in the catalog** — `RuntimeFilesystemBinding` (boring-bash + agent mode.ts) and `FilesystemCatalogEntry` gain optional `provenance: 'agent-definition'`, surfaced by `/api/v1/filesystems`.
- **Exemplar** — the steward persona ships real knowledge (`.agents/personas/steward/knowledge/plan-quality-checklist.md`).
- Generic ids only (`agent_knowledge`); no tenant names in core. No new flag — rides `BORING_AGENT_FLEET`; flag-off behavior untouched.

## Proof

- `compileAgentDirectory.test.ts`: digest changes when knowledge bytes change; stable when unchanged; absent folder = no `knowledgeDir`; symlink inside `knowledge/` fails closed (`PATH_SYMLINK_ESCAPE`); non-UTF-8 fails closed.
- `loadConfiguredAgentFleet.test.ts`: fixture package's knowledge root + `sha256:` digest ride the composed spec.
- `createAgentHost.test.ts` boot smoke: `agent_knowledge` readable by the owning agent, absent for a sibling, mutation rejected with bytes intact (company_context-style no-leak), digest on `describe()`, and not leaked into the Environment-level catalog.
- `environmentHttpProjection.test.ts`: catalog surfaces `provenance`.
- `pnpm -C packages/agent test` (2003 tests), `pnpm -C packages/workspace test`, `pnpm -C packages/boring-bash test`, `pnpm check:skill-digests`, typecheck agent/workspace/core — green locally (two load-flaky files pass in isolation).


hachej:
Independent review (pi worker, gemini-2.5-pro): **PASS, no findings.** Contained-read fence complete (lstat symlink refusal + realpath containment, bounds before read); digest sorted+content-based (reorder/rename-proof); sibling isolation + readonly verified non-vacuously; fail-closed on unmountable declared folder; shared types tenant-clean.

hachej:
## Question before merge: is `agent_knowledge` readonly a boundary or a label?

Raised from the #1107 gate discussion (comment there), where the proposed decision is that
definition packages are **workspace-visible** under `.agents/` — the fleet composes from
the same root the workspace serves. This PR's own exemplar already assumes that shape:
`.agents/personas/steward/knowledge/plan-quality-checklist.md`.

If that holds, the same bytes become reachable two ways:

- `agent_knowledge`, mounted readonly, `provenance: 'agent-definition'`, scoped to the
  owning agent;
- the `user` filesystem at `.agents/personas/steward/knowledge/…`, under whatever
  editability policy the workspace applies (#997 / #938 / #578).

Two mounts of one file with different write policies. That is not automatically wrong, but
it means "readonly" needs to be saying something precise. Which is it?

1. **A security boundary** — the agent must not rewrite its own knowledge at runtime. Then
   readonly on `agent_knowledge` is correct and holds regardless of what the `user`
   filesystem allows, because the two mounts have different *subjects*: the agent versus
   the human. Worth stating explicitly in the PR description, since the current wording
   reads as a property of the files rather than of the agent's access to them.

2. **A statement that this is definition content, not user content** — then, with `.agents/`
   workspace-visible, readonly is cosmetic: the human edits the file through the workbench
   anyway, and the mount's policy only shapes what the UI implies.

My read is that (1) is what you want and largely what the code does — the no-leak and
mutation-rejected tests point that way — but the description doesn't distinguish the two,
and the difference decides whether "knowledge is editable" is a supported workflow.

### On digest-pinning vs editability

These are not in conflict, and I don't think this PR needs to change to accommodate
editing. The digest is the definition's identity *as composed*; editing a knowledge file
drifts it; a reload recomposes and reconciles. That reload affordance is already tracked as
#1171. The only thing to avoid is treating digest stability as a reason the files must be
immutable on disk — it isn't, it's a reason a reload has to exist.

### Ask

Just a clarification of intent, plus a line in the description saying which of the two the
readonly mount is. If it's (1), nothing here needs to change. If it's (2), the mount policy
is worth revisiting before this lands, because the editability story for definition files
gets decided by whatever ships first.

hachej:
Superseded by #1202, which consolidates the full #1107 chain (discovery + knowledge + install) + the S0 three-seat roster into one branch (rebased clean, MERGEABLE). #1202 carries this PR's content, adapted to the 3-seat roster. Branch preserved — reopen if #1202 does not land. Closing to keep the review surface single (#1202).

🤖 Generated with Claude Code.

Issue #1187, opened 2026-08-10
Epic: migrate the Boring Factory onto the boring-ui CLI workspace (dogfood the product)


hachej:
Owner-defined lane (2026-08-10): the factory — today orchestrated from Claude Code sessions + beads + repo skills — migrates onto boring-ui's own CLI workspace.

Target shape (to be refined by the plan):
- The factory runs as a boring-ui workspace via the CLI hub: fleet seats for the factory roles (orchestrator/planner/worker/reviewer — .agents/factory/fleet.yaml already exists), sessions per lane.
- Beads/work items → the workspace Tasks surface; owner gates and validations → Inbox Human Intention items (the epic #1110 loop already ships inbox cards); recurring loops (CI watch, lane heartbeat) → Automations.
- Worker routing (pi+Sol / gemini reviews / Claude taste) expressed through the workspace's model/provider selection instead of ad-hoc shell dispatch; the present-pr review artifact (#1180/#1184) becomes a workspace pane/artifact.
- Claude Code sessions remain the escape hatch, not the control plane.

Why now: the substrate is merged or queued (fleet loader, multi-agent UI #1102, per-automation agent selection #1143, durable streams, agent-definitions-as-packages #1107 chain, ask-user inbox), and the strategy report (docs/direction/state/2026-08-08.md) flags factory DX as evidence the seams need first-party use.

Process: plan-gate first (docs/issues/<n>/plan.md r1 [owner gate]) with Today/Delta against how the factory ACTUALLY runs today (read skill-library/boring-v2/MODEL-CARD.md, .agents/factory/, .agents/skills/{plan,exec,triage}/, docs/procedures/boring-loop.md, the beads workflow); then PR-sized migration slices, each keeping the old path working until its replacement is proven in a real lane.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QtPPbToGDvkARQZuYKFyWN

hachej:
**gh-1187 S0/S1 field report — three gaps filed as issues:**

- #1195 — `MODEL_TIER_CANDIDATES` is Anthropic-only, so `policy.yaml#models.seats` tiers resolve to nothing on this Gemini-funded instance. Sharpens the plan's own "which keys the hub can see silently determines spend" risk.
[...]

PR #1204, opened 2026-08-10
feat(tasks): epics-first drill-down view (gh-1187 S5)

hachej:
Delivers S5 of the Boring Factory migration epic (gh-1187), which pulls gap **G6**: "see all the active epics and drill down into their beads".

## Today / Delta

**Today.** `BoringTaskCard.epic` already exists and the beads adapter already maps a bead's parent onto it (`beadsSource.ts` `nativeParent`). The board exposes that as a single **epic filter dropdown** plus two flat views (kanban, list). To answer "which epics are active", you pick epics from a `<select>` one at a time — the epic level has no surface of its own, and nothing tells you an epic exists until you open the dropdown.

**Delta.** A third view mode, `epics`: an epics-first list, each epic a collapsible row carrying its counts and status breakdown, expanding to its beads. Active-only by default. No adapter, schema, or route work — this is a view over data the board already loads.

## Design rationale

- **The epic level gets its own rows, not a filter.** Epic identity is the adapter-scoped composite key `${adapterId}:${epic.id}` — the same key the existing filter uses, now exported once from `taskEpicsModel.ts` (`epicGroupKey`) and consumed by both, so a filter selection and a group can never disagree.
- **"Active" is legible, and hiding is never silent.** An epic is active while at least one of its beads sits in a non-terminal column. Terminal is id-based (`closed`, `done`, `completed`, `cancelled`, `archived`) so beads (`closed`) and GitHub Issues (`done`) both get the split with no adapter coupling. The default is active-only, but the Active/All segmented control is always visible and the line beside it always states the exact count withheld — `"3 active · 2 fully closed epics (7 tasks) hidden"`. The empty state says the same thing and names the way back.
- **Every row states its shape before you open it.** Per-column pills in board column order (terminal columns rendered muted) plus an `open/total` badge, so scanning epics is a read, not a click-through. Beads whose status is not in the board config still count, under their raw status id — nothing vanishes from a total.
- **One filter pipeline.** The epics view groups `filteredTasks`, so source multiselect, tag, epic, and column visibility compose unchanged. Turning a column off removes its beads from every epic's counts, which can legitimately move an epic out of the active list — the withheld counter reports that.
- **One card, one dialog.** The list view's `TaskCard` wiring was extracted into a single `renderTaskCard`, shared by both views; clicking a bead in an epic opens the same `TaskDetailDialog`.
- **The board did not grow into a monster.** Pure grouping/aggregation lives in `taskEpicsModel.ts` (mirroring `taskBoardModel.ts`), presentation in `TaskEpicsView.tsx`. `TaskKanbanBoard.tsx` gained ~40 net lines, having also shed the duplicated card block.
- **Accessibility.** Each epic header is a `<button>` inside an `<h3>` with `aria-expanded` and `aria-controls` pointing at its panel; the panel is `hidden` when collapsed. The activity toggle is a labelled `role="group"` with `aria-pressed`; the withheld-count line is `aria-live="polite"`. The view switcher keeps its `aria-label="Task view mode"` pattern and gains a labelled "Show epics view" control. Fully keyboard operable.

## Read-only guarantee

Per plan decision 3, the Tasks panel is a read-only mirror for the whole migration and beads stays canonical. **No write path was added or touched.** The diff is four front files; no adapter, no server source, no route, no tool, no shared type. Grouping is derived at render time from `listTasks()` output. `plugins/tasks/scripts/prove-beads-readonly.mjs` remains the standing proof and is unaffected.

## Test evidence

```
pnpm -C plugins/tasks test
  Test Files  20 passed (20)
       Tests  178 passed (178)     # 162 before, +16 new

pnpm -C plugins/tasks typecheck    # clean
pnpm lint:invariants               # plugin-invariants ok (488 files); all alignment checks PASS
```

New coverage — note nothing previously covered the epic filter or the view switcher at all:

- `taskEpicsModel.test.ts` (8): adapter-scoped key collision, terminal-column recognition, grouping + title ordering with "No epic" pinned last, open/closed counts and activity, status breakdown ordering and colors, statuses absent from the board config, raw-id title fallback, and the active/closed split reporting exactly what it withholds.
- `TaskEpicsView.test.tsx` (8): switching to the epics view and reading counts, the active/closed toggle in both directions, expand/collapse with `aria-expanded`/`aria-controls` verified against a real panel node, keyboard operation from switcher to bead, epic-filter and tag-filter composition, column visibility emptying an epic into the closed set, view-mode persistence across a remount, and opening the shared `TaskDetailDialog` from a bead inside an epic.

## Screenshots / evidence

_Placeholder — owner-facing navigation surface; screenshots to be attached from the S5 proof run (owner navigates active epics → beads and finds S1's lane and its lease state without opening a terminal)._

## UI-review spec

Not added in this PR — see the report/thread for the reasoning and the follow-up recommendation.

### DIFF
--- a/plugins/tasks/src/front/TaskEpicsView.test.tsx
+++ b/plugins/tasks/src/front/TaskEpicsView.test.tsx
@@ -1,0 +1,182 @@
+import { render, screen } from "@testing-library/react"
+import userEvent from "@testing-library/user-event"
+import { beforeEach, describe, expect, test } from "vitest"
+
+import { WorkspacePluginClientProvider } from "@hachej/boring-workspace"
+import type { BoringTaskAdapter, BoringTaskCard } from "../shared"
+import { TaskKanbanBoard } from "./TaskKanbanBoard"
+
+const COLUMNS = [
+  { id: "open", title: "Open" },
+  { id: "in-progress", title: "In progress" },
+  { id: "closed", title: "Closed" },
+]
+
+function card(id: string, statusId: string, title: string, epic?: { id: string; title: string }, tags?: string[]): BoringTaskCard {
+  return { id, number: id, title, statusId, adapterId: "beads", ...(epic ? { epic } : {}), ...(tags ? { tags } : {}) }
+}
+
+const TASKS: BoringTaskCard[] = [
+  card("bd-1", "open", "Lane worktree pool", { id: "bd-100", title: "Factory migration" }, ["lane"]),
+  card("bd-2", "closed", "Lane registry", { id: "bd-100", title: "Factory migration" }),
+  card("bd-3", "closed", "Retired spike", { id: "bd-200", title: "Old spike" }),
+  card("bd-4", "in-progress", "Unparented chore"),
+]
+
+function beadsAdapter(tasks: BoringTaskCard[] = TASKS): BoringTaskAdapter {
+  return {
+    id: "beads",
+    label: "Beads",
+    capabilities: { move: false },
+    getBoardConfig: async () => ({ adapterId: "beads", columns: COLUMNS }),
+    listTasks: async () => tasks,
+  }
+}
+
+function renderBoard(adapter: BoringTaskAdapter = beadsAdapter(), workspaceId = "epics-test") {
+  return render(
+    <WorkspacePluginClientProvider agentTypeId="default" apiBaseUrl="" workspaceId={workspaceId}>
+      <TaskKanbanBoard adapters={[adapter]} />
+    </WorkspacePluginClientProvider>,
+  )
+}
+
+async function openEpicsView(user: ReturnType<typeof userEvent.setup>) {
+  await screen.findByText("Lane worktree pool")
+  await user.click(screen.getByRole("button", { name: "Show epics view" }))
+}
+
+describe("TaskKanbanBoard epics view", () => {
+  beforeEach(() => localStorage.clear())
+
+  test("switches to an epics-first list with per-epic counts and a status breakdown", async () => {
+    const user = userEvent.setup()
+    renderBoard()
+    await openEpicsView(user)
+
+    const epic = screen.getByRole("button", { name: /Factory migration/ })
+    expect(epic).toHaveAttribute("aria-expanded", "false")
+    expect(epic).toHaveTextContent("bd-100")
+    expect(epic).toHaveTextContent("Open1")
+    expect(epic).toHaveTextContent("Closed1")
+    expect(epic).toHaveTextContent("1/2")
+    expect(screen.getByRole("button", { name: /No epic/ })).toBeInTheDocument()
+    // Collapsed by default: beads are not rendered until the epic is expanded.
+    expect(screen.queryByText("Lane worktree pool")).not.toBeInTheDocument()
+  })
+
+  test("hides fully closed epics by default and restores them through the visible toggle", async () => {
+    const user = userEvent.setup()
+    renderBoard()
+    await openEpicsView(user)
+
+    expect(screen.queryByRole("button", { name: /Old spike/ })).not.toBeInTheDocument()
+    expect(screen.getByText(/1 fully closed epic \(1 task\) hidden/)).toBeInTheDocument()
+
+    await user.click(screen.getByRole("button", { name: "All epics" }))
+    expect(screen.getByRole("button", { name: /Old spike/ })).toBeInTheDocument()
+    expect(screen.getByRole("button", { name: "All epics" })).toHaveAttribute("aria-pressed", "true")
+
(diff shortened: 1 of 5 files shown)


hachej:
## Hub evidence — the view works; the data does not exist. **Do not merge as "epics drill-down works".**

Driven against the **real factory beads graph** (`.worktrees/factory`, `br` v0.2.16, 288 real tasks: 226 beads + 62 GitHub issues, zero source errors). The owner's hub on :5210 was never touched; every process started for this run was killed afterwards and the factory worktree is git-clean.

### What works
Three-mode switcher (`Show kanban/list/epics view`), persistence, per-column status pills, the `open/total` badge, the Active/All toggle with its withheld-count line, filter composition (tag `391` → group recomputed to `87/130` with correct pills), `TaskDetailDialog` opening on a real bead from the epics view, and **keyboard-only a11y verified**: Tab reaches the epic header (`aria-expanded=false`), Enter expands it with a visible focus ring. Pill totals were recomputed independently from the raw API payload and match exactly (43+1+10+9+46+7+85+87 = 288).

### The headline problem
**0 of 288 tasks carry an `epic` field** — 0/226 beads and 0/62 GitHub. The epics view therefore renders **one group, "No epic", holding all 288 cards in a flat scroll**. On the only real dataset available it is strictly worse than the existing kanban and list views. The Epic filter dropdown has exactly one option ("All epics") — a dead control on all three modes, and pre-existing.

### Root cause — two upstream blockers, both pre-existing on `main`

**B1 — the adapter's `br list` args cannot return parent edges.** The exact `LIST_ARGS` in `plugins/tasks/src/server/beadsOperations.ts:41` (`br list --all --json --no-auto-flush --no-auto-import --no-db`) returns only `dependency_count` / `dependent_count`, with **no `parent` and no `dependencies` key**. Verified directly against the factory store — the returned keys are:

```
acceptance_criteria, assignee, close_reason, closed_at, compaction_level, created_at,
created_by, dependency_count, dependent_count, description, external_ref, id, issue_type,
labels, notes, original_size, priority, source_repo, source_repo_path, status, title, updated_at
```

`nativeParent()` (`beadsSource.ts:134-141`) consequently always returns undefined and line 178 never sets `epic`. The data *does* exist — `.beads/issues.jsonl` has **161 `parent-child` edges** and 11 `issue_type: epic` beads, and `br show --json` *does* return `dependencies[]` with `dependency_type: "parent-child"`. Only the list call is blind. Secondary: `beadsSource.ts` reads `dep.nativeType` while `br` emits `dependency_type` — that name needs checking too, even once edges arrive.

**B2 — the CLI hub cannot load beads tasks at all.** `packages/cli/src/server/workspacePluginRoutes.ts:93` calls `createTaskSourceRegistryFromConfig(workspace.plugins?.tasks, { workspaceRoot })` and **never passes `beadsOperations`**, so the source returns `TASK_BEADS_RUNTIME_UNAVAILABLE`. The sole non-test caller of `createWorkspaceBeadsOperations` in the repo is `apps/workspace-playground/src/server/dev.ts:65`. This evidence run had to fall back to the playground host for that reason. (Separately: the factory workspace registry entry declares no `plugins.tasks.providers`, so Tasks is empty on the owner's hub today regardless.)

### Consequence for the plan
`docs/issues/1187/plan.md` G6 states the Beads adapter "*does* map a bead's parent to `epic` (`beadsSource.ts:178`)" and concludes the slice is "cheap because no adapter or schema work is needed". **The mapping code exists; the data never reaches it.** G6's premise is wrong, and S5's stated proof — *"the owner navigates active epics → beads"* — is unreachable until B1 is fixed. That is an adapter change behind a security-reviewed argv firewall (likely `br list --type epic` plus a bounded `br dep tree` per epic, ~12 calls), which is its own slice, not a tack-on here.

**Recommendation:** hold this PR. The view is correct, tested (178 tests) and CI-green, but it should land with or after the adapter fix, otherwise it ships a navigation surface with nothing to navigate. Two beads wanted: *beads adapter populates `epic` (B1)* and *CLI hub wires `beadsOperations` (B2)*.

### Rough edges (for the UI-loop follow-up)
- Expanding a group mounts **288 card rows at once** — no virtualization, noticeably heavy scroll.
- Bead IDs render uppercased and untruncated (`WT-391-FORWARD-GH912-LIVE-TRANSCRIPT-8R4G.1`), eating much of the row width.
- Clicking a card **title** does nothing; detail opens only via a hover-revealed icon. Consistent with kanban, but a discoverability miss in a drill-down view whose rows read as list items.
- When "No epic" is the only group, its dashed-border treatment reads as an empty-state placeholder rather than "all your work".
- Latent once B1 lands: `taskEpicsModel.ts`'s `task.epic?.title ?? task.epic?.id ?? key` will show raw bead IDs as epic titles whenever `br` omits the parent title.

PR #1205, opened 2026-08-10
[gh-1187 S3] two-artifact worker handoff: demo pane + in-workspace presentation

hachej:
## What

Milestone M1 slice **S3** of the ratified factory plan (`docs/issues/1187/plan.md`) — the **two-artifact worker handoff**, and G5's `pull`.

A worker's final act in-workspace is now: generate the present-pr page → open **both** panes → mark the bead. The owner validates entirely inside the workspace.

### 1. URL / live-demo pane (new UI surface)

Built-in workspace plugin `url-pane`, panel component `url-pane.panel`, params `{ url, title? }`. Registered as a default alongside `filesystemPlugin`, so any host that composes the workspace gets it with zero wiring — same lineage as the HTML viewer.

**Security boundary.** Nothing is fetched server-side, so this is not an SSRF surface; the risk is an agent (possibly prompt-injected) framing an arbitrary origin inside the owner's workspace. Mitigation:

- `shared/urlPane.ts` is the single decision function: `http:`/`https:` only, no credentials in the URL, origin must match an allowlist pattern. Patterns are exact origins or `scheme://host:*` — a **port** wildcard only; host wildcards are deliberately unsupported (`http://localhost.attacker.com` is rejected, and there is a test for it).
- Default = loopback only. `BORING_URL_PANE_ALLOWED_ORIGINS` opts in to more; an explicitly **empty** value means closed.
- Two enforcement points, one function: the front re-resolves before setting an iframe `src` and **fails closed** if `GET /api/v1/ui/url-pane/policy` is unreachable; `POST /api/v1/ui/commands` 400s a disallowed url so an agent gets a reason instead of an invisible blocked pane.
- Sandbox never includes `allow-same-origin`; `referrerPolicy: no-referrer`.

### 2. present-pr artifact as a pane

No product change was needed (the plan's own finding) — the fix is making the path work: present-pr now writes to `.handoff/` **inside the lane worktree**, because the workspace file API is relative-only and a scratchpad page outside the worktree can never be opened as a pane. `exec_ui openFile` on the `.html` resolves to the existing sandboxed HTML viewer.

### 3. Convention wired

`.agents/skills/present-pr` step 5/6 and `.agents/skills/exec` now record the handoff as generate → open both panes → `ask_user` card → mark the bead. `exec` digest repinned (`pnpm check:skill-digests` green). `exec_ui`'s tool description documents the demo pane so a worker discovers it without reading docs.

Docs: `packages/workspace/docs/URL_PANE.md`.

## Proof

- `packages/workspace`: **2081 passed / 10 skipped**, typecheck (front + server projects) clean.
- New tests: allowlist matrix (`javascript:`/`data:`/`file:`, credentials, host-wildcard bypass, fail-closed empty policy), route policy + command rejection, pane renders sandboxed iframe / blocked state / fails closed without a client, plugin registration.
- `lint:invariants`, `lint:workspace-plugin-invariants`, `check:skill-digests` green.
- **Hub proof** (running factory hub, workspace `factory-080c4903`): see the comment below.

## Not in this PR

Merging. S4 (the Inbox intention that links both panes) is the next slice.


hachej:
## Hub proof — both panes, real content

Driven with Playwright against a hub built from this branch (`packages/cli` + front bundle rebuilt), port 5311, workspace `s3-proof` rooted at the lane worktree. **The running factory hub on 5210 (workspace `factory-080c4903`) was not touched** — it is used as the *live URL being embedded*, which is the strongest available demo content.

```
--- policy endpoint ---
{"origins":["http://localhost:*","http://127.0.0.1:*","http://[::1]:*"]}

--- artifact 1: present-pr page in the HTML viewer ---
[openFile/.handoff/pr-1205-presentation.html]        200 {"seq":3,"status":"ok"}

--- artifact 2: live demo pane -> the RUNNING factory hub ---
[openPanel/url-pane.panel http://127.0.0.1:5210/]    200 {"seq":4,"status":"ok"}

--- negative: disallowed origins rejected at the command edge ---
[openPanel https://example.com/]                     400 {"error":"url_pane_origin_not_allowed",
                                                          "message":"Origin https://example.com is not in the URL pane allowlist."}
[openPanel http://localhost.attacker.com/]           400 {"error":"url_pane_origin_not_allowed",
                                                          "message":"Origin http://localhost.attacker.com is not in the URL pane allowlist."}

frames: ["http://127.0.0.1:5311/workspace/s3-proof",
         "http://127.0.0.1:5210/workspace/factory-080c4903"]
```

The second frame URL is the payoff: the factory hub **booted and routed to its own workspace inside the pane** — its chat, its live `[br-1187] owner gate` session, all rendering. The pane chrome (title, URL, reload, open-in-tab) sits above it.

Artifact 1 is this PR's own present-pr page, generated with `--out .handoff/pr-1205-presentation.html` and opened by `exec_ui openFile` into the existing sandboxed HTML viewer — the convention applied to itself, as the plan requires.

### What the drive found (and fixed)

The first run rendered the embedded hub **blank**: with a strict sandbox the frame gets an opaque origin, so every module script it loads is fetched in CORS mode from `origin: null` and refused. A plain Vite dev server — i.e. every real worker demo — would have failed the same way. Fixed in ed05f18 by granting `allow-same-origin` **only when the target origin differs from the workspace's own origin**; a same-origin frame with `allow-scripts allow-same-origin` could remove its own sandbox, so that case keeps the strict attribute. Different port = different origin, so every real demo is covered. Tests lock both directions.

### Known limitation (not fixed here)

present-pr emits its context diagram as `<pre class="mermaid">`, which claude.ai artifacts render natively but the workspace HTML viewer does not — the diagram shows as its source text in the pane. The rest of the page (header, CI tally, sankey, tree, diffs) renders fully. Worth a follow-up bead if the diagram is load-bearing for the owner's review.

### CI

All checks complete, **0 failures**: Lint, Typecheck, Unit Tests, Unit Tests Changed, Invariants, E2E, UI Review, Bundle Budget, Action Pins, Dependency Age, Runtime Refactor P8, Reference Images and Remote Worker Smoke — SUCCESS. (Unit Tests Full / Core Bundle Budget / publish jobs skipped by path filters, as normal for a PR.)

Issue #1167, opened 2026-08-08
[epic #1081] remote-worker nonce store: per-tenant sub-budget (cross-tenant DoS, LOW) + SBX1.4/1.5 audit hand-offs


hachej:
From the #1163 gate-2 Fable audit (PASS-with-findings):

1. **[LOW]** `DEFAULT_MAX_ACCEPTED_CAPABILITY_NONCES = 100_000` is per-registry (per worker), not per-tenant, and fail-closed by rejecting new capabilities. A tenant holding valid capabilities can saturate the shared budget within the 5-min TTL → `REMOTE_WORKER_CAPABILITY_NONCE_STORE_EXHAUSTED` 503 for co-located tenants. Add a per-tenant sub-budget in a later SBX1.3 slice.
2. **SBX1.4 constraint:** post-restart replay is currently fail-closed only because nonce store and binding records share one volatile lifecycle. Any persistence of binding records/receipts MUST land atomically with persistent nonces, plus a regression test asserting a consumed nonce survives simulated restart.
3. **SBX1.5/resolver-slice constraint:** `RemoteWorkerCredentialReferenceSchemaV1` deliberately carries no workspaceId — the future resolver must bind executionId/bindingId/providerId to the authenticated workspace at resolution time.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QtPPbToGDvkARQZuYKFyWN

hachej:
Backlog cleanup 2026-08-26: converted to a deferred idea bead in the beads tracker — `wt-391-forward-idea-nonce-tenant-budget-nh3j` (label `idea`, resurfaces in ~90d or when its promotion trigger fires). Full context preserved in the bead description; reopen there, not here.

Issue #1081, opened 2026-08-05
Epic: sandbox worker runtime — salvage SBX1.3 (Docker+runsc session-lifetime)


hachej:
Post-graduation factory epic. Resurrects the missing middle of the SBX1 sequence. Salvage branch feat/808-sbx1-3-runtime (was PR #916). Absorbs the lane sandbox anchor (PR #1016).

hachej:
Scope addition (owner-ratified): absorb #1060 slice 2 — semantic vs physical runtime identity separation + exact migration authorization (byte preservation, target-before-CAS, serialized locking; forbidden identity hash noted in #1060). Rationale: migration safety only matters when runtime placement actually changes, which is this epic's territory.

hachej:
Folded into PR #1409: docs/plans/multiagent-shell/programs.md is the canonical tracker (architecture already merged via #1220/#1394).

PR #1219, opened 2026-08-10
docs(1081): SBX1.4 internal-first execution plan r1 [owner gate]

hachej:
## What changed

- Adds the owner-gated SBX1.4 execution plan for epic #1081.
- Freezes the one-rented-KVM-VM, internal-first topology and five ordered implementation slices.
- Defines deterministic proof, two-line review, rollback, and explicit non-goals.

## Why

SBX1.3 proved the runsc runtime in process, but the V1 daemon, durable nonce replay defense, admitted-image pin, production box provisioning, and seneca cutover still need an executable owner-ratified sequence.

## Plan slices

1. S1 — minimal daemon, static-secret auth, and runtime proxy
2. S2 — persistent nonce store and #1167 atomicity
3. S3 — qualification-bound image digest pinning
4. S4 — rented-VM provisioning script and manual qualification
5. S5 — seneca production flip with rollback

## Validation

- `git diff --check`
- Structural check: five S1–S5 sections, each with Today, Delta, proof, and rollback
- Opus 4.8 / T2: clean
- Fable / T1: clean

No runtime code or production configuration changed. Do not merge before owner approval.

🤖 Generated with Codex (gpt-5.6-sol).

### DIFF
--- a/docs/issues/1081/plan-sbx14.md
+++ b/docs/issues/1081/plan-sbx14.md
@@ -316,7 +316,11 @@ permanent two-secret mode.
 
 Rotate in this order: (1) install the new secret on the daemon as secondary and
 restart/drain-check it so both old and new verify; (2) switch seneca's issuer
-and verifier to the new secret and confirm a fresh health/create/delete canary;
+to the new secret while its verifier accepts receipts from EITHER secret for
+the duration of the overlap (the daemon keeps signing with its configured
+primary — the old secret — until step 3, so a new-secret-only verifier would
+reject every receipt mid-rotation), and confirm a fresh health/create/delete
+canary;
 (3) wait out the bounded lifetime, promote the new secret to daemon primary,
 drop the old secret, and restart/drain-check again. On suspected compromise,
 stop new admission and retire active sessions before this sequence; overlap
@@ -619,7 +623,12 @@ admission record for that box.
 - Format/mount a dedicated operator-selected data volume as ext4 or XFS with
   project quotas, persist the mount, and make `--check` prove `prjquota` is
   active. The script must require an explicit block device/mount target and
-  refuse `/`, the repository, or an unresolved variable.
+  refuse `/`, the repository, or an unresolved variable. Procurement
+  constraint (Gate 0): the VM SKU must offer an attachable second block
+  volume — verify before purchase on the chosen provider (OVH/Infomaniak/
+  Hikube all sell attachable volumes, but not on every SKU) — and, for the
+  Tailscale cert path, MagicDNS + HTTPS certificates must be enabled on the
+  tailnet.
 - Build/install `/usr/local/libexec/boring-workspace-quota` as root-owned,
   non-writable by the daemon's callers, with its digest in the qualification
   bundle. Configure the worker service, nonce-state directory, workspace root,
@@ -769,9 +778,11 @@ production canary.
 - Set `qualificationMaxAgeMs` explicitly to seven days for this internal-first
   box. The operator re-runs S4 qualification at least every six days and after
   any kernel, Docker, runsc, daemon/provider, helper, policy, or image change.
-  The named owner is the seneca production operator; provisioning creates a
-  recurring six-day operations-calendar reminder assigned to that owner and
-  links it to the current admission record and runbook. Missing the reminder
+  The named owner is the seneca production operator. Creating the recurring
+  six-day requalification reminder is a MANUAL admission-record step the
+  operator performs at first admission (the provisioning script prints the
+  exact reminder text and the admission-record/runbook links to include; it
+  does not pretend to reach a calendar). Missing the reminder
   is not a waiver: expiry still fails create closed.
   Installing refreshed immutable evidence requires draining the worker and
   restarting the daemon; old evidence remains the rollback artifact. New
--- a/docs/issues/1081/plan-sbx14.md
+++ b/docs/issues/1081/plan-sbx14.md
@@ -1,0 +1,780 @@
+---
+github: https://github.com/hachej/boring-ui/issues/1081
+issue: 1081
+state: needs-owner-approval
+updated: 2026-08-10
+revision: r1
+flag: BORING_AGENT_MODE=remote-worker
+track: owner
+---
+
+# gh-1081 — SBX1.4 internal-first sandbox service execution plan r1
+
+## Outcome
+
+Ship one owner-operated EU sandbox worker for seneca production. Seneca keeps
+the agent/control plane; `BORING_AGENT_MODE=remote-worker` sends each agent
+session to a Docker container running under gVisor `runsc --platform=systrap`
+on one rented KVM VM. The worker exposes the already-merged remote-worker V1
+protocol through a minimal HTTP daemon authenticated by one static shared
+bearer secret.
+
+The first production admission is deliberately manual: provision the rented
+VM, run the repository's committed qualification harness on that exact box and
+artifact cohort, require green evidence, then copy the resulting immutable
+digests into the single-worker config consumed by seneca and the daemon. There
+is no fleet controller in v1.
+
+Success is a seneca production canary session whose filesystem and exec calls
+run on the EU worker, whose in-sandbox `uname -r` reports the gVisor sentinel,
+and which can be returned to `vercel-sandbox` by one environment rollback.
+
+## Ratified topology and trust boundaries
+
+```text
+seneca production (agent/control plane)
+        |
+        | HTTPS + one pre-shared bearer secret
+        v
+one rented EU KVM VM (OVH for v1)
+        |  provider hypervisor = boundary 1
+        v
+minimal V1 worker daemon -> Docker -> one runsc/systrap sandbox per session
+                                      runsc sandbox = tenant boundary
+```
+
+- The VM is rented from a provider. It is **not** a self-hosted nested VM on
+  bare metal. Infomaniak or Hikube are acceptable future CH placements; OVH is
+  the ratified EU placement and seneca target for v1.
+- `runsc --platform=systrap` needs no `/dev/kvm`. The 2026-08-10 proof ran the
+  same Docker+runsc shape inside a nested KVM guest and passed all 11 hostile
+  probes, so v1 does not provision, require, or pass through `/dev/kvm`.
+- One runsc sandbox is created per agent session. For the internal-first
+  deployment, that is the tenant boundary. VM-per-tenant is a later enterprise
+  placement option, not a v1 requirement.
+- The VM's OS is controlled by us. An openat2-capable host/runsc cohort, a
+  dedicated `prjquota` data volume, and the root quota helper are provisioning
+  facts, not application-level fallbacks. Missing facts fail qualification and
+  keep the box out of use.
+- The one static secret is the sole v1 authentication root. S1 supplies the
+  currently missing shared codec that lets the V1 client issue short-lived,
+  request-bound bearer capabilities and lets the daemon authenticate them and
+  sign binding receipts. There is no user database, RBAC service, tenant
+  credential table, or second token class.
+
+## Today / target delta
+
+| Area | Today on `origin/main` | SBX1.4 delta |
+| --- | --- | --- |
+| Remote protocol | `createRemoteWorkerProvider.ts`, `protocolClient.ts`, and `pairProxies.ts` implement the V1 client/provider against an abstract transport. | A real bounded HTTP/SSE transport and daemon implement the V1 endpoints. |
+| Runtime | `RunscSessionRuntimeV1` creates, execs, mutates files, renews, and retires real Docker+runsc sessions in-process. | The daemon authorizes and proxies every V1 operation to this runtime; no second runtime is built. |
+| Replay | `SingleUseNonceStoreV1` is a registry-scoped in-memory `Map` plus expiry heap; restart loses consumed nonces. `bindingRegistry.ts` is also volatile. | Only nonces become durable, atomically consumed, globally bounded, and per-workspace bounded. Bindings/receipts stay volatile. |
+| Image admission | The create request carries expected digests, but `sessionRuntime.ts:startContainer` runs the supplied image without comparing it to admitted evidence. | Startup verifies the manually admitted bundle/evidence and every container start/replacement requires its exact `repository@sha256:...` image. |
+| Box readiness | The nested-KVM proof is green, while this week's host integration is honestly non-admitting because its older runsc returns `ENOSYS` for `openat2` and its volume lacks `prjquota`. | The rented VM is provisioned with a known-good systrap runsc, openat2, `prjquota`, and the installed root helper; the committed harness is green on that exact box. |
+| Seneca | Production safety accepts `vercel-sandbox`; a legacy V0-style remote-worker adapter exists but `remote-worker` is not a built-in mode. | Seneca composes the V1 `@hachej/boring-sandbox` provider and flips behind `BORING_AGENT_MODE=remote-worker`; the prior mode remains the immediate rollback. |
+
+## Decisions that constrain every slice
+
+1. One epic branch may carry the plan, but implementation is five ordered,
+   independently reviewed PRs. Each PR remains within the normal review budget
+   (about 1,500 added production lines); if it cannot, the implementer returns
+   to the owner instead of silently widening a slice.
+2. The daemon owns trusted filesystem roots. No HTTP request may supply a host
+   path, Docker socket, image override, or qualification override.
+3. Production uses HTTPS. Plain HTTP is allowed only for in-process/loopback
+   tests. TLS termination and the server certificate are box configuration;
+   the V1 fleet config continues to pin CA and server name.
+4. Stable V1 schemas, request-size ceilings, capability lifetime, binding
+   checks, error codes, retry semantics, startup sweep, and hard expiry are
+   reused. The daemon does not invent a parallel wire protocol.
+5. Bindings and receipts remain in memory. Under issue #1167, a future change
+   that persists either must persist nonces in the same atomic change and must
+   repeat the restart regression. That future change cannot be split across
+   PRs.
+6. Qualification is manual per box. A green run is necessary but is accepted
+   only when it names the exact worker/provider/workload cohort used by the
+   daemon and seneca config. Drift fails closed at daemon startup and provider
+   health comparison.
+7. No real seneca traffic reaches the box until S1-S4 are merged, S4 evidence
+   is green, and S5's exact configuration is owner-approved.
+
+## Gate 0 — prove openat2 before funding implementation
+
+This is an operator prerequisite, not a code PR or a sixth delivery slice. The
+2026-08-10 nested-KVM proof established systrap viability but did not exercise
+the workspace helper's mandatory `openat2` call. The older host run proved that
+`release-20260706.0` returns `ENOSYS`; no supplied evidence yet proves a runsc
+release that admits the real workspace path.
+
+Before S1 begins, provision a disposable rented KVM VM, start with
+`release-20260803.0`, and run the existing SBX1.3 integration preflight against
+candidate releases until the `workspace-openat2-fs` and `symlink-swap-race`
+follow-ups are absent and both `workspace-fs` (with `openat2Probe: true`) and
+`session-create` are passed. Record the exact release tag, binary digest, host
+kernel, guest sentinel, and raw output beside the plan. If no candidate passes,
+stop: the owner chooses between pinning a patched gVisor build or deferring
+SBX1.4 as refusal-only. No realpath fallback and no waiver are permitted.
+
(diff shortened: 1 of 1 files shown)


hachej:
L1 review (Opus): **PASS-with-required-changes**

Read-only review of `docs/issues/1081/plan-sbx14.md` @ `agent/docs-sbx14-plan`, against the SBX1.4 scoping doc, issue #1167, the owner rulings, and the merged SBX1.3 code on `origin/main`.

This is a strong plan. Every attach point it names is real, its Today column is factually accurate (I checked each claim against `origin/main`, not against the prose), Gate 0's assertions map to probe ids that actually exist, and the #1167 atomicity constraint is honored exactly in its cheapest-compliance form. The required changes below are doc-level; none invalidate the slice order.

## Verified true (spot-checks, not vibes)

- `sessionRuntime.ts` — `startContainer` is at line 656 with a second call site at 684 (clean replacement). S3's "top of `startContainer()` covers initial start *and* replacement" is correct.
- `singleUseNonceStore.ts` — `consume(nonce, expiresAtMs, nowMs)` is 3-arg in-memory + expiry heap. S2's 4-arg port change is an accurate delta.
- `bindingRegistry.ts:204` constructs `new SingleUseNonceStoreV1(...)` internally — there is no injection seam today, so S2's "introduce an injectable constructor option" is real work, not a restatement.
- `transport.ts` is interface-only and already carries an in-code `SBX1.4 transport must…` comment — S1 attaches where the code expects it.
- All qualification scripts exist. `build-qualification-bundle.mjs` takes `process.argv[2]` as `<cohort-spec.json>` — exactly the S4 invocation. `RUN_RUNSC_INTEGRATION=1` is the real env gate.
- Gate 0's green criteria name real probe ids: `workspace-openat2-fs`, `symlink-swap-race`, `session-create` vs `session-create-fail-closed`, `workspace-fs` with `openat2Probe: true`, `operator-follow-up` status. Gate 0 is runnable as written.
- **The S5 precedence bug is real.** `packages/core/src/app/server/createCoreWorkspaceAgentServer.ts:1226-1234`: `workerBaseUrl` selects `createRemoteWorkerModeAdapter` before `BORING_AGENT_MODE` is ever consulted. Catching this before it silently routed V1 traffic through the V0 adapter is the single most valuable thing in this document.
- **#1167 constraint 2 honored exactly**: nonces durable, bindings/receipts explicitly NOT persisted, the regression test named verbatim ("consumed nonce survives simulated restart"), and a forward rule blocking any future binding-persistence PR that doesn't share one atomic transaction. This is the cheapest compliant shape and the plan says so.
- **S4 carries all three environment facts**: openat2 (host + gVisor proof), `prjquota` (format/mount + `--check` proves active + fill/sibling/host-reserve), systrap (`runsc install -- --platform=systrap` as an observable runtime arg).
- **No silent multi-tenant creep.** The non-goals section is unusually disciplined, and the per-workspace nonce sub-budget is pre-emptively framed as an availability control rather than a tenancy claim.
- **Rollback session-history story is honest and correct.** Transcripts stay on `BORING_AGENT_SESSION_ROOT` (AGENTS.md rule 9), and the plan explicitly refuses to claim remote workspace writes reappear in a new Vercel sandbox.

## Findings

### H1 — [HIGH] The daemon's exposure and threat model are the plan's biggest hole
The daemon is a brand-new internet-facing-or-not attack surface and the plan never decides which. Decision 3 says "Production uses HTTPS"; S4 says "keep the HTTP listener loopback-only behind TLS termination". Neither names **what terminates TLS**, **who is allowed to connect at the network layer**, or **what the blast radius of the one static secret is**.

That blast radius is severe and unstated: S4 ratifies running the worker service **as root** with Docker CLI authority, and the static secret is "the sole v1 authentication root" from which the client mints capabilities. Therefore *secret compromise ≈ root on the VM*. A plan whose entire purpose is a tenant boundary must say that out loud and enumerate the compensating controls.

Also missing at S1: auth-failure rate limiting / backoff, connection and concurrent-session caps, and slowloris/idle-connection bounds. Body size (8 MiB) and SSE lifetime are bounded; connection volume is not.

Also missing: a secret **rotation procedure**. S2's rollback says "rotate the static secret" but no mechanism exists — with a single-valued secret, rotation is a synchronized seneca+daemon restart with a hard failure window unless the daemon accepts two secrets during an overlap.

### H2 — [HIGH] S3 is over-scoped and breaches the plan's own review budget
Decision 1 caps a slice at ~1,500 added production lines and requires the implementer to return to the owner rather than widen silently. S3 is *already* widened at plan time: it contains the actual security gate (admitted-cohort load + digest equality before `docker run`) **plus** a full `qualify-docker-runsc-isolation.mjs` V2→V3 envelope rework, a new observe/bound two-mode CLI, real quota-helper wiring replacing the no-op stub, a new qualification-workspace-root input with `findmnt` validation, and an external-image mode for the integration script. The scoping doc sized the pin gate alone as S-M; this is comfortably L.

The pre-declared split should be in the plan, not discovered mid-implementation.

### M3 — [MEDIUM] S2's durability claim is process-restart-only unless pragmas are pinned
S2 specifies `BEGIN IMMEDIATE` but no `journal_mode` and no `synchronous` setting. Under WAL + `synchronous=NORMAL`, a power loss or hard VM reset can lose the last committed transactions — i.e. a consumed nonce can resurrect, which is precisely the property S2 exists to guarantee. Additionally, the cross-process fencing claim (and its "two connections, exactly one `accepted`" test) depends on the DB file living on a **local** filesystem; SQLite locking is unsafe on NFS/network volumes. Neither is stated.

### M4 — [MEDIUM] S4's proof transcript depends on a CLI/env contract S3 never names
S4's admission block is otherwise the best-specified proof in the document, but it invokes `--observe-only`, `--cohort-spec-out`, `--workload-image`, `--qualification-bundle`, and `BORING_RUNSC_WORKLOAD_IMAGE`, `BORING_RUNSC_WORKSPACE_ROOT`, `BORING_RUNSC_USE_INSTALLED_QUOTA_HELPER`. I confirmed **none of these exist on `origin/main`** — they are S3 deliverables described only in prose ("an explicit observe-only mode", "a narrow opt-in", "one explicit qualification workspace-root input"). If S3 ships the same capabilities under different flag names, S4's transcript is not runnable and the operator improvises during a privileged admission run. S3 must own the exact contract as a named deliverable.

### M5 — [MEDIUM] S5's "export canary data before rollback" has no mechanism, and in-flight sessions at flip time are unhandled
The rollback story is honest that workspace-only writes do not follow the env flip, and then requires "export required canary data before finalizing rollback" — but the V1 protocol has no bulk export, only per-file fs reads, and the plan names no alternative (root-side `docker cp` / host bind-mount copy on the VM). As written this is the one step in an otherwise concrete rollback that is a vibe. Separately, nothing says what happens to a session that is mid-work when the environment revision is restored.

### L6 — [LOW] #918 gate (b) is re-worded, not just implemented
Gate (b) said "persist/**boot-epoch** the capability nonce". S2 drops the epoch column and argues transactional global uniqueness subsumes fencing. I agree technically — for a single local DB file, `BEGIN IMMEDIATE` is strictly stronger than an epoch column, and the concurrent-connection test is the better proof. But this is a change to recorded gate wording and should be surfaced at the owner gate rather than resolved inside a slice body.

### L7 — [LOW] The per-workspace nonce sub-budget is an accepted addition over the minimal path
The scoping doc listed 1.4-D as deferrable (LOW, availability-only, moot for a single-tenant canary). S2 includes it. This is cheap and sensible co-located with the storage change — just label it as a deliberate addition so the "narrowed 1.4" framing stays honest.

### L8 — [LOW] S4 transcript hygiene and an unowned calendar commitment
`sudo -i` followed by per-command `sudo` is redundant (the plan explains why, which is fine), but `build-qualification-bundle.mjs` is the single unsudoed command writing under root-owned `/var/lib/boring-worker` — inconsistent enough to cause a copy-paste stumble. Separately, `qualificationMaxAgeMs = 7 days` with "operator re-runs at least every six days" is a recurring human commitment with no named owner or reminder mechanism; a missed window fails create closed in production.

## Required changes (machine-actionable)

1. **[H1]** Add a `## Daemon exposure and threat model` section to the plan stating, at minimum: (a) the static secret's blast radius is host-root-equivalent because the daemon runs as root with Docker authority; (b) the named TLS terminator (e.g. nginx/Caddy on the box) and where the cert comes from; (c) the network-layer restriction — provider firewall allowlist of seneca egress IPs, or a WireGuard/Tailscale overlay with the listener bound to the overlay interface — plaintext-on-private-net must be explicitly ruled in or out, not left implicit; (d) that mTLS was considered and deferred, with the reason.
2. **[H1]** Add to S1 Delta: auth-failure rate limiting/backoff, a max-concurrent-connection and max-active-session cap, and idle/header timeouts. Add a corresponding acceptance bullet ("repeated bad-bearer requests are throttled before runtime calls").
3. **[H1]** Add a `### Secret rotation` subsection to S1 or S4: daemon accepts a primary + optional secondary secret during a bounded overlap window; document the rotate order (daemon adds secondary → seneca switches → daemon drops old) and reference it from S2/S3 rollback text instead of the bare phrase "rotate the static secret".
4. **[H2]** Split S3 in the plan into **S3a — admitted-cohort load + `startContainer` digest pin** (the security gate; the `qualificationAdmission` / `sessionRuntime` / `dockerArgv` / `fleetAdmission` tests) and **S3b — real V3 harness upgrade** (observe/bound modes, quota-helper opt-in, workspace-root input, external-image mode, `isolationEvidenceDocker` tests). Update the slice-order diagram and note that S4's admission run is blocked on S3b, while S3a alone unblocks fail-closed posture.
5. **[M3]** Add to S2 Delta: pin `journal_mode` and `synchronous=FULL` (or explicitly downgrade the durability claim to process-restart-only and say so), and require the nonce DB to reside on a local (non-network) filesystem, asserted by S4 `--check`.
6. **[M4]** Add to S3 (S3b after the split) an explicit "CLI and environment contract" bullet list naming every flag and env var S4's transcript uses, verbatim: `--observe-only`, `--cohort-spec-out=<path>`, `--workload-image=<repo@sha256:…>`, `--qualification-bundle=<path>`, `BORING_RUNSC_WORKLOAD_IMAGE`, `BORING_RUNSC_WORKSPACE_ROOT`, `BORING_RUNSC_USE_INSTALLED_QUOTA_HELPER`, `BORING_BUSYBOX_BINARY`. State that `build-qualification-bundle.mjs <cohort-spec.json>` (positional, already exists) is the consumer of `--cohort-spec-out`.
7. **[M5]** Replace S5's "export required canary data" with a named mechanism (root-side copy from the daemon-owned bind-mount source on the VM, or a bounded protocol fs-read loop) and add a step: drain/close all canary sessions before restoring the environment revision, so no session is in flight across the flip.
8. **[L6]** Add one line to `## Owner gate / next action` flagging that approving r1 also ratifies replacing #918 gate (b)'s boot-epoch requirement with transactional cross-connection nonce uniqueness.
9. **[L7]** Label the per-workspace nonce sub-budget in S2 as a deliberate addition over the scoping doc's minimal path (#1167 finding 1, previously deferrable).
10. **[L8]** Make the S4 transcript's privilege usage uniform (drop the per-command `sudo` under `sudo -i`, or add it to the `build-qualification-bundle.mjs` line), and name an owner plus reminder mechanism for the six-day requalification cadence.

Items 1-7 are required before owner approval. Items 8-10 are strongly recommended.

hachej:
L1 required changes applied (items 1-7; Sol's run was interrupted at verification — edits completed and pushed by the orchestrator). Point-by-point: H1 → new daemon threat-model section (root+Docker authority stated plainly, transport decided, auth-failure rate limiting, connection caps, static-secret rotation procedure); H2 → S3 split into S3a (pin gate) / S3b (harness rework, CLI/env contract); M3 → SQLite WAL + synchronous=FULL pinned, local-filesystem requirement stated; M4 → S3b names the exact flags/env vars S4's transcript invokes; M5 → root-side canary archive after sessions drain + in-flight session handling at flip. L2 (Fable) review proceeds on this head.

hachej:
L2 review (Fable): **PASS-with-required-changes**

Final line-2 review of `docs/issues/1081/plan-sbx14.md` @ `e17242958` (post-L1 head). I re-read the full document and verified each of L1's seven required items against the actual text, not the applied-changes comment. Verdict: the plan is owner-ready after one small required fix — the new secret-rotation procedure, as written, fails its own step-2 canary.

## L1 resolution audit — all seven genuinely applied, not claimed

1. **H1 threat model** — real and specific. The section names the blast radius out loud ("static secret is host-root-equivalent"), decides transport concretely (Tailscale-only ingress, provider firewall denies public, Caddy terminates TLS on `tailscale0` with a `tailscale cert` certificate, daemon loopback-only, plaintext forbidden anywhere but loopback), rules plaintext-on-private-net explicitly out, and records the mTLS deferral with a sound reason and a reconsideration trigger (ingress outside the tailnet). The Tailscale-over-firewall-allowlist rationale (no stable seneca egress IPs) is correct for a Vercel-hosted control plane. S4 `--check` proves the bind, cert, ACL, and no-public-listener facts. This is a decided design, not a gesture.
2. **H1 connection/auth bounds** — in S1 Delta with concrete numbers (128 conns, 32 sessions, 10 s header, 15 s idle, 10 failures/min token buckets with bounded key state) and matching acceptance bullets including slowloris. Applied.
3. **H1 rotation** — a real primary/secondary procedure with bounded overlap and startup guards exists, and S2/S3a rollback text references it instead of the bare phrase. Applied — but the ordering is defective (Finding 1 below).
4. **H2 split** — S3a (pin gate, S-M) / S3b (V3 harness, M) with the slice diagram updated and the correct dependency statement: S3a alone gives fail-closed posture, S4's admission run blocks on S3b. Applied.
5. **M3** — `journal_mode=WAL` + `synchronous=FULL` pinned per connection, startup fails if the pragma doesn't take, local-filesystem requirement stated with named forbidden classes, and S4 `--check` asserts the resolved DB path's filesystem type. Applied, and the Node `>=22.19.0` / flagless `DatabaseSync` pin is a good addition beyond the ask.
6. **M4** — S3b carries a verbatim "CLI and environment contract" block naming every flag/env var S4's transcript uses, including the `--cohort-spec-out` → positional `build-qualification-bundle.mjs` handoff and the retained `RUN_RUNSC_INTEGRATION=1` gate. The S4 transcript uses exactly these names. Applied.
7. **M5** — replaced with a named mechanism (root-side copy of the daemon-owned bind-mount source into digested archives under `/var/lib/boring-worker/rollback-exports/`) plus an explicit drain/close-before-flip step, mirrored in the end-to-end rollback story. Applied.

L8's transcript is also now uniformly root (`sudo -i`, no mixed `sudo`), and the requalification cadence has a named owner. L6 and L7 lines are present where L1 asked.

## L2 findings

### 1 — [REQUIRED] The rotation order breaks receipt verification at its own step 2

S1's rotation procedure: daemon signs receipts **with the configured primary** throughout the overlap; step 2 says "switch seneca's issuer **and verifier** to the new secret and confirm a fresh health/create/delete canary." At that moment the daemon's primary is still the old secret (promotion is step 3), so it signs binding receipts with the old key while seneca verifies with only the new one — the step-2 canary's receipt verification fails, and the operator is stranded mid-rotation with a procedure that says the failure is impossible. The daemon-side dual acceptance is specified; the client-side is not.

Fix is one sentence, either form: (a) seneca's receipt **verifier** accepts both secrets during the bounded overlap while its issuer moves to the new secret, or (b) the daemon signs receipts with the **newest** configured secret (the secondary) during overlap, so a seneca that has switched verifies cleanly and one that hasn't still verifies via its own dual window. Whichever is chosen, the step-2 canary must be the proof that the chosen form works.

### 2 — [RECOMMENDED] S4 assumes an attachable second block device; verify the SKU before Gate 0

`--apply` requires an explicit `--data-device` and refuses `/` — correct — but many rented KVM VPS lines ship exactly one disk. On OVH this silently constrains the purchase to Public Cloud instances (or VPS+additional-disk options) where a block volume can be attached. One line in S4 ("the rented VM SKU must provide an attachable dedicated block volume; verify before Gate 0 provisioning") turns a mid-provisioning dead end into a purchase-time check. Same applies to future Infomaniak/Hikube placements. Also note `tailscale cert` requires MagicDNS + the tailnet HTTPS feature enabled — worth one clause in `--check`'s prerequisites so the cert step can't surprise the operator.

### 3 — [RECOMMENDED] "Provisioning creates a recurring six-day operations-calendar reminder" is the one non-executable step in S4/S5

A shell provisioning script cannot create a calendar entry. The owner is named (seneca production operator) and the backstop is honest (expiry fails create closed regardless), so this is not a gap in the security story — but the plan should either name the actual mechanism (ops-calendar entry created manually as part of the admission record checklist, or a beads/routine reminder) or say plainly that the reminder is a manual admission-record step. As written it reads like automation that won't exist.

### 4 — [NOTE, no change] Per-source auth rate limiting is near-vacuous in this topology — and that's fine

With the tailnet ACL admitting only seneca, the "by tailnet source address" limiter has effectively one key; the server-wide bucket is the operative control. The forwarded-header handling (reject on non-loopback) is correct and becomes load-bearing only if ingress ever widens, which is exactly when the threat model says to revisit. No change needed; recording so the owner doesn't over-credit the per-source limiter.

## L2 lens rulings

**Trust chain (coherent):** secret custody is symmetric HMAC shared by exactly two parties; capabilities minted client-side add request-binding, not authority separation — and the plan says so honestly ("sole authentication root"). Reaching the listener requires tailnet membership under an ACL naming only seneca, so secret theft alone is insufficient without network position, and network position alone (including a Tailscale coordination-plane compromise adding a rogue node) is insufficient without the secret. Two independent factors; no single silent point defeats the runsc boundary. Revocation = stop admission + retire sessions + rotate, and the plan correctly states overlap restores availability, not trust in a stolen secret. The one residual single point is seneca itself, which holds the secret and legitimate reach — unavoidable in any topology where seneca is the control plane, and correctly out of scope.

**Operational honesty:** Gate 0's stop condition (no waiver, owner chooses pin-patched-gVisor vs defer) is the right shape for the openat2 risk; S4's environment facts are all present and provable via `--check`; the transcript is committed and now backed by a named CLI contract. The two real ops risks are Findings 2 and 3, both recommended-level.

**Strategic fit (clean):** the provider/protocol contract is untouched — the daemon implements the V1 routes `protocolClient.ts` already names, the nonce store is already keyed by authenticated `workspaceId`, `sandboxId` derivation and the domain-separated codec are server-side details a per-tenant issuance layer can replace without touching endpoints, and the placement config already speaks in buckets. I find **no v1 decision that would need redoing for multi-tenant** rather than extending; the swap points are the secret-derivation root and the ACL, both replaceable in place. No gold-plating either — the one addition over minimal (per-workspace nonce sub-budget) is labeled as such per L7.

**L6 ruling — the #918 gate (b) rewording is genuinely stronger, not a weakening.** Boot-epoch fencing delivers "restart invalidates everything issued before it"; durable transactional uniqueness delivers "a consumed nonce is never accepted twice — across restarts *and* across concurrent processes/connections," which epoch tagging alone does not give you. The property lost is incidental: unconsumed, unexpired pre-restart capabilities remain valid after a restart under the new scheme, where epoch-fencing would have killed them. That was a side effect of the mechanism, never gate (b)'s security claim — those capabilities are legitimate, short-lived, and request-bound, and the compromise case is handled by rotation, not reboot. With the no-volatile-fallback + fail-closed-on-corrupt-DB rule, the owner is ratifying a strict improvement on the property that matters. The explicit ratification line in the owner gate is the right disclosure. **No silent weakening — approve as worded.**

## Verdict

**PASS-with-required-changes.** Finding 1 (rotation ordering) is required before owner approval — it is a one-sentence fix to a procedure the plan itself designates as the rollback path for S2/S3a boundary crossings, so it must be correct on paper. Findings 2-3 are recommended; Finding 4 is a note. Everything else, including the #918 rewording, is fit for the owner gate as written.

hachej:
L2 findings applied: (1-required) rotation step 2 now keeps seneca's verifier dual-accepting during the overlap while only the issuer switches — receipts signed by the daemon's still-old primary verify throughout; (2) Gate-0 procurement constraints added (attachable second block volume on the chosen SKU; MagicDNS+HTTPS for tailscale cert); (3) the six-day requalification reminder is now an honest MANUAL admission-record step (script prints the text; no fictional calendar integration). With L2's L6 ruling (transactional uniqueness = strict strengthening, ratifiable as worded), **PR #1219 has now passed the full review chain: Sol authored → Opus L1 (7 required items applied) → Fable L2 (1 required + 2 recommended applied) → ready for the owner gate.**

hachej:
Superseded by #1220 (owner directive: unify the SBX1.4 execution plan and the sandbox architecture into one owner-gate PR). The full plan content (docs/issues/1081/plan-sbx14.md) is now on the #1220 branch, with its adversarial review lineage preserved in the plan header and docs/issues/1081/README.md index — L1 (Opus, e17242958) + L2 (Fable, 0aedd1a9d), whose review comments live in this PR's history. Nothing lost; #1220 is the single sandbox plan gate.

PR #1220, opened 2026-08-11
[sovereign-fleet] docs: sandbox bridge + sovereign fleet architecture [owner gate] (#1081)

hachej:
## Summary

Reconciles the #1081 sandbox plan to the owners final decisions. Docs only; no runtime code or configuration changed. Do not merge before explicit owner approval.

## Final v1 architecture

- Primary bridge: Blaxel hardware-isolated microVMs behind SandboxProviderV1; regular Volume is a phase-0 persistence exception. E2B managed is the alternate. Modal, Daytona, and Beam are trusted-pilot-only.
- Owned target: containerd -> Kata Containers -> CI-pinned Firecracker, one hardware/KVM microVM per sandbox on shared EU bare metal. Share the host, never the boundary.
- Storage product: self-hosted Apache-2.0 SeaweedFS, exposing the same plain tenant files through POSIX FUSE and S3. Durable /workspace stays user-browsable and versioned; local /scratch handles POSIX-heavy work.
- File events: guest inotify plus S3 notifications feed an at-least-once watch stream with dedupe, monotonic cursors, reconnect, and gap reconciliation. No polling correctness path.
- Runtime behavior: one sandbox per user session, tenant/externalId tags, idempotent create, session pool, 60-second idle suspend, fast resume, heavy-image prewarming, network-off with SeaweedFS-only egress, metering, and fail-closed graceful degradation.

## Build and cutover

The plan now follows Gate 0 -> adopt Blaxel and build SeaweedFS + Kata/Firecracker in parallel -> qualify exact cohorts -> canary and cut over by configuration. Active sandbox-hours and egress are metered from day one. Blaxel is about $4-41/month at pilot/growth; one AX102 is about $278/month flat; the raw crossover is about 3,357 active sandbox-hours/month, with approximately 3,000 hours/month as the planning trigger.

V2 keeps SandboxProviderV1, the hardware-microVM isolation class, guest capability shape, and SeaweedFS data plane. It adds an owned snapshot-fork fleet, warm pools, and snapshot-locality scheduling, likely using Cloud Hypervisor. That is an internal VMM swap, not a storage/data-product or contract re-platform.

## Security invariants retained

- arXiv:2606.08433 is applied only within its stated single-tenant scope; shared-kernel gVisor is disqualified as the public tenant boundary.
- gVisor remains optional inner defense-in-depth; microsandbox/libkrun remains rejected.
- Short-lived tenant-prefix/action-scoped credentials; no control-plane secrets or other-tenant data on sandbox hosts.
- CI-owned Firecracker pins, separate Spectre/MDS workstream, fail-closed startup, default-deny egress, 11 hostile probes, quotas, rollback, and Gate 0.

## Files

- Rewrote docs/issues/1081/plan-sbx14.md.
- Rewrote docs/issues/1081/tech-choice.md.
- Updated docs/issues/1081/README.md.
- Added the requested managed-provider and cost references plus their Daytona link dependency.

## Proof

- pnpm lint:invariants: pass.
- git diff --check: pass.
- Local Markdown link validation across the six committed docs plus the existing security evaluation: pass.

## Review

- Independent owner-spec review: clean after two passes.
- Required fixes applied: stale-doc routing, exact /workspace vs /scratch event semantics, S3 event dedupe/cursor/gap recovery, phase-0 Volume-to-SeaweedFS transition, literal host-data invariant, and explicit v2 VMM-swap wording.
- Reviewed SHA: 1715230b9.

## Risk / rollback

This is documentation only. No production runtime changes. PR remains an owner gate and must not be merged without explicit approval.

## Owned build plan update

Added `docs/issues/1081/plan-own-sandbox.md`: a dependency-ordered, checklist-style execution plan from Gate 0 through cutover, with M0/M1/M2 milestones, per-phase acceptance and effort, exact AX102 provisioning, Kata/Firecracker and SeaweedFS tasks, guest-daemon/provider wiring, the 11 hostile probes, rollback, operability, and an explicit M0 definition of done. The live `spike/own-sandbox` lane is cited as pending Gate 0 evidence; no uncommitted spike result is represented as passed.

Validation for this update: `pnpm lint:invariants` and `git diff --cached --check` passed.


hachej:
Adversarial review (Fable): unified sandbox plan — architecture + §9 grounded API + references, reviewed as one owner-gate. Scope per instructions: NOT re-reviewing S1–S5 mechanics (already L1-Opus/L2-Fable'd); attacking the newly unified material and the plan as a whole.

## Verdict: **PASS-with-required-changes**

The four-layer architecture, the build-verdict, and the discipline guardrails are sound and the references are unusually honest (UNVERIFIED flags, refusal-only admission named as the biggest blocker). But the unified doc makes three claims it cannot currently back — the "grounded" citation for the auth divergence, the "same public API" dogfood claim as applied to v1 auth, and the v1→v2 exit criteria as a public-opening gate — plus one direct architecture↔plan contradiction. None kill the direction; all must be fixed before ratification, because this is the document future decisions will cite.

---

## REQUIRED

**R1 — The Docker-AuthZ CVE citation is broken and circular (§9.4, §5 OWN row).**
§9.4 cites `(ref: [SURVEY] — Docker-AuthZ CVE finding)` and the §9 grounding key describes `[SURVEY]` as "(Docker-AuthZ CVE / build-vs-adopt)". **`references/build-vs-adopt-survey.md` contains no Docker-AuthZ CVE finding at all** (grep: zero hits for "AuthZ"; its only CVE-adjacent text is "narrow authz" in an unrelated sentence). The actual chain is circular: architecture §9.4 → `[SPEC]` §3.1 → "the Docker-AuthZ CVE lesson (`sandbox-service-architecture.md` §5 OWN row)" → back to the architecture doc. The security keystone of the whole OWN disposition — "own your security edge" — is grounded in nothing but itself. The lesson is probably right (CVE-2024-41110, Docker AuthZ plugin bypass, is real), but a doc that declares "No decision here is ungrounded" (§9, line 346) may not dress an ungrounded assertion as cited. Fix: cite the actual CVE with a primary source in a reference file, or delete the "grounded in the CVE lesson" framing and own it as a design judgment.

**R2 — Architecture contradicts the plan on replay defense: boot-epoch fencing vs transactional uniqueness.**
Architecture §1 exit-criterion 3 and §9.4 both state consumed nonces "survive a daemon restart via **boot-epoch fencing** (SBX1.4-C / #1167)". The plan's S2 says the opposite, explicitly: "The earlier **boot-epoch proposal is intentionally subsumed** by transactional global nonce uniqueness … V1 stores no epoch column", and the owner gate "explicitly ratifies replacing #918 gate (b)'s boot-epoch requirement". `[SPEC]` §3.1 repeats the stale boot-epoch wording too. Ratifying #1220 as-is ratifies both the mechanism and its replacement in the same PR. Fix the architecture doc and spec to the S2 mechanism (which is the reviewed one).

**R3 — The dogfood rule's "same API a public customer would use" is false at the auth layer, and the exit-criteria grep claim is unenforceable as written (§3, exit-criterion 5, §9.4).**
Same *verbs*, yes. Same *trust position*, no. In v1 Seneca **holds the daemon's static shared secret** — which the plan's own threat model calls **host-root-equivalent** — and **mints its own capabilities client-side** (S1: "client-issued capability → daemon verification"). A public customer would never hold that secret; they'd present an API key to an edge **compat shim that mints capabilities server-side** (§9.4) — a component that is **not scheduled in v1, v1.1, or v2** anywhere in §9.6 or `[SPEC]` §5. So the one component that makes Seneca's path and a public customer's path actually identical does not exist and has no build slot; until it does, Seneca *is* the privileged special path, by construction, in exactly the capability-issuance dimension §9.4 sells as the product's security differentiator. Additionally:
- Exit-criterion 5 claims "No `Seneca-special` bypass exists … (grep-enforceable; see §3)" — but §3 only says to *prefer* a grep-enforceable invariant; no grep pattern, no `check:invariants` rule, and no slice implements one. As written, criterion 5 cannot be evaluated.
- A concrete bypass the rule already fails to catch exists **today**: S5 documents that Core "currently selects that legacy adapter when `BORING_WORKER_BASE_URL` is present **before** it resolves `BORING_AGENT_MODE`", silently bypassing V1. No grep for "Seneca" finds that — it's a config-precedence privilege, not a named flag.
Fix: (a) restate §3 honestly — v1 dogfoods the *wire contract and execution path*, not the public auth position; (b) schedule the compat shim (or an internal issuer/verifier split that simulates it) before any "public consumer parity" claim; (c) define the actual enforced invariant (e.g. a `check:invariants` rule: no production code path branches on caller identity; V1-mode + legacy-env fails closed — S5 already delivers the second half) and point criterion 5 at it.

**R4 — The v1→v2 exit criteria do not gate what their heading claims: "safe to open to the public" (§1).**
The six criteria are all v1-hygiene (soak, admission, replay, pinning, no-bypass, runbook). They contain **nothing about the isolation/tenancy escalation** that §4 itself says the public step requires. Missing, at minimum:
1. **A tier decision for untrusted default-tier customers.** §4 says opening to untrusted strangers "**likely** needs the microVM tier … for customers who **demand** a dedicated boundary" — which frames microVM as a premium option, while §1's v2 row still offers "gVisor (shared, dense)" as a selectable tier. So the doc never actually decides whether an anonymous self-serve stranger lands on shared gVisor or a microVM. Modal proves shared gVisor *can* be a product — with a dedicated security org, patch cadence, and abuse pipeline behind it, none of which appear in the v2 backlog. Either state "untrusted self-serve requires the microVM tier until an explicit owner risk-acceptance for shared gVisor" or add the operational preconditions (gVisor CVE-response SLO, escape canaries running continuously — currently listed only inside SBX1.5) as public-opening criteria. "Likely" is not a decision in the highest-stakes sentence of the document.
2. **Egress/abuse controls.** The v2 backlog (§5) has auth, quotas, billing, console — and **no network egress policy, no abuse/anti-crypto-mining/outbound-scanning story** at all. v1's posture is egress-denial (probed in S4); a public product cannot be egress-deny-only (E2B has `allow_internet_access`), and the moment egress opens, abuse handling is a launch blocker, not a nice-to-have.
3. **Multi-tenant auth existing at all** — criterion 5's "only path" property plus §5's backlog implies it, but nothing in 1–6 requires per-tenant identity, DoS quotas, or tenant-isolation testing before the first stranger arrives.
The text partially saves itself ("meeting them is the gate to start the v2 productization backlog") but the section heading and criterion framing say "stable enough to open to the public". Fix: rename the criteria to what they are (v1-complete / start-v2 gate) and add an explicit second gate — "public-opening criteria" — that includes the tier decision, egress/abuse, and multi-tenant auth.

**R5 — "Near-drop-in E2B-compatible" overclaims what the coverage map shows (§9.0, [SPEC] §4).**
Honest for the minimal loop (create → run → read/write → kill). Not honest as "drop-in" for real E2B SDK usage, per the plan's own coverage map: **no streaming stdout/stderr on the wire** (the single most-used E2B ergonomic — `onStdout` callbacks), **no background commands / sendStdin / PTY**, **no `connect()`/`list()`/`getInfo`/metadata**, **no `envs` value pass-through** (create-time `envs` is E2B's standard secrets path; `credentialRefs` is better but is a client code change, and S1 **fail-closes non-empty `credentialRefs`** in v1 anyway), **no `getHost(port)`** (kills every dev-server/preview use case), **6 MiB fs transfer cap, no signed-URL bulk path**, and **`@e2b/code-interpreter` `runCode` is never mentioned** — the surface most published E2B agent examples actually use. Plus R3: the API-key shim that makes the SDK's auth model work at all is unscheduled. Fix: replace "near-drop-in" with the defensible claim — "E2B's lifecycle/exec/fs verb shape, minimal-loop compatible at v1.1 (streaming + list + shim), full SDK compat is a v2 deliverable" — and add `runCode`/code-interpreter to the coverage map explicitly, even if the answer is "never".

**R6 — Architecture demands two "must be right NOW" interfaces that no slice builds (§2 Layers 2 and 4).**
Layer 2: "The placement seam **must exist as a named interface** the control plane calls, even though v1's impl is a constant. If the control plane assumes 'one box' inline, v2 needs surgery." Layer 4: "**Align boring-bash to envd's shape now**." Grep S1–S5: no `placeSession` (or any placement interface) appears in any slice — S5 hardcodes "all 256 placement buckets to the admitted EU worker" in config — and no slice touches boring-bash/envd alignment. So the unified doc's central discipline ("interfaces product-shaped now") is asserted by the architecture and delivered by nothing. Either add the placement interface to S1/S5 scope (it's small), or downgrade both from "must be right NOW" to "v2-entry refactor" and accept the surgery the doc itself predicts. As written the two halves of the PR contradict each other on the doc's own governing rule.

## RECOMMENDED

**N1 — Dangling reference:** §5 spike step 1 cites "`remote-container-daemon-survey.md`" for the Daytona/faasd license traps; no such file exists in `references/` (content lives in `build-vs-adopt-survey.md`). Fix the filename.

**N2 — Stale spec header:** `control-plane-api-spec.md` line 9 still says "E2B-surface section (§1) **pending** the E2B research pass" while §1 opens with "Verified against official SDK reference". Delete the stale banner — it undermines the "verified" claim §9 leans on.

**N3 — Modal-density grounding is softer than §9.5 implies.** "Modal runs dense multi-tenant production on gVisor" rests on Modal's verified docs quotes plus a "Modal **markets** 100,000+ concurrent" line, with the ISO file itself flagging that Modal's engineering blogs 404'd `[UNVERIFIED]`. Fine as directional evidence; phrase it as Modal's claim, not established fact, since it carries the shared-gVisor-tier weight in R4.

**N4 — §9.4's "stronger auth" marketing should state the v1 caveat inline.** The capability/nonce model is genuinely stronger *per-request*, but in v1 both issuer and verifier derive from **one static host-root-equivalent secret** (plan threat model). A leaked capability is dead; a leaked secret is the whole box. The plan says this plainly; the architecture doc, which is the sellable artifact, never does. One sentence fixes it.

**N5 — TAKE/ADAPT/OWN/RE-HOST is honest only because the extraction spike exists — make the spike blocking.** The survey's own E2B verdict ("auth entangled with control plane and Postgres; Nomad/Consul/Terraform fleet … absurd surface for one VM") and `[E2B-INT]`'s per-component liftability notes ("core is liftable **but heavy**"; client-proxy "code cloud-welded") justify skepticism that TAKE is a clean lift — especially the UFFD/NBD/cgroups core, and the ongoing cost of tracking a live upstream fork (unmentioned anywhere). §5's spike is the right control; state explicitly that **no v2 microVM bead may be created before the spike's go/no-go**, and add "upstream-tracking/maintenance cost" to the spike's required outputs.

**N6 — Public-prefix claim is unowned.** §9.0 says the prefix "moves" to `/v1/...`; every slice, and S5's manual proof, ships `/internal/v1/...`. Harmless, but assign the rename a home (v1.1) so the architecture doc isn't describing routes that won't exist.

---

**Citation spot-check summary (task item 1):** `[E2B-INT]` citations (UFFD in orchestrator, envd analog, backend-neutral api contract) — verified accurate. `[ISO]` Fly/Modal/Vercel quotes — verified present and fairly used (modulo N3). `[SPEC]` coverage-map citations — accurate and commendably honest. `[SURVEY]` as cited for the Docker-AuthZ CVE — **misrepresents its source** (R1). One dangling filename (N1), one internal spec contradiction (N2), one architecture↔plan contradiction (R2).

The plan survives the attack on its core: build-not-adopt is decisively grounded, the layer seams are right, and §6's discipline guardrail is the best section in the document. Fix R1–R6 and this is ratifiable.

https://claude.ai/code/session_01MqBvPpW8gLJQtzXvoAxee9

hachej:
Adversarial review (Fable) — required fixes applied. Point-by-point (commit 2c94ac52c, CI green). No merge; leaving for owner gate.

**R1 (broken/circular Docker-AuthZ citation) — FIXED (grounded, not papered).** The finding genuinely did not exist in `build-vs-adopt-survey.md` (confirmed: zero "AuthZ" hits beyond "narrow authz"). Added a real **"Security lesson — the Docker-AuthZ CVE class"** section to the survey citing the primary source **CVE-2024-41110** (Docker Engine AuthZ-plugin bypass, CVSS 10.0, body-less-request forwarding; regression of the 2018 class) with advisory/commit pointers, framed as the failure mode of bolt-on-authz-in-front-of-a-privileged-socket that every "adopt" candidate shares. Reworded arch §9.4 and the §5 OWN framing to "a **design judgment** grounded in the CVE class" pointing at the now-real `[SURVEY]` section. The `[SURVEY]` citation now resolves; "no decision here is ungrounded" is true.

**R2 (arch↔plan contradiction on replay defense) — FIXED.** Replaced "boot-epoch fencing" with **transactional global nonce uniqueness across processes/connections** in arch §1 exit-criterion 3, arch §9.4, and `[SPEC]` §3.1 — each now states the boot-epoch proposal is intentionally subsumed (V1 stores no epoch column) and names the concurrent-connection "exactly one accepted, one replay" test as the fencing proof, matching plan S2.

**R3 (dogfood "same public API" false in v1) — FIXED honestly.** (a) §3 now states plainly that v1 dogfoods the **wire contract + execution path**, NOT the public auth trust position, and that **in v1 Seneca IS a privileged first-party consumer** in the capability-issuance dimension; "normal public consumer" is labeled a **v2 target, not a v1 fact**. (b) The edge compat shim is now **scheduled** as a named v2 deliverable (§5 "Multi-tenant edge auth / compat shim" bead) and a §4a public-opening precondition. (c) Defined "grep-enforceable" concretely: the enforced invariant = (1) no production path branches on caller identity (a `check:invariants` rule — flagged **not yet implemented by any slice**), (2) V1-mode + legacy env fails closed. (d) The `BORING_WORKER_BASE_URL` **config-precedence** bypass is called out as the real bypass and noted as **closed by plan S5** (fail-closed), with the point that a grep for "Seneca" would never catch it. §1 criterion 5 rewritten to point at this invariant and admit it is not yet mechanically evaluable.

**R4 (exit criteria don't gate public opening) — FIXED.** Renamed the six criteria to the **"v1-complete / start-v2 gate"** (explicitly v1-hygiene) and added a distinct **§4a "public-opening gate"** evaluated *after* it, with: (1) firm isolation-tier decision — **untrusted public self-serve requires the microVM/Firecracker tier + continuous SBX1.5 admission; shared gVisor is trusted/first-party only until an explicit written owner risk-acceptance** (resolving the §4 "likely" hedge); (2) **egress + abuse controls** (default-deny egress policy, per-tenant rate/spend caps, abuse detection) — added to the §5 backlog as hard blockers; (3) **multi-tenant auth must exist** (per-tenant identity, compat shim, DoS quotas, tenant-isolation tests). "Open to public" is now a separate gate, not implied by v1-complete.

**R5 (E2B-compat overclaim) — FIXED.** Downgraded "near-drop-in E2B-compatible" to **"E2B-SHAPED SUBSET"** in the §9 intro and §9.0, with an explicit covered-vs-gaps list: streaming stdout/stderr, background/`sendStdin`/PTY, `connect`/`list`/`getInfo`/metadata, `envs` value pass-through (vs value-free `credentialRefs`), `getHost(port)`, 6 MiB cap/no signed-URL bulk, and the unscheduled edge auth shim. Added **`@e2b/code-interpreter` `runCode`** to the §9.2 coverage map (answer: never a wire verb; client-side wrapper over `exec`). GTM restated as "E2B-familiar surface, subset coverage, migration needs adaptation."

**R6 (unbuilt "must be right NOW" interfaces) — FIXED.** §2 Layer 2 and Layer 4 cells + layer notes + §6 bullets now honestly state that **no S1–S5 slice ships a `placeSession` code interface or the boring-bash↔envd alignment**; both are marked **v2-entry** (placement extracted as a refactor; envd alignment measured by the §5 spike step 4). Clarified in **plan S5** that the 256-bucket→one-box config **is** the placement seam's single-box implementation (the degenerate one-box case, not an inline hardcode), with the named code interface deferred to v2.

**Recommended applied:** N1 (dangling `remote-container-daemon-survey.md` → `build-vs-adopt-survey.md`), N2 (removed stale "§1 pending E2B research pass" spec banner, replaced with the verified-against-SDK statement), N4 (inline v1 single-static-secret / host-root-equivalent caveat added to §9.4), N6 (`/v1/` public-prefix rename given a v1.1 home; internal `/internal/v1/...` noted as today's reality). N3/N5 (Modal-density phrasing, spike-blocking) not applied this pass — flag if you want them.

CI: green (Invariants, Lint, Typecheck, Unit Tests, Runtime Refactor P8, Action Pins, Dependency Age all pass). Not merged.

https://claude.ai/code/session_01MqBvPpW8gLJQtzXvoAxee9

hachej:
Closing plan+adversarial pass (Fable): full /plan-methodology check + adversarial attack on the unreviewed +442-line deepening (commit `bed35083d`), verification that all prior review fixes held, fixes applied and self-verified in commit `65c79f756`.

## Verdict: **PASS** (owner-gate ready)

## 1. /plan-methodology conformance — CONFORMS

Checked against `.agents/skills/plan/SKILL.md` → `docs/procedures/issue-plans.md`:

- Frontmatter (github/issue/state/updated/flag/track) ✓; Today/Delta framing per slice ✓ (and I verified the Today claims against real source, below); proof-per-slice with exact commands ✓; per-slice **Review budget** verdicts with the Decision-1 return-to-owner rule ✓; Out of Scope / owner gate / adversarial-review-before-gate lineage ✓. The methodology crosswalk table is accurate, not decorative.
- **One accepted deviation, explicitly declared:** `**Bead:** materialized at owner gate (Sx)` instead of `br-###` — correct for a docs-only PR that creates no beads; the owner-gate section says materialization is the approval act. Conformant in substance.
- **Owner-gate reminder (not a plan defect):** the skill requires the plan-approval intention to link a visual plan review doc (`visual-review-doc.md`) and `br dep cycles`/`bv --robot-insights` before Beads handoff — those are gate-time acts for whoever runs the gate, noted here so they aren't skipped.

## 2. Prior fixes HELD — all six verified in current text

| Prior required fix | Status |
| --- | --- |
| Transactional nonce uniqueness, no boot-epoch (R2) | ✓ arch §1 crit-3, §9.4, `[SPEC]` §3.1 (lines 178-180), plan S2 — all four say subsumed, no epoch column |
| §4a public-opening gate (microVM for untrusted, egress/abuse, multi-tenant auth) (R4) | ✓ §4a intact, firm tier position, §5 carries the blocker beads |
| Honest dogfood framing (R3) | ✓ §3 intact ("Seneca IS a privileged first-party consumer… v2 target, not v1 fact") — but see finding F3: §0's thesis sentence still contradicted it. Fixed. |
| E2B-shaped-subset, not drop-in (R5) | ✓ §9.0 gap list intact incl. `runCode`; deepening's S1/S5 sections consistently restate "subset", no re-inflation |
| Docker-AuthZ CVE citation real (R1) | ✓ `build-vs-adopt-survey.md` §"Security lesson" exists with CVE-2024-41110 primary-source detail; §9.4 cites it as design judgment |
| placeSession/envd honestly v2-entry (R6) | ✓ arch §2 both rows + layer notes + §6; plan S5 placement-seam section and S1 Layer-4 subsection carry the honest "measured, not built" caveats |

## 3. Adversarial attack on the +442-line deepening — findings

Spot-checked every code citation the deepening added, against real source in this worktree:

**Verified accurate (no action):** `sessionRuntime.ts` `startContainer` at line 656, call sites 308/684, `buildDockerRunArgv` at 661 — all exact. `bindingRegistry.ts` interfaces at lines 34/40, options at 53, inline `new SingleUseNonceStoreV1` ~204 — exact. `singleUseNonceStore.ts` 3-arg `consume` — exact. `protocolClient.ts` names exactly the 7 `/internal/v1/...` paths in the S1 route table. `[SPEC]` §2.1/§2.2 claims match `remoteWorkerProtocolV1.ts`/`sandbox.ts` (`onStdout`/`onStderr`/`onHeartbeat`/`signal`/`timeoutMs` all present). e2b-internals §1/§3/§5/§6/§8 quotes used by the S1-S5 "Grounded in E2B" notes are all faithfully quoted, including the appendix-confirmed `SandboxService {Create, Update, List, Delete, Pause, Checkpoint}`. `packages/boring-bash` exists. The SQLite DDL, S1 route→RPC table, and pin-gate description are mutually consistent. The "what v1 does NOT create" list is consistent with S1-S5 (no client-proxy/Redis/ClickHouse/Nomad/auth-service/template-manager appears in any slice).

**F1 — REQUIRED, fixed: the tier→module map overclaimed `SandboxProviderV1`.** The provider-seam row said the contract "deliberately mirrors that RPC set" and "the daemon's routes proxy to it." Neither is true: the real `providerV1.ts` surface is `create`/`invalidate?`/`close?` plus the pair's `Sandbox`/`Workspace`/`dispose()` — no `Update`/`List`/`Pause`/`Checkpoint` methods — and e2b-internals §8's own wording is "**should** mirror exactly that RPC set" (a recommendation, not a fact). Worse, the seam sits on the **consumer side** of the wire (Seneca composes the `remote-worker` provider through it); the daemon's routes proxy in-process to `RunscSessionRuntimeV1`, exactly as S1's own module map says — so the row contradicted the plan's S1 section. Rewrote the row honestly: consumer-side frozen contract; does NOT currently mirror the six-RPC set; the mirroring is a v2-entry alignment when the microVM provider lands.

**F2 — fixed: "six V1 routes" vs the 7-row table.** Three spots (tier-map row 1, `router.ts` bullet, S1 Grounded bullet) said six; the table and `protocolClient.ts` have seven (health + six capability-guarded). Corrected to "seven (health plus six capability-guarded)".

**F3 — fixed: §0 residual contradiction of the R3 dogfood fix.** Architecture §0 still asserted "Seneca is a normal consumer of the public surface, not a privileged internal integration" — flatly contradicting the R3-fixed §3. Marked as target framing with a pointer to §3's honest v1 status.

**F4 — fixed: S2 Delta vs concrete-transaction ordering.** Delta said budgets-then-insert; the concrete section says insert-then-budgets. Functionally equivalent under the held `BEGIN IMMEDIATE` lock, but the two descriptions of the same transaction now match (checks after insert, abort on failure).

**Not findings (attacked and survived):** the DDL + `BEGIN IMMEDIATE` fencing claim is sound (write lock at statement start, PRIMARY KEY collision = replay); the S4 transcript uses exactly the S3b CLI/env contract names; `build-qualification-bundle.mjs <cohort-spec.json>` positional handoff matches the real script; the E2B ops table's manual-vs-continuous split matches Decision 6 and the non-goals; the v1.1 `/v1/` prefix ownership (N6 fix) is consistently restated in the deepened S1.

## 4. Self-verification

Re-read all five edits in context; grepped both docs for residual "deliberately mirrors" / "six … routes" / "normal consumer" claims — none remain except the quoted-as-recommendation form. The rewritten provider-seam row now agrees with S1's module map and with arch §2 Layer 3 (which never claimed RPC mirroring). CI green on the fix commit.

**Verdict: PASS.** Nothing deliberately left open except the two gate-time acts named in §1 (visual review doc + bead-graph hygiene commands at materialization). Not merged — owner gate.

https://claude.ai/code/session_01MqBvPpW8gLJQtzXvoAxee9

hachej:
## Restructure + rewrite: split the monolith into three single-voice docs

`docs/issues/1081/plan-sbx14.md` had accreted into a hard-to-review monolith
edited by several agents in sequence — terminology drift, redundancy, stitching
seams, and one doc trying to be three at once. It is now split into focused,
single-voice documents. **Restructure + rewrite-for-coherence only — no substance
change.** All checks green (Invariants, Lint, Typecheck, Unit Tests, Runtime
Refactor P8).

### New layout (`docs/issues/1081/`)

| Doc | Role | What it holds |
| --- | --- | --- |
| `direction/sandbox-service-architecture.md` | **WHY + WHAT** | Four-layer model, dogfood-to-product staging, isolation ladder, sovereignty, harvest-E2B. Trimmed: execution/API detail moved out, pointers left behind. |
| `plan-sbx14.md` | **HOW + WHEN** | S1–S5 slices (Today/Delta, files-touched, proof, Bead + Review-budget, Blocked-by), v1 module layout, provisioning/ops (S4), prod flip (S5), **the v1-complete exit criteria + the public-opening gate** (moved in from the architecture doc), review protocol. |
| `api-spec.md` **(new)** | **THE CONTRACT** | The Layer-1 control-plane API surface: v1 endpoints, the capability + single-use-nonce auth handshake, the E2B-shaped-subset coverage map (have/partial/new), the `SandboxProviderV1` mapping, E2B-compatibility posture. Extracts the architecture doc's §9 + the plan's HTTP-surface/auth grounding, consolidated with `references/control-plane-api-spec.md`. |
| `references/` | **EVIDENCE** | Unchanged raw research. |

`README.md` reordered as the index: architecture (why) → plan (how/when) →
api-spec (contract) → references (evidence).

### What moved where
- **§9 "v1 control-plane API shape"** (was in the architecture doc) → **`api-spec.md`**; a pointer stub remains at §9.
- **Plan's "HTTP surface — minimal v1 endpoint list" table** + the capability-vs-E2B-API-key **auth grounding** → **`api-spec.md`** §2/§3.4; the plan now references it and no longer restates schemas.
- **v1-complete exit criteria** (was architecture §1 gate) + **public-opening gate** (was architecture §4a) → **`plan-sbx14.md`**; the architecture doc keeps the firm isolation-tier *position* (§4a resolution) and points to the plan for the operational checklist.
- All in-plan cross-refs retargeted from `architecture §9.x` / `control-plane-api-spec §x` → `api-spec §x`.

### KVM terminology fix (consistency, no substance change)
The confusing "KVM VM" label for the v1 host is gone. **v1 host = "rented Linux
VM"** (provider-virtualized, **no `/dev/kvm`** — runsc/systrap is user-space).
KVM / nested-virt / bare-metal / microVM are now reserved for the **v2**
Firecracker sections, with the v1-vs-v2 boundary stated crisply where the two
meet (topology, Gate 0, S4, non-goals). Also applied one-term-per-concept passes
(daemon/control-plane, capability/nonce, slice/module names) and removed
redundant restatements + reconciled stitching seams.

### Adversarial-review fixes + verified facts — all confirmed surviving
- Transactional nonce uniqueness across processes/connections; **no boot-epoch column** (S2 / #1167).
- The public-opening gate (now in the plan) and the honest dogfood framing (**v1 Seneca is a privileged first-party consumer** in the capability-issuance dimension).
- **E2B-shaped subset, not drop-in**; the real **CVE-2024-41110** citation (Docker-AuthZ, CVSS 10.0).
- Verified exact source line numbers, unchanged: `startContainer` **656/661/308/684**, `bindingRegistry.ts` **34/40/53/204**, `singleUseNonceStore` 3-arg `consume` today, `protocolClient` route set.
- `placeSession(request)→box` and the boring-bash↔envd alignment kept explicitly marked **v2-entry** ("measured, not built" in v1).
- Every slice keeps its **Bead** + **Review-budget** fields.

### No substantive contradiction found
The split surfaced no genuine substantive contradiction to resolve. One honest
tension already documented in the source (and preserved verbatim, not resolved):
**v1-complete criterion 5's part (a)** — the `check:invariants` "no branch on
caller identity" rule — **is not yet implemented by any slice**, so criterion 5 is
not mechanically evaluable yet. This was already flagged as owed work in the
architecture doc (§3) and remains flagged in both the plan and api-spec; it is a
known gap, not a contradiction introduced by the restructure. Flagging it here per
instruction so the owner sees it at the gate.

**Not merged** — owner gate.

hachej:
## gVisor platform-security validation folded in

Added the new `gvisor-platform-security.md` research report and indexed it, then folded its verdict into:

- `plan-sbx14.md` at the v1 topology / `--platform=systrap` choice, with the report plus the official gVisor platform and security-model citations.
- `sandbox-service-architecture.md` at the gVisor isolation ladder, clarifying that systrap and the gVisor KVM platform share the same Sentry security boundary; the difference is performance, not hardening.

Result: v1 on a rented Linux VM with systrap and no `/dev/kvm` is explicitly validated as **not a security compromise**. There is no security reason to provision a KVM VM; nested virtualization would add no gain and can be slower.

Proof: `git diff --check` passed. CI on `3202d1ac5` is green: Invariants, Lint, Typecheck, Unit Tests, Runtime Refactor P8, Action Pins, Dependency Age, and summaries all passed (non-applicable jobs skipped).

Not merged.

hachej:
Added a standalone tech-choice **decision record**: `docs/issues/1081/tech-choice.md` (indexed as item 4 in the issue README, above the references).

It's the "why we chose what we chose" narrative — readable end-to-end by an engineer or investor. Each of the ten decisions is structured as **Question → Options → Evidence (verbatim primary-source quotes + citations) → Decision → Reasoning**, synthesizing the checked-in `references/`:

1. **Build vs adopt → BUILD.** R1–R7 survey (nothing clears R1–R3+R5–R6 with a clean R4 seam) + the Docker-AuthZ CVE class (CVE-2024-41110, CVSS 10.0; the `Content-Length: 0` AuthZ-bypass) making "filter the Docker API" unsound.
2. **Isolation → gVisor v1, microVM/Firecracker v2.** Modal ("stronger isolation than most other container runtimes"), Fly ("unlikely… in Go… likely in the C-language Linux kernel"; "only worried about the integrity of the kvm subsystem"), Vercel, Cloudflare quotes; isolation ladder; our economics differ from Modal's density driver.
3. **gVisor platform → systrap (not KVM).** gVisor docs: platform = performance, the Sentry is the boundary; KVM buys no hardening.
4. **Managed k8s vs daemon → daemon on a rented Linux VM.** gVisor-on-managed-nodes undocumented on all 3 CH/EU providers; provider owns the node image.
5. **Orchestration → daemon+placement → Nomad+Firecracker → k8s+KubeVirt.** E2B's Nomad+Firecracker in prod is the validation.
6. **Harvest E2B (v2) → yes, TAKE/ADAPT/OWN/RE-HOST, re-hosted sovereign** (Apache-2.0: job specs, envd, UFFD).
7. **Hosting → CH/EU sovereign, own bare metal / rented CH VM.** CLOUD Act reasoning; Vercel iad1-only; E2B a US company.
8. **Daemon language → Node/TS in `packages/boring-sandbox`**, reusing `SandboxProviderV1` + merged V1 schemas.
9. **Transport → Tailscale-only ingress + capability+nonce** (not reusable API keys — the "own your security edge" CVE lesson).
10. **Product path → dogfood Seneca → open to public, behind the v1-complete + public-opening gates.**

All quotes preserved verbatim with source URLs. CI: **11 pass, 8 skipping, 0 failures** (Lint, Typecheck, Invariants, Unit Tests all green). Docs-only change. Not merging — leaving for the owner gate.

hachej:
Cloudflare currency fix (owner flag on the stale 2018 source).

**What I verified:** The isolation-comparison Cloudflare entry cited only Zack Bloom's Nov 2018 *Cloud Computing without Containers* post, which describes Cloudflare's original **V8-isolates-only** model. That is no longer the whole story — Cloudflare has since shipped **Cloudflare Containers** (announced Developer Week 2025, open beta late June 2025): a Linux container compute model that runs code in any language/runtime alongside Workers, each instance paired with a Durable Object sidecar for lifecycle/state.
- Announcement: https://blog.cloudflare.com/cloudflare-containers-coming-2025/
- Docs: https://developers.cloudflare.com/containers/

**What I changed (both files):**
- `docs/issues/1081/tech-choice.md` and `docs/issues/1081/references/isolation-choices-primary-sources.md`
- Kept the 2018 quote, now explicitly labeled the **isolates-era / isolates-primitive** data point (the boundary we are *not* taking — language-restricted JS/Wasm, software-only, resting on V8).
- Added a **currency note** in each: Cloudflare is a two-era story, so 'isolates only' is inaccurate; the addition does not mischaracterize Cloudflare's current overall capabilities.
- Preserved everything else, including the **isolates < gVisor < Firecracker** ladder — that conclusion is about isolates-as-a-primitive and remains valid.

Confirmed via WebFetch on both Cloudflare URLs above, so no [UNVERIFIED] tag was needed.

CI: green (docs-only; all substantive jobs pass, code jobs skipped). Commit `ded8c0aa2`. **Not merging** per instruction.

hachej:
Thermo simplification review (Fable): SBX1.4 plan — ruthless subtraction pass

**Verdict: OVER-BUILT for its own stated v1.** The stated goal is "Seneca (one tenant) off Vercel onto our own gVisor box, fast." The plan's actual v1 is a multi-tenant-shaped control plane wearing a single-tenant costume. The threat model the plan itself states — one tenant, Tailscale-only ingress, secret is **host-root-equivalent**, the only capability issuer is the client itself — makes several whole slices security theater *for v1*: they defend against attacker classes that cannot exist until the public-opening gate. Estimated cut: **1 full slice (S2), ~30-40% of S1's delta, and ~40% of the plan document**, with zero weakening of the fail-closed posture. Honest v1 = 4-5 slices, S1 comfortably inside budget.

The key structural observation: **Seneca holds the static secret and mints its own capabilities client-side** (architecture §3 says this plainly). Therefore in v1 the entire capability/nonce/receipt edifice is cryptographically equivalent to `Authorization: Bearer <secret>` — anyone who can replay a capability either captured it on a WireGuard-encrypted tailnet or loopback (i.e., already owns the box or the tailnet) or holds the secret (game over per the plan's own words). Keeping the *existing* SBX1.3 registry in-memory is fine (it's already merged, composition is cheap). Building **new durable, transactional, per-tenant-budgeted machinery** around it in v1 is defending a boundary that doesn't exist yet.

## CUT LIST (ranked by LOC/complexity saved)

**1. CUT S2 entirely (persistent SQLite nonce store) — defer to public-opening gate.**
Largest single win: an entire slice (1-3 days), `persistentNonceStore.ts`, WAL/synchronous/busy-timeout pragmas, node:sqlite floor pinning, local-fs-type assertions in S4 `--check`, restart/concurrency regression suites, the "never roll back to volatile" rollback doctrine. What it defends: replay of a *captured* capability across a daemon restart. Who can capture one: an attacker on the tailnet or loopback — who per the plan's own threat model is already host-root-equivalent. Single tenant + client-is-issuer means nonce durability protects nothing real in v1. The existing in-memory `SingleUseNonceStoreV1` (already merged) stays as-is. Risk of cutting: a replay window across restart *if* the tailnet is compromised without the secret being compromised — a contradiction in the plan's own model. #1167's atomicity constraint becomes moot (nothing persists → nonces and bindings trivially share one lifecycle, which the plan admits is today's accidental-but-correct state). Re-open S2 verbatim when the edge compat shim / first non-first-party client arrives. **CUT.**

**2. CUT the per-workspace nonce sub-budget (S2 delta, 4-arg port change).**
The plan's own tell: acceptance tests are literally "tenant A can exhaust only its sub-budget; tenant B remains accepted." There is no tenant B. This is multi-tenant fairness engineering in a one-tenant system, and the plan even confesses it ("#1167 finding 1 previously treated it as deferrable"). Falls out automatically with cut #1. **CUT.**

**3. CUT the dual-secret rotation machinery (S1 "Secret rotation" section).**
Primary+secondary credential files, same-value startup check, bounded-overlap expiry, 3-step choreography, cross-slice rollback references. For one operator, one client, one box: rotation = update two config values, restart both ends, ~seconds of canary downtime, with `vercel-sandbox` as the always-available fallback. On suspected compromise the plan already says stop admission first — at which point overlap buys nothing. Risk: a brief availability gap during rotation of an internal canary. Zero-downtime rotation is a real requirement the day a second, independently-administered client exists — not before. **CUT** (~100-200 LOC + tests + three procedure references).

**4. CUT `authRateLimiter.ts` (failed-auth token buckets, per-source backoff, forwarded-header trust logic).**
Who brute-forces the bearer? The tailnet ACL admits exactly one node: Seneca, which *holds the secret*. The limiter defends a public edge the daemon deliberately does not have. Constant-time comparison (keep, it's one function call) already defeats timing probes. The forwarded-source-header parsing/rejection sub-mechanism exists solely to feed this limiter. Risk: none in v1's topology; the connection/session caps (keep, see below) already bound resource abuse from a buggy client. Restore at the public edge, where it belongs (edge shim, api-spec §3.4). **CUT** (~150-250 LOC + slowloris/saturation test surface shrinks).

**5. SIMPLIFY S3b: keep "hostile probes green on the exact box + exact pinned image," defer the full V3 observe→bundle→bound→verify 4-phase evidence formalism.**
The load-bearing security property is: *the exact box passed the committed hostile probe suite (11/11), openat2/prjquota/helper are real, and the daemon only ever starts the one pinned digest* (S3a). That is deliverable with the existing integration harness pointed at the external pinned image + the real quota helper (two narrow opt-ins) and a manual admission record (transcript + digests). The immutable-bundle digest-binding, deterministic cohort-spec, cross-layer dual-field-name digest equality tests, and strict verifier choreography are **SBX1.5 fleet-admission machinery pulled forward** — they exist so *unattended automation* can trust evidence. v1's admission is one human on one box reading one transcript. Risk: weaker mechanical drift-detection between qualification and config (mitigated: S3a's startup pin + `--check` still fail closed on digest mismatch); honest owner call, but the default should be defer. Saves most of an M slice + the 4-phase S4 transcript ceremony.

**6. CUT the Caddy/Tailscale-cert TLS layer; serve the daemon directly on `tailscale0`.** *(Owner call — flagged honestly.)*
Tailscale is WireGuard: every packet on the overlay is already encrypted and node-authenticated. HTTPS-over-Tailscale is a third crypto layer whose cost is real: Caddy install/config, `tailscale cert` renewal, the MagicDNS+HTTPS-certificates procurement constraint (Gate 0!), loopback-proxy topology, CA/server-name pinning in the fleet config, and four extra `--check` assertions. "Plaintext across the tailnet is forbidden" sounds like a security invariant but the tailnet is not plaintext on any wire. Risk of cutting: sole reliance on tailnet integrity for transport (a Tailscale-coordination-plane compromise loses encryption *and* the ACL together — though it loses the ACL either way). If defense-in-depth wins, keep it — but then say it's defense-in-depth, not a requirement. Saves an ops component + a procurement gate.

**7. CUT the 256-bucket placement config; config = one worker URL + digests.**
256 buckets → one box is not "the degenerate case of a scheduler," it's a data structure with exactly one distinct value, plus bucket-assignment validation code nobody exercises. The architecture doc already concedes no `placeSession` interface ships. The v2-entry refactor that introduces buckets can introduce buckets. Risk: none; the seam is the *config file's existence*, not its cardinality. **CUT.**

**8. TRIM the 7-day qualification expiry / 6-day manual requal treadmill → requalify on change.**
Re-running the full S4 admission weekly forever, on a box whose kernel/runsc/image change only when the operator changes them, is calendar ceremony. Requalify on *change* (kernel, Docker, runsc, image, helper — the plan already lists these triggers) plus `--check` at every daemon start. Risk: undetected in-place drift on an unattended box; if the owner wants a freshness bound, 30 days matches the actual risk better than 7. Owner call, default = on-change.

**9. CUT ~40% of the plan document itself.**
The four "Grounded in E2B" sections, the tier→module crosswalk table, the "Layer-4 alignment target" subsection, and the ops-concept mapping table (~350-400 lines) restate `references/e2b-internals-architecture.md` and the architecture doc inside an *execution plan*. The methodology/template crosswalk table is pure process porn (AGENTS.md rule 8: artifacts only when they hard-gate a capability — a table proving the plan followed the plan-template gates nothing). The public-opening gate appears in full in both the plan and architecture §4a — keep one, link the other. None of this changes what gets built; it changes what six implementation agents must re-read every session.

## WHAT MUST STAY (security-load-bearing, do not touch)

- **Fail-closed startup on missing/mismatched qualification facts** and the **S3a digest pin at `startContainer()`** (both start and replacement paths). Small, real, the heart of the admission posture.
- **Tailscale-only ingress, no public listener, provider firewall deny, ACL = Seneca only.** This *is* the v1 security boundary at the network layer.
- **Egress-deny probes + the 11-probe hostile suite green on the exact production box** (Gate 0 + the simplified S4). Non-negotiable.
- **No host paths / Docker socket / image override on the wire; daemon owns roots** (Decision 2). Load-bearing regardless of tenancy.
- **Root quota helper root-owned/non-writable; prjquota real, not stubbed.**
- **Body bounds, strict schemas, connection/active-session caps, startup sweep, SIGTERM drain.** Cheap, protects against Seneca's own buggy/runaway agents — v1's *actual* threat.
- **S5's legacy `BORING_WORKER_BASE_URL` precedence fail-closed fix.** The one real bypass in the codebase today; genuinely the most important fix in the plan.
- **`BORING_AGENT_MODE` env rollback + drain-before-flip.** The safety net that makes a fast canary honest.
- **Gate 0 (openat2 proof before funding S1).** Correctly placed: cheapest possible falsifier first.
- **Constant-time secret comparison, no secret logging, deterministic idempotent sandboxId.** Each ~a few lines; keep.

## On the LOC question (item 3 of the brief)

The plan's own hedge — "S1 MUST confirm ≤1,500 LOC or return for an S1a/S1b split" — is the red flag it appears to be. Router+codec+transport+watcher+bin is a plausible ~700-900, but the rotation dual-secret logic, the rate limiter, the forwarded-header trust code, and their mandated adversarial test surface (slowloris, saturation, rotation-overlap states) are what push it toward the split. Cuts #3 and #4 remove exactly the inflation and make the "600-750 LOC daemon" claim honest again.

## Net proposal

v1 = **Gate 0 → S1-lite (daemon, existing in-memory registry, static bearer + codec, caps, no limiter, no dual-secret) → S3a (digest pin) → S4-lite (provision + probes-green + pinned image + manual record) → S5 (flip, precedence fix, drain/rollback)**. One slice fewer, two S1 subsystems fewer, one ops layer fewer (owner call), and every fail-closed property intact. Everything cut has a named re-entry trigger: the edge compat shim / first non-first-party client (cuts 1-4) or SBX1.5 fleet automation (cut 5).

https://claude.ai/code/session_01MqBvPpW8gLJQtzXvoAxee9

hachej:
## LEAN V1 applied — plan split into v1 + triggered v2 (thermo simplification review)

Applied the thermo simplification review (Fable) to the sandbox plan. Commit `8d0fd9eb0` on `docs/1081-sandbox-service-architecture`. Docs-only. **CI: 11/11 green.** Not merged — owner gate.

### Lean v1 slice sequence
`Gate 0 → S1-lite → S3a → S4-lite → S5` — one slice fewer than r1, two S1 subsystems fewer, every fail-closed property intact.

- **Gate 0** — prove openat2 before funding (untouched).
- **S1-lite** — daemon + V1 transport + one token codec, using the **already-merged in-memory SBX1.3 nonce registry** (bindings/nonces volatile), **no** rate limiter, **no** dual-secret rotation. boring-bash needs **no new adapter** (the existing remote-worker provider is the adapter; the daemon matches its protocol). Honest LOC **~600–750**.
- **S3a** — `startContainer` digest pin + fail-closed qualification load (untouched, verified line numbers preserved: `startContainer()` L656, callers L308/L684, `buildDockerRunArgv` L661).
- **S4-lite** — provision + **11-probe hostile suite green on the exact box + exact pinned image** + manual admission record (transcript + digests). The 4-phase observe→build→bound→verify formalism is deferred.
- **S5** — seneca flip carrying **the single most important fix: the legacy `BORING_WORKER_BASE_URL` precedence fail-closed fix** (the one real bypass today). Placement = **single worker URL** (interface stub), requalify-**on-change**.

### Moved to `plan-v2-hardening.md` (each with a re-entry trigger)
| Deferred item | Re-entry trigger |
| --- | --- |
| Persistent nonce store + #1167 atomicity | any durable binding/record persistence **OR** pre-multi-tenant |
| Per-workspace nonce sub-budget | 2nd tenant |
| Dual-secret zero-downtime rotation | multi-operator / compliance policy |
| Auth rate limiter | any non-Tailscale / public ingress |
| S3b evidence formalism + fleet-admission automation | public-opening gate / SBX1.5 |
| Multi-box placement scheduler (256-bucket map) | box #2 |
| Requalification automation (7-day treadmill) | fleet scale |

Because v1 persists nothing, **#1167's persist-nonces-atomically-with-bindings constraint is N/A for v1** (nothing persisted = nothing to make atomic). v1 rotation = coordinated restart with Vercel fallback. Cross-linked v1↔v2 per item.

### ⚑ Owner decision left open — Caddy/TLS over Tailscale
**Not cut** — flagged as an owner call. Tradeoff: **Tailscale/WireGuard encryption alone** (tailnet is already encrypted + node-authenticated on every wire; drops Caddy, `tailscale cert` renewal, MagicDNS+HTTPS procurement gate, loopback-proxy, 4 `--check` assertions) **vs. +Caddy+cert** (defense-in-depth; a coordination-plane compromise loses only encryption, not the ACL). S4 provisioning + S5 config support whichever the owner picks.

### Untouched / preserved
S3a digest pin, fail-closed startup on qualification facts, Tailscale-only ingress (no public listener), egress-deny + 11-probe suite on the real box, no-host-paths-on-wire, root-owned quota helper, body/connection/session bounds, S5 precedence fix, env rollback + drain, Gate 0, the public-opening gate, and all verified line numbers / adversarial L1+L2 fixes that survive the cut.

### README index reordered
architecture → tech-choice → **plan-sbx14 [v1]** → **plan-v2-hardening** → api-spec → references.

hachej:
Corrected isolation-model rewrite pushed in bd6fddf90.

Proof of work:
- docs-only diff; worktree clean after commit
- pnpm lint:invariants passed
- git diff --check passed
- local Markdown link validation passed for all 16 touched documents
- independent standards review: APPROVE
- independent security/spec review: APPROVE
- independent maintainability review: APPROVE

Key decision: v1 adopts E2B self-hosted plus Firecracker, including client-proxy and required Redis routing, on one admitted sandbox host with separate trusted management infrastructure. Firecracker has no virtio-fs, so the remote workspace is a durable tenant volume/snapshot served through the API, never a host directory mount.

PR remains owner-gated. Do not merge without explicit owner approval.

hachej:
CI follow-up: all required checks completed successfully; merge state is CLEAN. PR remains open and owner-gated, with no merge performed.

hachej:
## Owner-decision reconciliation  commit 44303a42e

### Summary

Replaced the stale E2B-for-v1 and opaque-volume conclusions with the finalized v1: Kata as the adopted containerd runtime wrapper, Firecracker as the microVM/KVM engine and tenant boundary, and tenant-readable S3 plus an in-guest local-POSIX sync-hybrid storage model. The isolation spine remains unchanged. No merge performed.

### Exact sections changed

- docs/issues/1081/plan-sbx14.md
  - Correction and outcome / Non-negotiable isolation model
  - v1 technology choice: Firecracker engine, Kata runtime wrapper
  - Storage product: S3 system of record + local POSIX working disk
  - File transfer principle: copy in, never host-mount
  - Vehicle evaluation and pick / Evaluated but not selected
  - Operator invariants
  - Gate 0, S1-lite, provider-behavior reference spike, S3a, S4-lite, S5
  - Guardrails, launch criteria, v1-to-v2 boundary, and non-goals
- docs/issues/1081/tech-choice.md
  - Decision 2: Kata runtime wrapper + Firecracker engine
  - Decision 3: S3 data substrate + sync-hybrid POSIX disk
  - Decision 4: E2B deferred to v2; microsandbox/libkrun rejected
  - Decisions 5-8: qualification, operator ownership, provider behaviors/hosting, explicit v2 alternatives
  - Decisions-at-a-glance table and sources
- docs/issues/1081/README.md
  - Read-order descriptions and decision summary
- Companion-current docs reconciled after independent review found they still contradicted the new plan:
  - docs/direction/sandbox-service-architecture.md: v1, four layers, storage/file transfer, launch/operator invariants, lean scope
  - docs/issues/1081/api-spec.md: Kata/Firecracker adapter boundary, lifecycle, S3 workspace contract, credentials, provider proof
  - docs/issues/1081/plan-v2-hardening.md: Kata/Firecracker v1 baseline and explicit E2B/Firecracker-adopt vs owned/Cloud-Hypervisor branches
  - docs/issues/1081/references/README.md: Gate 0 and E2B research status

### Proof

- `pnpm lint:invariants`  PASS
- `git diff --check`  PASS
- local Markdown link validation for all seven changed docs  PASS
- independent standards re-review  APPROVE
- independent owner-spec re-review  APPROVE
- adversarial review findings applied: companion consistency, local-disk host-data wording, atomic version manifests, writer lease/direct-edit conflicts, failed-flush fencing, unsynced /inputs, action-scoped credentials, Mountpoint limitation, fail-closed optional runtime, and v2 branch clarity
- GitHub CI at 44303a42e  11 required checks PASS, 0 failures; merge state CLEAN

### Risk / rollback

Docs only. Revert 44303a42e if the owner changes the decision. The branch and PR remain owner-gated.

### Next

Owner review. Do not merge without explicit owner approval.

hachej:
## Owner handoff — final sandbox reconciliation

- [x] Reconciled plan-sbx14.md to the final Blaxel bridge -> owned Kata/Firecracker + SeaweedFS -> measured cutover architecture.
- [x] Rewrote tech-choice.md as the final decision record.
- [x] Updated the issue README index and marked contradictory earlier docs as historical where they conflict.
- [x] Added the managed-provider and cost references plus their Daytona link dependency.
- [x] Independent two-pass owner-spec review is clean on SHA 8830cf104.
- [x] Local proof: pnpm lint:invariants, git diff --check, and local Markdown link validation pass.
- [x] GitHub CI is green; PR merge state is clean.
- [ ] Owner review/approval.

The PR remains open and unmerged.

hachej:
Proof of work

What changed:
- Added `docs/issues/1081/plan-own-sandbox.md`, the phased build plan for the owned Kata + Firecracker + SeaweedFS target.
- Added dependency ordering, M0/M1/M2 milestones, phase effort/acceptance, security/operability gates, and the M0 single-box definition of done.

Automated verification:
- `pnpm lint:invariants` ✅
- `git diff --cached --check` ✅

Manual validation:
- Confirmed all relative document links exist.
- Checked the requested Gate 0 and Phases 1–6 requirements against the final `plan-sbx14.md` and `tech-choice.md`.

Waiver / known gaps:
- The live `spike/own-sandbox` branch has not committed results yet; its four Gate 0 evidence items are explicitly pending in the plan.

Reviewed head: `1715230b96a95224200d00416d01548bdc791c93`. No merge performed.

hachej:
Scope update per owner ruling (2026-08-12): **v1 ships without S3 versioning** — storage is plain durable S3 + dual POSIX access, exactly what the spike validated. Versioning is now a deferred feature gated on a compliance or undo-agent-changes requirement. Gate 0 no longer requires versioning evidence; the SeaweedFS spike item narrows to cross-prefix denial + event notification. Tradeoff on record: destructive agent overwrites are unrecoverable in v1.

hachej:
Owner directive applied in eabd36e37.

Reframe
- The sovereign fleet is now the subject and target architecture: operator-controlled EU bare metal, Kata launching pinned Firecracker micro-VMs, SeaweedFS dual S3 + POSIX `/workspace`, ephemeral `/scratch`, and `SandboxProviderV1` as the stable seam.
- Blaxel is demoted to a short interim production section. It serves current EU workloads until sovereign M0 is green, then remains only as the qualified rollback through controlled cutover/soak.
- M0 is stated honestly as one qualified box. Multi-host lifecycle, warm pools, snapshot/fork, locality scheduling, and automated qualification live in the scale plan.

Rename map
- `plan-sbx14.md` -> `sandbox-sovereign-design.md`
- `plan-own-sandbox.md` -> `sandbox-sovereign-build.md`
- `plan-v2-hardening.md` -> `sandbox-sovereign-scale.md`

Single ownership
- All concrete Gate 0 evidence, including the preserved Blaxel API-fit matrix, lives only in the build plan.
- Cutover/rollback policy lives in design; executable qualification and cutover steps live in build.
- Scale work and stable control-plane references are linked instead of restated.

Consistency
- Direction, README, API spec, technology record, and every inbound reference were updated.
- The 2026-08-12 no-versioning ruling is explicit throughout: v1 has plain durable S3 with dual POSIX access, and destructive overwrites have no object-history recovery.
- Verification: no stale old filenames repo-wide; all local Markdown targets resolve; renamed frontmatter parses with `updated: 2026-08-13`; `git diff --check` passes; independent second-pass review is clean.

No merge performed.

hachej:
Storage benchmark review amendment — head `861f4c533d25738114951ce3c7b7342e57879234`

- Corrected the `/scratch` rationale: SeaweedFS `weed mount` is a cached POSIX client, not an s3fs-style shim; git, pnpm, `node_modules`, and JavaScript build performance remain unmeasured. The retained evidence is the SQLite/fsync penalty, opt-in cross-mount locking, unestablished database-grade power-loss behavior, and SeaweedFS temporary-write guidance. The full 439-line research review is now committed.
- Added the Gate 0 storage benchmark gate on the real topology: compare mount versus local device for representative `git clone`/`git status`, pnpm install producing `node_modules`, JavaScript build, and SQLite transaction throughput. The threshold is explicitly owner-set at Gate 0; if the mount meets it, general work may run on `/workspace` and `/scratch` narrows to database files and other fsync-heavy cases.
- Made publication safety a v1 control-plane/guest-daemon requirement. Normal teardown cannot silently destroy unpublished admitted outputs; the block-versus-surface mechanism remains an explicit owner question. This is required because v1 has no versioning and an unpublished sole copy would otherwise be unrecoverable.

Also surfaced all research unknowns in the sovereign design open-questions section. Fresh-eyes review found one M0 checklist omission; fixed in this head, and the second pass is clean.

Proof: `git diff --check` passed; all three sovereign frontmatter blocks parse with `updated: 2026-08-13`; `pnpm lint:invariants` passed; PR checks are green (11 passed, 0 failed). No merge performed.

hachej:

Do not merge as-is, but do not close it either — the decision content is real, this is not vapor. Reconcile and shrink. 5,044 additions across 21 doc files, CI green, `CLEAN`, zero path conflicts; base ~50 commits behind main. Four independent problems, any one of which blocks the gate.

### 1. The gate is on a decision that already shipped nine days ago

`tech-choice.md:3` stamps "FINAL OWNER DECISION. Date: 2026-08-12" and `:91` makes Blaxel the primary bridge *pending* a Gate 0 spike verifying it satisfies `SandboxProviderV1` (`sandbox-sovereign-build.md:93-99`, still an unchecked box). **The Blaxel adapter merged to main the same day.** Verified on `origin/main`: `f33d7b077` "feat(sandbox): Blaxel provider adapter (EU-sovereign bridge) — gh-1081 (#1236)", 2026-08-12, followed by `7bcd94826` (#1249) wiring it through full-app; `packages/boring-sandbox/src/providers/blaxel/createBlaxelSandboxProvider.ts` and `packages/agent/src/server/runtime/modes/blaxel.ts` both exist; `blaxel` is a registered runtime mode (`resolveMode.ts:6-8`). Yet `sandbox-sovereign-build.md:394` still reads "Do not disturb runsc or **the planned** Blaxel provider." The doc is gating a choice that is already code.

Corroborating: none of the 83 sandbox-matching beads in `.beads/issues.jsonl` reference #1220 or its gate; the SBX1 subtree (`wt-391-forward-6gd.*`) is parked under a `deferred` parent, last touched July 2026; issue #1081 has one comment and hasn't moved since 2026-08-05. **Nothing is blocked on this gate** — which is good news: there is no cost to reworking it first.

### 2. Two checkable claims about the current runtime are false

- **`BORING_AGENT_MODE=remote-worker`** is the YAML `flag:` in the frontmatter of *both* `sandbox-sovereign-build.md:6` and `sandbox-sovereign-design.md:7`, and `build.md:397` asserts it "remains the public control-plane path." That value **throws** today: `resolveMode.ts:18-22` raises `Invalid BORING_AGENT_MODE "remote-worker". Expected direct, local, blaxel, or vercel-sandbox.` (A `modes/remote-worker.ts` file exists, but it's not in the builtin enum, so the documented env value is unusable.)
- **"No silent fallback to local, direct, runsc, gVisor"** (`tech-choice.md:209`) reads as a current invariant. The effective default is the opposite: `packages/cli/src/server/cli.ts:638-656` defaults unset `BORING_MODE` to `direct` (no sandbox) on every platform, with an explicit comment overriding bwrap autodetect. Restate as a *target* invariant.
- Scale check: SeaweedFS is Decision 4 (the whole data plane) and `git grep -i seaweed` over `packages/` on `origin/main` returns zero hits; same for `kata` and `firecracker` (the only `firecracker` hit is an unrelated comment in `packages/agent/src/shared/sandbox.ts:30`). Fine for a plan — but it means Decisions 2 and 4 are 100% greenfield while the prose reads as if reconciling with something that exists.

**Credit:** `api-spec.md:24-38` is the one section with honest Today/Delta framing ("Current source already provides… does **not** provide…") and it checks out — `SandboxProviderV1` genuinely exists at `packages/boring-sandbox/src/shared/providerV1.ts:97-109`.

### 3. The headline choice contradicts the evidence file it names as controlling

`tech-choice.md:10-12` names `references/managed-sandbox-providers-comparison.md` as controlling evidence. That file's own verdict (`:144`) is: "Bridge on E2B managed… hold Blaxel as a confirm-then-adopt second source", and `:76-86` classifies Blaxel as "PROMISING but UNVERIFIED on the two things that matter most… it claims 'microVM' yet never names Firecracker/KVM in its own docs — the escape-critical guarantee is branding, not a verifiable statement… trusted-pilot-only." Three written questions to Blaxel (`:136-139`) have no recorded answer.

`tech-choice.md:91` inverts this to "Blaxel is the primary short-term bridge", warranted solely by `:14-17` ("The owner has since confirmed that Blaxel's isolation is an acceptable hardware microVM class") — an owner assertion standing in for the vendor confirmation the evidence demanded. `references/sandbox-cost-model.md:27` still carries `microVM (hypervisor unnamed)` while `:136` calls Blaxel the "cheapest **confirmed** hardware-isolated sandbox SDK". The doc set contradicts itself on the one fact that matters, and the thesis is "Share the host, never the boundary" — so the boundary of the primary v1 path being unverified is the central unresolved risk. The price of resolving it is ~$150/mo (E2B base fee, `tech-choice.md:112`).

### 4. Cost math is right; the sequencing inverts it

The crossover arithmetic checks out: $0.0828/hr is consistent across E2B (`1×$0.0504 + 2×$0.0162`) and Blaxel (`2 GB × $0.0414`); $278 / $0.0828 = 3,357 hr/mo ✓; $41.40 at 500 hr ✓. But `references/sandbox-cost-model.md:163-170` concludes self-hosting only wins at ~5,000 hr/mo and warns the owned box at pilot scale is "$278–556/mo in fixed hardware plus real ops labor to serve $4–41 of usage" — while the ~3,000 hr/mo planning trigger appears only at `sandbox-sovereign-build.md:519-521`, inside **Phase 5**. Gate 0 → Phase 6 is 33–51 engineer-days (`build.md:28-29`) committed with no volume gate in front of it. You build the fleet, then dashboard whether you should have.

### 5. Anti-ceremony: ~1,278 lines land already-superseded

`references/README.md:18-24` labels six of thirteen reference files "Historical"/"superseded": `isolation-choices-primary-sources.md` (233), `gvisor-platform-security.md` (83), `managed-k8s-ch-eval.md` (222), `sbx14-scoping.md` (134), `control-plane-api-spec.md` (310), `e2b-internals-architecture.md` (296). `references/` is 2,945 of the 5,044 lines, and `control-plane-api-spec.md` (310) duplicates `api-spec.md` (290) with a "superseded" note instead of a deletion. Checking in 1,278 lines of self-declared-obsolete evidence is the anti-ceremony rule violated literally — git history is the archive. Ironically `sbx14-scoping.md:27` is the only file with real delta framing ("Today (delta framing): slices 1-3 already give the whole V1 client side + runsc runtime in-process") — the one grounded reality doc is filed as history while the forward plan has none.

### 6. Authority moved under it

PR #1317 merged 2026-08-17, landing `docs/plans/long-term/ratified/`. That pack mentions **no** Kata, Firecracker, SeaweedFS, or sovereign fleet. Main's HEAD added AGENTS.md hard rule 11 requiring architecture proposals to be reconciled with the ratified plan and never silently supersede a frozen ruling. The owner already applied exactly this to the adjacent sandbox PR #1256 on 2026-08-19. Same treatment applies here. Also, main added `docs/issues/1081/plan-blaxel-adapter.md` on 2026-08-12 which this PR's `README.md` index does not list. (#1256 is otherwise not in conflict — zero path overlap, and its body defers: "deployed V0 remains an explicit Agent-owned path until #1220 replaces it".)

### Fix list before merge

1. Reconcile with `docs/plans/long-term/ratified/ARCHITECTURE-PLAN.md` — mandatory under AGENTS.md rule 11. This is now the live question, not "approve/reject".
2. Add a real Today/Delta at the top of `sandbox-sovereign-build.md`, modeled on `api-spec.md:24-38`, and mark Gate 0's Blaxel item done-by-#1236 rather than pending.
3. Fix the two false runtime claims (the `remote-worker` frontmatter flag; `tech-choice.md:209`).
4. Delete the ~1,278 lines of self-declared superseded references.

**Owner decision needed:**
- **The Blaxel contradiction (§3):** either record the vendor answers to the three written questions, or state plainly that primary-bridge status rests on an owner risk acceptance over unverified evidence — and say what un-does it.
- **Sequencing (§4):** gate Phases 1–6 on the ~3,000 hr/mo trigger the doc's own economics imply, or state why schedule risk outweighs them.

PR #1236, opened 2026-08-12
feat(sandbox): Blaxel provider adapter (EU-sovereign bridge) — gh-1081

hachej:
## Summary

- register `blaxel` as a sandbox provider and runtime mode across sandbox, agent, core, CLI, workspace, and full-app wiring
- implement the `/workspace` Workspace and Sandbox contracts on `@blaxel/core` 0.3.11, pinned to an EU region with a persistent Blaxel Volume
- reconnect durable handles by deterministic sandbox name, recreate terminal compute against the same Volume, and journal crash-safe host seeding under a guest `flock`
- add provider conformance, security, provisioning, resolver/race, mode, handle-store, production-safety, and credential-gated live-smoke coverage

## Why this crosses package boundaries

Blaxel is a built-in runtime mode, not only a `boring-sandbox` provider. The Workspace/Sandbox pairing invariant requires one `RuntimeModeAdapter` to select both sides of the same execution context, so mode registration necessarily reaches agent, core, CLI, workspace, and full-app composition. Core also namespaces durable handles by provider so a workspace cannot reconnect a Blaxel runtime through a Vercel handle.

The small tsconfig/Vitest edits are source-build resolver plumbing: downstream packages compile the agent host from monorepo source, which statically imports the Blaxel provider subpath. Their explicit source aliases must therefore resolve that transitive import. This is existing monorepo alias duplication, not additional Blaxel behavior. Centralizing those aliases is follow-up repository tooling work rather than part of the provider adapter.

## SDK bridge and known limits

- `stat`, `rename`, and recursive `mkdir` use strictly quoted guest commands because SDK 0.3.11 has no matching filesystem primitives
- cancellation and timeout use named processes plus `process.kill()`; this is best-effort remote cancellation and cannot provide local-process signal fidelity
- the combined stdout/stderr byte limit is enforced locally after Blaxel returns terminal output because SDK 0.3.11 cannot cap output server-side; callbacks are intentionally not synthesized from terminal buffers
- snapshots/forking and Agent Drive remain out of scope because they are private-preview capabilities

## Validation

- `packages/boring-sandbox`: 55 files, 559 tests passed
- `packages/agent`: 221 files passed, 2 skipped; 2036 tests passed, 17 skipped; one unrelated 5-second credit-ledger timing test timed out and passed 13/13 when its file was rerun with `--testTimeout=15000`
- sandbox and agent package typechecks passed
- `pnpm lint:invariants` passed
- Blaxel production-safety gate: 2/2 passed
- credentialed live smoke reached Blaxel but resource creation is blocked by the configured account's `0/0` Volume quota; cleanup completed and identifiers/secrets are redacted

Refs #1081


hachej:
**Live EU smoke now passes** after the Volume quota grant: `{ok:true, region:eu-fra-1}` — sandbox + Volume created, attached, exercised, released. The smoke surfaced three busybox incompatibilities in the guest command layer (`flock -w`, `realpath -e/-m`, `--` operand) — fixed with portable equivalents; 46/46 provider tests and invariants green. The production blocker noted at PR open is cleared.

hachej:
Follow-up after scope review:

- documented why the built-in provider necessarily crosses `boring-sandbox`, agent, core, CLI, workspace, and full-app (paired Workspace/Sandbox runtime invariant + provider-namespaced durable handles)
- documented that the tsconfig/Vitest fan-out is existing source-build alias duplication, not provider behavior; centralizing it belongs in separate repo-tooling work
- fixed the Blaxel preflight lock to use a unique `mktemp` path
- raised the already-observed cold Core composition test allowance from 10s to 15s after CI measured ~11.8s

Targeted proof: Blaxel provider conformance 22/22; CoreWorkspaceAgentFront 26/26.

hachej:
Temporarily closing/reopening to retrigger required checks after the prior concurrency run remained stuck on an obsolete head.

PR #1241, opened 2026-08-12
[wt-391-forward-ktwq] fix(workspace): ChatLayout drawer dialog semantics (gh-1234)

### LINKED ISSUE #1234: ChatLayout drawers lack dialog semantics (focus trap, Escape, scroll lock)
## Problem
The ChatLayout left/right drawers render as plain positioned divs: no `role=dialog`/`aria-modal`, focus is not trapped, Escape does not close them, and background scroll is not locked while a drawer is open. Keyboard and screen-reader users can tab straight out of an open drawer into obscured content.

## Expected
Overlay drawers behave as modal dialogs: focus trapped inside while open, Escape closes, focus returns to the trigger on close, background scroll locked, `role=dialog` + `aria-modal=true` + labelled.

## Scope
`packages/workspace` ChatLayout drawer surfaces only. No visual redesign.

## Origin
UI review loop finding (bead wt-391-forward-ez6). Filed as the factory lifecycle test epic: intake → plan → gate 1 (inbox intention) → exec → owner playground validation via inbox.

hachej:
## Summary
- route Escape through one explicit active-drawer dismissal policy (workbench-left before Sessions)
- preserve unrelated workbench surfaces when dismissing the Sessions drawer in the default chat layout
- add regression coverage for modal labelling, initial focus, forward/reverse focus wrap, per-drawer Escape, trigger focus restoration, and shared body scroll locking

Closes #1234

## Proof of work

Current head: `3c477260747250bb43ff5cf226b84ccf4ea53d1f`

Automated verification:
- `pnpm --filter @hachej/boring-workspace exec vitest run src/front/layout/__tests__/presets.test.tsx` ✅ 51 tests passed
- `pnpm --filter @hachej/boring-workspace typecheck` ✅
- `pnpm lint:invariants` ✅
- Requested literal `pnpm --filter @hachej/boring-workspace test -- src/front/layout/__tests__/presets.test.tsx` ⚠️ pnpm forwards the separator as `vitest run -- ...`, so Vitest runs the entire package suite rather than the file; 1,886 tests passed, 2 unrelated server tests timed out, and 6 workers timed out under load. The correctly scoped Vitest command above is green.

Review:
- fresh-eyes pass found the default-chat Escape path closed an unrelated workbench; fixed in `3c4772607`
- second fresh-eyes pass: clean, no blocker
- thermo second pass (Sol xhigh / T1): clean

Screenshot/demo:
- N/A — semantic keyboard/focus behavior; DOM and focus assertions are the stronger proof.

Manual validation:
1. In a ChatLayout host, close Sessions, focus the visible **Sessions** edge button, and activate it.
2. Confirm focus enters the Sessions drawer. Press Tab repeatedly and Shift+Tab repeatedly; focus must wrap within the drawer.
3. Press Escape; only Sessions closes and focus returns to the **Sessions** trigger.
4. Open the workbench-left drawer with its visible host control. Confirm focus enters it and Tab/Shift+Tab cannot escape.
5. Press Escape; the workbench-left drawer closes and focus returns to its trigger.
6. While either drawer is open, attempt to scroll the page background; it must remain locked. If both are open, closing one must keep the lock until the other closes.

Known gaps:
- No visual redesign; no screenshot required by the approved plan.


hachej:
Proof of work — head `7562a62fdb014378c4f56fae888e837fd1ea58b7`

What changed:
- Complete ChatLayout Sessions/workbench-left modal dialog semantics and active-layer focus containment.
- Escape closes only the active drawer; trigger focus returns, with shell-local composer fallback for default-open drawers.
- Body scroll locking is preserved across overlapping drawers/shells; no visual redesign.

Automated verification:
- `pnpm --filter @hachej/boring-workspace exec vitest run src/front/layout/__tests__/presets.test.tsx` ✅ 55 tests
- affected WorkspaceAgentFront keyboard-focus regression ✅
- `pnpm --filter @hachej/boring-workspace typecheck` ✅
- `pnpm lint:invariants` ✅
- CI run 31602654487 attempt 2 ✅ (attempt 1 had an unrelated CLI 5s timeout)

Manual validation:
1. Open **Sessions**; focus enters it. Tab/Shift+Tab wrap within it.
2. Escape closes Sessions only and returns focus to its opener (or active chat composer for default-open Sessions).
3. Open **Workbench left panel**; focus enters it. Tab/Shift+Tab remain trapped.
4. Escape closes it and restores focus.
5. Background scrolling remains locked until all drawers close.

Review artifact: `.handoff/pr-1241-presentation.html` (workspace pane).
Known gap: none within gh-1234 scope.

hachej (commented):
THERMO-NUCLEAR REVIEW — VERDICT: REVISE / DO NOT MERGE

Reviewed the full diff at `7562a62fdb014378c4f56fae888e837fd1ea58b7`, the durable `wt-391-forward-ktwq` contract and gh-1234 Gate 1 comment (the referenced `docs/issues/1234/plan.md` is absent from the branch and `origin/main`), existing UI primitives, real drawer consumers, commit history, and focused tests.

## P1 — portaled child layers are broken by the bespoke drawer trap

`ChatLayout.tsx:805-815` redirects every `focusin` outside the drawer DOM subtree back into the drawer. That is not a valid modal-stack boundary: UI-kit `Dialog`, `AlertDialog`, `Sheet`, `Select`, and menus portal their interactive content to `document.body` (`packages/ui/src/dialog.tsx`, `alert-dialog.tsx`, `sheet.tsx`, `select.tsx`). This is an existing, reachable path: `FileTreeView.tsx:1005-1025` opens a portaled delete `AlertDialog`, and `FileTreeView.tsx:1242-1260` opens a portaled file-root `Select` from workbench-left.

After one animation frame, the new listener pulls focus out of the child layer and back into the drawer, potentially fighting the Radix FocusScope. The nested modal/select is therefore unusable or unstable.

Escape has the same layering bug. `ChatLayout.tsx:248-249` routes drawer Escape through `useKeyboardShortcuts`, whose listener is registered on `document` in capture phase (`useKeyboardShortcuts.ts:49-63`) and calls `preventDefault` before closing the drawer. A later-opened Radix layer cannot own Escape first; the underlying drawer closes while the child layer is still active or disappears only because its parent content unmounts. Nested modal/menu Escape must dismiss the top layer, not its drawer.

Required proof: open a portaled Dialog/AlertDialog/Select from a drawer, await an rAF, and assert focus remains in the child; then press Escape and assert only the child closes.

## P1 — multiple ChatLayout shells can enter a perpetual focus ping-pong

Every active trap installs its own document-level `focusin` redirect (`ChatLayout.tsx:805-824`). `activeDrawerStack` gates Escape only (`ChatLayout.tsx:209-217`); `handleFocusIn` never checks it. With shells A and B open, focusing B schedules A to redirect into A; that focus schedules B to redirect into B, alternating each animation frame.

The new multi-shell test at `presets.test.tsx:1315-1327` is green theater for this case: it focuses and fires Escape synchronously before a frame runs, then teardown cancels the queued redirect. It does not prove multi-shell focus containment.

## P1 — modal infrastructure is in the wrong abstraction layer

This PR deepens a bespoke document-global modal manager inside an already oversized layout component: module-global drawer stack, global scroll counter/style snapshot, document listeners, rAF focus redirects, derived top-drawer flags, and shell-specific focus fallback. `ChatLayout.tsx` grows from 1,129 to 1,191 lines.

`packages/ui` already provides canonical Radix-backed `Sheet` and `Dialog` primitives. Radix already composes FocusScope, dismissable-layer stacking, nested-layer Escape behavior, focus restoration, and `react-remove-scroll`; workspace already uses `Sheet` for analogous mobile navigation in `PluginTabsWorkspaceShell.tsx`. The code-judo fix is to use or extend that canonical primitive for persistent/inline drawer content, leaving ChatLayout with one explicit active-drawer model and `onOpenChange`, rather than maintaining parallel modal infrastructure.

## P2 — scope expansion

The approved bead covers the two drawers within one ChatLayout. Cross-shell global arbitration is extra scope: module-global `activeDrawerStack` plus the cross-shell Escape test, and module-global scroll refcount plus its cross-shell test. The bead history itself records multi-ChatLayout global arbitration as out of gh-1234 scope. The hardening required for overlap inside one layout is in scope; cross-shell policy is not.

## P2 — incomplete/weak lifecycle proof

- `presets.test.tsx:1207-1242` says the session drawer closes, but after Meta+1 only repeats that workbench has focus; it never asserts Sessions actually closed, so the test can pass if the shortcut is a no-op.
- No focused test covers an opener unmounting during close or proves the delayed composer fallback cannot steal focus after a later interaction.
- No nested portal/modal test covers the P1 paths above.

No skipped/todo/only tests were found. The old `closeSurface` assertion changing from two calls to one matches the intended active-drawer Escape policy and is not dishonest weakening. The focused suite is green, but it does not exercise the failing layer interactions.

## P3 — test isolation

Viewport and body-overflow restoration are inline rather than guaranteed by `afterEach`/`finally`, so an earlier failed assertion can contaminate later tests.

SSR access is effect-bound/guarded. Ordinary single-document body-lock cleanup and the trigger-unmounted fallback look sound. Those clean paths do not outweigh the P1 modal-layer correctness and maintainability failures.

hachej:
Response to the thermo-nuclear review at `7562a62fd` — remediated in `f33419237`.

Approach chosen:
- I investigated the existing UI-kit Radix `Sheet`/`Dialog` primitives first.
- Direct adoption is not safe for these drawers: `SheetContent` always portals/fixes its content and overlay, while ChatLayout’s drawers are inline flex children with width animation/resizing; normal Radix content unmounts when closed, breaking the intentionally persistent Sessions panel, while `forceMount` leaves closed FocusScope/DismissableLayer, `hideOthers`, and RemoveScroll behavior mounted. Two roots would also use mount order rather than ChatLayout’s explicit desktop/mobile visual priority.
- I therefore used the review’s scoped fallback: extracted the remaining inline-drawer lifecycle into `useModalDrawer.ts`, with subtree-only keyboard handling and shell-local scroll/focus behavior.

P1 resolutions:
1. Portaled child layers: removed the document-wide `focusin` redirect entirely. The drawer only wraps Tab within its own subtree, so portaled Radix Dialog/AlertDialog/Select/menu content keeps focus.
2. Escape layering: removed drawer Escape from the capture-phase `useKeyboardShortcuts` path. Escape now bubbles only within the active drawer subtree; a portaled nested Radix layer receives and dismisses itself first.
3. Multi-shell ping-pong: removed all drawer `requestAnimationFrame` focus redirects and the module-global drawer stack. The two-shell regression test asserts focusing the second shell schedules no rAF redirect and remains stable.
4. Abstraction/size: modal lifecycle moved out of `ChatLayout` into a 90-line dedicated hook module; `ChatLayout.tsx` is now 1,069 lines versus the reviewed 1,191.

P2/P3 resolutions:
- Removed cross-shell Escape arbitration and global scroll-lock refcount/policy; scroll locking is local to one ChatLayout and still covers overlap between its two drawers.
- The overlap test now explicitly asserts Sessions is collapsed after Meta+1 before checking workbench focus.
- Added proof for nested Radix focus + Escape ownership, two-shell no-rAF behavior, disconnected opener fallback, and delayed composer focus not stealing a later interaction.
- Viewport, timers/mocks, and body styles now clean up in `afterEach`.
- Added the missing `docs/issues/1234/plan.md` and `plan-review.html`.

Proof at `f33419237`:
- `pnpm --filter @hachej/boring-workspace exec vitest run src/front/layout/__tests__/presets.test.tsx` — 56/56 passed.
- `pnpm --filter @hachej/boring-workspace typecheck` — passed.
- `pnpm lint:invariants` — passed.
- Fresh read-only remediation review — CLEAN; no P1–P3 blocker remains.

CI is running on the new head; I will follow it to conclusion.

hachej:
Follow-up at `786eab28c`: the first CI run exposed a full-suite integration case where switching sessions replaces the focused session-row node. I fixed it without reintroducing document-global focus management:

- drawer callbacks now live in refs, so host callback identity changes cannot rerun autofocus;
- a drawer-subtree-only `MutationObserver` repairs focus only when a child replacement leaves `document.activeElement === document.body`;
- the integration test now sends Escape through `user.keyboard`, exercising the real focused bubble path instead of dispatching directly on `document`.

This cannot interfere with nested Radix portals: their mutations occur under `document.body`, and while a child layer owns focus the observer’s body-focus guard is false. It also schedules no rAF and cannot create cross-shell focus ping-pong.

Current-head local proof:
- WorkspaceAgentFront integration file: 90/90 passed.
- ChatLayout presets: 56/56 passed.
- Workspace typecheck: passed.
- `pnpm lint:invariants`: passed.
- Follow-up independent review: CLEAN.

A fresh CI run is now running on `786eab28c`; I will poll it to conclusion.

hachej:
Final CI update: run 31620780305 is green at head 6d1e01378. Changed-unit passed (11m17s), UI Review passed (10m37s), and typecheck, lint/invariants, E2E, bundle budget, and reference-image/remote-worker smoke all passed. The prior affected-unit failures were the same cold-start CLI integration test exceeding Vitest’s 5s default; its existing subprocess budget is 10s and neighboring cold-start tests use 15–30s, so commit 6d1e01378 applied the existing 15s test-only convention. The two UI rerun mismatches were nondeterministic and occurred in different replay states; the unchanged third replay and deterministic component baselines passed without weakening thresholds. PR remains unmerged.

PR #1260, opened 2026-08-13
fix(cli): allow authorized repository skill symlinks

hachej:
## Problem
CLI workspace startup failed when a repository-managed Pi skill symlink such as `.pi/skills/exec -> ../../.agents/skills/exec` was encountered:

```text
Pi resource symlinks are not allowed: .../.pi/skills/exec
```

The runtime-scope error was a secondary setup failure caused by this rejected resource digest.

## Fix
- Allow symlinks in CLI mode only when their resolved target remains inside authorized roots.
- Continue rejecting symlinks by default in workspace/remote modes.
- Include the symlink target in the resource digest.
- Add regression coverage for authorized internal symlinks.

## Validation
- Workspace resource tests: 58 passed.
- CLI resource digest measured on `boring-ui-v2`: 393ms.
- Agent typecheck passed.
- CLI bundle build passed.

Full CLI browser startup timing could not be completed because this checkout lacks a built CLI frontend and has unrelated stale workspace/plugin build artifacts; `pnpm build:full` is required for that environment-level check.

### DIFF
--- a/packages/agent/src/server/piResourceDigest.ts
+++ b/packages/agent/src/server/piResourceDigest.ts
@@ -39,6 +39,8 @@ export interface PiResourceDigestInput {
   readonly extensionPaths?: readonly string[]
   /** Lexical roots explicitly authorized by the embedding Workspace/CLI. */
   readonly authorizedRoots: readonly string[]
+  /** CLI workspaces may contain repo-managed symlinks (e.g. .pi/skills/*). */
+  readonly allowInternalSymlinks?: boolean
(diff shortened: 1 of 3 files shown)


hachej:
The required UI Review failed in the existing CI run due to a screenshot replay mismatch in the unrelated workspace command-palette fixture (`UI_REVIEW_REPRODUCE_SCREENSHOT_MISMATCH`, desktop explore-0017). The implementation does not touch UI code. I reran the failed jobs; monitoring the retry.

PR #1263, opened 2026-08-13
fix(cli): restrict internal symlink traversal

hachej:
## Problem
PR #1260 allowed all internal symlinks in CLI resource digests. On large workspaces such as Seneca, this can cause the digest to traverse dependency trees and saturate the CLI process, making even `/api/v1/workspace/meta` slow or time out.

## Fix
Keep the security exception narrow: CLI may follow symlinks only beneath the workspace's `.pi/skills` root, and only when the resolved target remains inside authorized roots. Symlinks elsewhere remain rejected, including `.pi/npm/node_modules`.

## Validation
- Agent typecheck passed.
- Workspace resource digest tests: 58 passed.
- Regression preserves rejection of external symlinks and allows authorized internal links.


hachej:
Superseded: rather than narrowing symlink traversal, we will remove the unnecessary CLI symlink exception. CLI already loads `.agents/skills` directly.

PR #910, opened 2026-07-22
#900 spike: prove full-catalog Composio capabilities

hachej:
## Summary

Complete issue #900's live Composio capability gate before changing product contracts.

**Result: pass. Slice 900.1 implementation may proceed.**

## Issue / Slice

#900, slice 900.1 preflight.

## What Changed

- Adds a repeatable, bounded full-catalog Session capability command.
- Records actual sanitized full-catalog and exact account-pin proof.
- Narrows the fastest launch to exactly one active account per toolkit; zero/multiple accounts fail closed and replacing requires revoke-then-connect.
- Adds no runtime `src/` implementation.

Live findings:

- Omitting `toolkits` exposes full-catalog GitHub search/schema.
- Raw v3.1 requires `workbench: { enable: false }`; the SDK `sandbox` alias is not normalized by this endpoint.
- Session exposes search/schema/manage-connections/multi-execute; workbench/bash are absent.
- Session headers alone return 401; project API key is required at exact allowlisted `https://backend.composio.dev`.
- Owner selected the existing Composio project.
- An owner-authorized temporary GitHub account proved unique ownership, exact returned Session pin, cross-user rejection, and a harmless authenticated identity read.
- The account was revoked and both Sessions were deleted and verified absent.

## Proof

- Full-catalog sanitized spike: exit 0 at `2026-07-22T20:50:37.665Z`.
- Exact account-pin proof: exit 0 at `2026-07-24T18:42Z`.
- `node --check plugins/boring-mcp/scripts/composio-capability-spike.mjs` — passed.
- `pnpm build:packages` — passed.
- `pnpm --filter @hachej/boring-mcp test` — 13 files / 99 tests passed.
- `pnpm --filter @hachej/boring-mcp typecheck` — passed.
- `pnpm --filter @hachej/boring-mcp build` — passed.
- `git diff --check` — passed.

## Review

- Initial security/spec review found origin, error-redaction, cleanup, bounds, and overclaim gaps; all were fixed.
- Final capability review: clean.
- Final narrowed single-account invariant/proof review: clean.
- Reviewed SHA: `c859ada38`.

## Security

- Exact MCP origin allowlist before forwarding the operator key.
- Redirect rejection, 15-second timeout, 512-KiB body cap, result/schema caps.
- Raw provider errors replaced with phase plus numeric codes.
- No existing/customer connected-account list or app-native customer tool was called.
- Every Session cleanup attempt is aggregated and requires subsequent 404.
[...]

PR #1309, opened 2026-08-15
[boring-mcp-plugin]

hachej:
## Summary

- Implements the Slice 900.1 discovery-only Composio backend behind an optional managed catalog seam.
- Creates unfiltered Sessions with `workbench.enable=false`, fixed-origin MCP transport, controlled search/schema calls, bounded transient metadata, exact account selection/pinning, opaque subjects, and retained verified cleanup leases.
- Keeps raw Composio execution/control/workbench/bash tools unreachable and template-free catalog entries non-executable.
- Preserves curated Notion/Airtable transport behavior; adds no 900.2 approval/call path, UI, release, custom URL, Registry, stdio, or vault work.

## Proof

- `pnpm --filter @hachej/boring-mcp typecheck` — passed.
- `pnpm --filter @hachej/boring-mcp test` — 14 files / 199 tests passed.
- `pnpm --filter @hachej/boring-mcp build` — passed (front/server/shared bundles and declarations).
- `pnpm --filter full-app typecheck` — passed.
- `git diff --check` — passed.
- Deterministic fake Composio HTTP + real MCP SDK fake Streamable HTTP coverage includes full catalog, pagination/bounds, zero/multiple/exact accounts, OAuth/session/operator-key canaries, exact pin echo, cleanup retry/drain, schema/source drift, curated compatibility, and meta-tool denial.

## Review provenance

- Fresh-eyes: `openai-codex:gpt-5.6-sol`, adversarial Slice 900.1/security mandate, SHA `dca76921154c1f7fd5a682f5a03ce3fe27fcad93` — **CLEAN** (`.handoff/reviews/fresh-eyes-dca7692.txt`).
- Thermo-nuclear maintainability: `openai-codex:gpt-5.6-sol`, structural/code-judo/lifecycle mandate, SHA `dca76921154c1f7fd5a682f5a03ce3fe27fcad93` — **REVISE** (`.handoff/reviews/thermo-dca7692.txt`).
- Open thermo findings: backend lifecycle is not a single closeable app-owned runtime; generation is global rather than source-local; account parser mirrors excess provider internals; managed probe reports synthetic success.

## Artifacts

- Draft implementation and proof are on this PR at exact head `dca76921154c1f7fd5a682f5a03ce3fe27fcad93`.
- No owner/orchestrator validation intention was created because mandatory thermo review is not green.

## Risk & rollback

- **State:** draft / not ready for merge.
- Residual risk: lifecycle/shutdown ownership can lose post-drain cleanup work; global generation can create noisy-neighbor cache churn; parser/probe boundaries need redesign.
- Rollback: do not merge; abandoning this branch leaves current curated behavior unchanged. If merged accidentally, revert the branch commits as one PR.

## Refs

- Bead `wt-391-forward-rjkl.2`
- Issue #900, Slice 900.1
- Capability spike PR #910
- Historical reland/revert PRs #937 and #946; receiver regression #941


hachej:
Mandatory review stop — not ready for owner/orchestrator validation.

Thermo review (`openai-codex:gpt-5.6-sol`, fresh read-only context, exact SHA `dca76921154c1f7fd5a682f5a03ce3fe27fcad93`) verdict: **REVISE**.

Open material findings:
1. Bind one application-owned runtime across route/Agent surfaces; close must stop admission, await active work, and drain every cleanup lease.
2. Replace backend-global cache generation with bounded source-local fencing.
3. Reduce account parsing to the minimal authoritative selection DTO rather than mirroring unrelated auth/provider fields.
4. Make managed probe perform a bounded health verification or return explicitly unsupported; do not synthesize success.

Per bead policy: no final intention and no merge request while review is not green.

hachej:
Owner decision

{"decision":"request-changes","notes":"in r 1317 we have explroe how opencode is sutrcutreing the tool mcp to avoid bloating context.. please review this in detail and adpat the apparoch we have to be inlined wit hthis "}

hachej:
## 🌙 Overnight review — NEEDS-REWORK (keep as draft — do not merge)

Correctly parked as a draft, and the body honestly lists four unresolved findings. CI 12 passed. The following are findings *beyond* the ones already listed, which is where the value is.

**D1 (decisive) — the account-row validator requires a field that appears to be invented.** `composioCatalogBackend.ts:397-398` demands every connected-account row own the properties `["word_id", "alias", "status_reason"]`, else the row is rejected with `Composio connected-account row was incomplete` (`:405-419`). `word_id` is not a field in Composio's connected-accounts API — it reads like a corrupted `workspace_id`. The only thing that proves it valid is the PR's own fake, written to satisfy the code: `composioCatalogBackend.test.ts:43` emits `word_id: \`word-${id}\``. If the real API does not return `word_id`, **every** real account is rejected and account selection never works — while 199 tests stay green. This is the exact "deterministic fake confirms the implementation, not the provider" failure mode. Verify against the live API before anything else; the whole strict-mirror block (`is_composio_managed`, `state.authScheme`, `created_at` as string, …) is brittle for the same reason and should validate only what the code uses.

**D2 (real bug) — the "per minute" budget is a per-idle-gap budget.** `guarded()` does `requestBudget.set(key, {calls: budget.calls + 1})` (`:708`), and `BoundedTtlCache.set` **resets `expiresAt` on every write** (`:95-97`). The 60s window never elapses while the source keeps calling: after 60 total calls a source is locked out with `RESOURCE_LIMIT_EXCEEDED` until it goes a full 60s completely idle. Needs a windowStart timestamp, not a TTL refresh.

**D3 (real bug) — one workspace's stuck cleanup lease breaks the catalog for everyone.** `guarded()` starts with `await drainPending()` (`:699`), and `drainPending` rethrows the first cleanup error (`:683`). `pendingCleanup` is a single backend-instance map keyed by session id (`:658`). So one Composio session that refuses to delete makes **every** subsequent `searchTools`/`describeTool` call, for every workspace and user on that host, fail with a cleanup error unrelated to the request. Drain should be best-effort/background, never a precondition that can fail the caller.

**D4 (security-adjacent) — unbounded in-memory retention of operator secrets.** `retainFailedCleanup` stores `{secret, sessionId}` (`:660-662`) in an unbounded `Map` with no TTL and no cap — the only unbounded structure in a file whose entire thesis is boundedness, and it's the one holding secret material.

**D5 (correctness, low) — global cache generation.** Already on the author's list, but concretely: `cacheGeneration` is one counter for all sources (`:657`), so a `forceProviderRefresh` on source A causes an in-flight result for source B to fail the `requestGeneration === cacheGeneration` guard (`:754`, `:784`) and never be cached. Under any concurrency this quietly disables caching.

**D6 (dead code)** — `composioCatalogBackend.ts:771`: `const sourceKey = sourceCacheKey(source)` shadows the outer `sourceKey` inside the `else` branch and is never read. Small, but it means neither lint nor three "clean/no material findings" reviews caught an unused shadowed local.

**D7 (over-engineering / reachability) — nothing constructs this.** `createComposioCatalogBackend` has no caller anywhere outside tests and the README; `managedCatalog?` is threaded as an optional parameter through six option interfaces (`appServerBinding.ts` ×5, `sourceHandlers.ts`, `toolCatalog.ts`). 1616 lines of provider code reachable only from a hand-written fake, plus a new session created and deleted per catalog call (two extra provider round trips each), plus a global 4-request concurrency ceiling that *throws* rather than queues (`:700-702`). For a "discovery-only slice" this is a large speculative surface. Also `probeSource` returns a synthetic `{tools: [], resources: []}` success for managed sources (`sourceHandlers.ts`) — a health check that cannot fail is worse than no health check.

**Recommendation:** confirm `word_id` against the live Composio API first (D1 — if it's invented, the account path has never worked and the suite is self-confirming), then D2/D3/D4. My recommendation is to shrink the scope substantially rather than resolve the thermo findings in place: cut the per-call session churn and the unreachable optional seam down to what slice 900.1 actually needs.

hachej:
Closing per owner decision. This draft implementation remains quarry/evidence only and is superseded by the combined inbound + outbound MCP planning work in #1415. Do not merge or revive this implementation branch; reland only through the reviewed serial slices and canonical dependency gates recorded there.

PR #1310, opened 2026-08-15
[Agent Package Lifecycle] Contract package isolation and retry lifecycle

hachej:
## Owner Review

**Bead / PR / issue:** `wt-391-forward-0ms8.7` · [PR #1310](https://github.com/hachej/boring-ui/pull/1310) · Agent Package Lifecycle  
**What changed / why:** Contracts local Agent package discovery, sibling failure isolation, truthful catalog identity, and retryable pre-effect request lifecycle. Current main's Gateway Retry implementation is now integrated as the single request-ledger state machine.  
**Why you are needed (matched protected boundary):** **PROTECTED**. The complete current-main diff changes public Agent contracts (`AgentRequestLedger`, catalog definition identity, Host/fleet behavior) and package-boundary seams used by Workspace, CLI, standalone, and playground compositions.  
**Recommendation / alternatives:** Review and approve exact artifact head `241ea3181459032016fe68bdb48503bb85190e29` after checking the reviewed code `212c346637174d147811c924c5c7ac1ca0b7d6c7`. Alternative: defer/close the PR; do not revive the stale approval at `dbdcdf41`.  
**Scope / base / head SHA:** current base/main=`d1c6968e2670b089703809ef632d960f8ac950e6`; merge=`2339a8469b3d6aec416e353f6c04a77e4416ebe5`; reviewed product/integration code=`212c346637174d147811c924c5c7ac1ca0b7d6c7`; docs-only artifact=`241ea3181459032016fe68bdb48503bb85190e29`; prior artifact=`0e58214f5cb0e01f1bc8b5a246ba57f34dfceb6e`.  
**What approval authorizes:** Merge of PR #1310 at exact artifact head `241ea3181` through protected `main` only. It does not authorize publish, release, deployment, file deletion, force-push, or later material revisions.  
**Risk / rollback:** Residuals are an uninspected out-of-repository implementation of the public ledger interface and permanent SQLite tombstone growth by design. Before merge, defer/close. After merge, revert the PR merge commit through protected main; no migration or data deletion is introduced.  
[...]

PR #1311, opened 2026-08-15
[wt-391-forward-xp3s.9] Reconcile persisted default Agents safely

hachej:
## Summary
- inventory persisted default-Agent cohorts and compare-and-set only legacy `NULL` rows to the one validated application default before routes serve
- preserve non-`NULL` values, fail startup when reconciliation does not converge, and close Core/AgentHost resources on every migration failure
- deny Agent execution stably when a persisted default is unavailable without blocking Workspace listing or session/history reads
- keep hostname out of repair/resolution and retain the existing single `createAgentHost()`/AgentGateway construction funnel

## Initial bug description

Migration `packages/core/drizzle/0024_workspace_default_agent_type_id.sql` deliberately introduced `default_agent_type_id` as nullable, leaving pre-existing Workspaces in the legacy `NULL` cohort. Before this PR, `packages/core/src/server/defaultAgentType.ts` also handled an unavailable persisted non-`NULL` Agent id by silently selecting the boot/first-fleet fallback. The result was an unreconciled legacy cohort plus a risk that a Workspace recorded for one Agent could execute a different Agent without an explicit migration or stable error.

This PR fixes that initial bug by inventorying cohorts at startup, compare-and-setting only app-scoped `NULL` rows to the validated application default before routes serve, preserving every non-`NULL` value, and denying Agent execution stably when that persisted Agent is unavailable while leaving Workspace list/session/history reads available.

**Owner request-changes addressed:** “can you just add to the handoff the init abug decirption ?” The same evidence-backed description is now the first prose section of `.handoff/pr-1311-handoff.html`.

## Proof
Exact head: `97e99d5651a2fb4261860b01deb401a368e990ad`

- `BORING_AGENT_FLEET=0 NODE_ENV=test pnpm --filter @hachej/boring-core exec vitest run src/server/__tests__/defaultAgentType.test.ts src/app/server/__tests__/createCoreWorkspaceAgentServer.provisioning.test.ts src/server/db/stores/__tests__/LocalWorkspaceStore.test.ts src/server/db/stores/__tests__/PostgresWorkspaceStore.test.ts src/server/db/__tests__/defaultAgentTypeId.migration.test.ts --no-file-parallelism` — **5 files, 184 tests passed**
- `BORING_AGENT_FLEET=0 NODE_ENV=test pnpm --filter @hachej/boring-workspace exec vitest run src/app/server/__tests__/createWorkspaceAgentServer.test.ts src/app/front/__tests__/WorkspaceAgentFront.test.tsx --no-file-parallelism` — **2 files, 152 tests passed**
- `pnpm --filter @hachej/boring-core typecheck` — **passed**
- `pnpm check:agenthost-cutover-matrix` — **20 rows, 42 final routes, 23 deleted historical routes, zero forbidden compatibility references**
- `git diff --check origin/main...HEAD` — **passed**

Focused cases cover NULL CAS/idempotency, already-set preservation, competing CAS winner, app isolation, unknown cohort inventory, config/default unification, migration cleanup/convergence, stable unknown-default execution denial, meta 409, session/history readability, rollback-compatible NULL reads, and no silent switch.

CI: final-head run `31893499413` is green after one failed-job rerun. The first E2E attempt hit the unrelated Agent property-baseline running-tool flake on both Playwright retries; the rerun passed E2E and every required CI gate. Workflow Invariants run `31893499427` is green.

## Review provenance
- reviewer: `openai-codex/gpt-5.6-sol`
- mandate: fresh-context adversarial fresh-eyes + thermonuclear review; refute scope, CAS/race safety, execution/read split, rollback, hostname isolation, and AgentHost funnel
- target: `97e99d5651a2fb4261860b01deb401a368e990ad`
- verdict: **clean**; no material findings or residual risk
- earlier reviews at `3b0e8a8` / `4d65b58` found startup cleanup, NULL convergence, config-default split, and error-envelope issues; all were fixed and re-reviewed at the final SHA

## Artifacts
- r2 present-pr handoff: `.handoff/pr-1311-handoff.html` (regenerated with the initial bug description, live CI table, issue/bead header, and opened before the r2 owner gate)
- running surface: N/A — backend/store behavior, no UI surface changed

## Risk & rollback
- Risk: startup now intentionally fails if the configured application default is absent from the fleet or a post-CAS `NULL` cohort remains; this prevents serving ambiguous Agent behavior.
- Rollback: revert the five bead commits. The schema remains nullable, session/history reads are unchanged, and migrated values remain readable by the prior default-aware cohort. Unknown non-`NULL` rows are never rewritten.

## Refs
- Bead: `wt-391-forward-xp3s.9`
- Epic D revision 3
- Decisions 28 and 29

### DIFF
--- a/packages/core/src/app/server/__tests__/createCoreWorkspaceAgentServer.provisioning.test.ts
+++ b/packages/core/src/app/server/__tests__/createCoreWorkspaceAgentServer.provisioning.test.ts
@@ -12,6 +12,12 @@ const mocks = vi.hoisted(() => {
   const hostRegisterDirectRoutes = vi.fn((_projection: any) => async () => {})
   const hostClose = vi.fn(async () => {})
   const acquireEnvironment = vi.fn()
+  const gatewayReadSessionState = vi.fn(async ({ ref }: { ref: unknown }) => ({
+    ref,
+    seq: 0,
+    summary: {},
+    state: { messages: [] },
+  }))
   return {
     createAgentHost: vi.fn(async (options: any) => {
       await options.fleetCompiler.compile({ agents: options.agents })
@@ -20,18 +26,22 @@ const mocks = vi.hoisted(() => {
         host: { close: hostClose, drain: vi.fn(async () => {}) },
         registerDirectRoutes: hostRegisterDirectRoutes,
         acquireEnvironment,
+        gateway: { readSessionState: gatewayReadSessionState },
         runWithWorkspaceAgent: vi.fn(),
       }
     }),
     hostRegisterDirectRoutes,
     hostClose,
     acquireEnvironment,
+    gatewayReadSessionState,
     provisionWorkspaceRuntime: vi.fn(async () => ({ changed: false, env: {}, pathEntries: [], skillPaths: [] })),
     collectWorkspaceAgentServerPlugins: vi.fn(),
     createWorkspaceUiTools: vi.fn(() => []),
     isMember: vi.fn(async (_workspaceId: string, _userId: string) => true),
-    getWorkspace: vi.fn(async (id: string) => ({ id, appId: 'test-app' })),
+    getWorkspace: vi.fn(async (id: string) => ({ id, appId: 'test-app', defaultAgentTypeId: 'default' })),
     getUser: vi.fn(async (id: string) => ({ id })),
+    inventoryDefaultAgentTypeIds: vi.fn(async (_appId: string): Promise<Array<{ defaultAgentTypeId: string | null; count: number }>> => []),
+    compareAndSetNullDefaultAgentTypeId: vi.fn(async (_appId: string, _value: string) => 0),
     actualCreateAgentHost: undefined as undefined | ((options: any) => Promise<any>),
     runtimeHost: {
       source: 'custom-adapter-host',
@@ -120,9 +130,12 @@ beforeEach(() => {
   vi.clearAllMocks()
   mocks.provisionWorkspaceRuntime.mockResolvedValue({ changed: false, env: {}, pathEntries: [], skillPaths: [] })
   mocks.acquireEnvironment.mockReset()
+  mocks.gatewayReadSessionState.mockClear()
   mocks.isMember.mockResolvedValue(true)
-  mocks.getWorkspace.mockImplementation(async (id: string) => ({ id, appId: 'test-app' }))
+  mocks.getWorkspace.mockImplementation(async (id: string) => ({ id, appId: 'test-app', defaultAgentTypeId: 'default' }))
   mocks.getUser.mockImplementation(async (id: string) => ({ id }))
+  mocks.inventoryDefaultAgentTypeIds.mockResolvedValue([])
+  mocks.compareAndSetNullDefaultAgentTypeId.mockResolvedValue(0)
 })
 
 vi.mock('../../../server/routes/index.js', () => ({
@@ -140,6 +153,10 @@ vi.mock('../../../server/db/index.js', () => ({
   PostgresWorkspaceStore: class {
     get(id: string) { return mocks.getWorkspace(id) }
     isMember(workspaceId: string, userId: string) { return mocks.isMember(workspaceId, userId) }
+    inventoryDefaultAgentTypeIds(appId: string) { return mocks.inventoryDefaultAgentTypeIds(appId) }
+    compareAndSetNullDefaultAgentTypeId(appId: string, value: string) {
+      return mocks.compareAndSetNullDefaultAgentTypeId(appId, value)
+    }
   },
 }))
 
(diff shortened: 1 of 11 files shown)


hachej:
Owner decision

```json
{"decision":"request-changes","notes":"can you just add to the handoff the init abug decirption ? "}
```

hachej:
## 🌙 Overnight review — MERGE-AFTER-FIXES

CI 14/14 green on head `97e99d5` — real green. The store-level CAS and the execute-denies/reads-survive split are the right design and the proof is real.

**Verified sound:** the CAS is genuinely safe — `UPDATE … WHERE app_id = ? AND default_agent_type_id IS NULL` (`PostgresWorkspaceStore.ts:824-833`) never rewrites a non-NULL row, is idempotent, and the conformance test proves exactly-one concurrent winner (`storeConformance.ts:721-734`). Also a genuine leak fix at `createCoreWorkspaceAgentServer.ts:1825-1827`: the old `if (hostMounted) app.close() else host.close()` never called `app.close()` (hence never `sql.end()`) on a pre-mount boot failure.

**Findings:**

- **C1 (medium) — rolling-deploy boot self-DoS.** `createCoreWorkspaceAgentServer.ts:1700-1707`: after the CAS, if a *second* inventory still shows any NULL row, boot throws `Workspace default Agent legacy reconciliation did not converge`. The only way a NULL appears between the two reads is a concurrent writer — i.e. an *old* instance still serving during a rolling deploy creating a workspace with a NULL default. The new instance then refuses to start, and keeps refusing while the old one serves traffic. The check should be convergence over the rows observed in `inventoryBefore`, or a warn-and-continue with retry. As written, the strictest possible interpretation is applied to the one situation the code cannot control.
- **C2 (low-medium) — the meta route uses a different default than everything else.** `:1746` passes `bootDefaultAgentTypeId: options.defaultAgentTypeId` (raw option), while `assertWorkspaceDefaultAgentExecutionAvailable` (`:1614`) and the backfill both use `applicationDefaultAgentTypeId` (`options.defaultAgentTypeId ?? rawConfig.defaultAgentTypeId`, validated). With `config.defaultAgentTypeId = 'reviewer'` and no boot option, a NULL-persisted workspace resolves to `agents[0]` from `/workspace/meta` but `'reviewer'` everywhere else. The same line also re-derives `agents.map(a => a.agentTypeId)` instead of reusing the hoisted `availableAgentTypeIds`. The commit "unify configured default initialization" and its test are one call site short of true. **Fix: pass `applicationDefaultAgentTypeId` and `availableAgentTypeIds` at `:1746-1747`.**
- **C3 (design fragility, low) — WeakMap-keyed workspace id.** `createCoreAgentScopeAuthority` adds a second WeakMap `workspaceIdsByClaim` (`:357`, `:379`, `:395-399`) and `coreEffectAdmission` looks the workspace id up by *object identity* of the verified claim (`:1627`). Works today only because `runtime.verify()` returns `record.claim` unchanged (`createAgentHost.ts:340-346`). Any future claim crossing a process boundary (the remote AgentGateway this epic is heading toward), a `structuredClone`, or a JSON round trip turns every execution effect into a thrown bare `Error` — which from inside `admit()` is not an `AgentGatewayError`, so it poisons the ledger record rather than denying cleanly. The claim already carries `workspaceScopeId = JSON.stringify([workspaceId, sessionNamespace])`; parsing it back is identity-free and one line.
- **C4 (edge, low) — 403 inside `admit()` poisons the request id.** `assertWorkspaceDefaultAgentExecutionAvailable` throws `httpError('workspace access denied', 403)` when the workspace vanished (`:1611`). `admit()` converts only `DEFAULT_AGENT_TYPE_UNKNOWN_SEAT` into a clean `rejected` (`:1631-1642`), so a deleted-mid-session workspace surfaces as `AGENT_REQUEST_OUTCOME_UNKNOWN` instead of a clean denial.
- **C5 (cost, low) — extra DB reads per prompt.** `verify()` already fetches the workspace (`:410-414`), `assertWorkspaceDefaultAgentExecutionAvailable` fetches it again for every `session.create/prompt/followup/command.execute` (`:1610`), and `directDispatcherResolver.runWithWorkspaceAgent` fetches it a third time (`:1788`). Two extra round trips per prompt for a value that changes ~never.

**Over-engineering:** `classifyWorkspaceDefaultAgentTypeCohorts()` is an exported, separately unit-tested API returning `{nullCount, knownCount, unknown[]}` where only `nullCount` is consumed as logic (`:1701`) — `knownCount`/`unknown` exist to decorate one `log.info`, at the cost of two aggregate queries per boot. Could be a local `countNulls()`. Also `hasErrorCode()` walks `.cause` 4 levels deep for a single `42P01` probe; depth 4 is arbitrary and deserves a constant or comment.

**Test-integrity note:** `createCoreWorkspaceAgentServer.provisioning.test.ts:238-240` **weakens an existing invariant** — `expect(hostOptions.effectAdmission).toBe(effectAdmission)` became `not.toBe(...)` + `toMatchObject({admit: expect.any(Function)})`. The wrapper justifies the change, but the replacement asserts almost nothing; it should assert the wrapper *delegates* to the injected admission (call-through), which is the property the original test protected. Otherwise the test set is honest and unusually thorough (parametrised failure stages, host-close-on-failure, CAS ordering vs route registration, 409 envelope, rollback-compatible NULL reads).

**Note:** the interface addition to exported `WorkspaceStore` (`server/app/types.ts:56-62`) is a breaking change for any out-of-tree implementor. In-tree there are only the two.

**Cross-PR duplication worth naming:** this PR and **#1310** independently invent the same "configured default agent must be in the fleet, else fail boot" rule — here via `resolveApplicationDefaultAgentTypeId()` (core), there via `ConfiguredDefaultAgentError` (workspace). Same rule, two classes, two packages, one exec stack. One shared helper in `@hachej/boring-agent` would do. (The two PRs are *not* stacked — independent siblings off the same old base `ee20171`, both 35 commits behind main.)

**Fix list before merge:** C2 (one-line default unification at `:1746-1747`) and C1 (scope the convergence check to pre-observed NULL rows, or downgrade to warn+retry), then merge-forward onto current `main` and re-run CI. C3/C4/C5 are follow-ups.

hachej:
**Review ladder record (weekend factory burn, cross-model pass: gpt-5.6-sol xhigh via codex; orchestrated by Fable session).**

**Verdict: REVISE — close, but two findings + rebase needed before owner queue.**

Confirmed the core ask: persisted `defaultAgentTypeId` naming a missing/excluded seat now **fails closed** (`defaultAgentType.ts:89-101` stable diagnostic + 409 `default_agent_type_unknown_seat`, admission denial at `createCoreWorkspaceAgentServer.ts:1609-1648`; honest tests). Findings:

1. **High — pre-schema health path misses the real upgrade failure.** `createCoreWorkspaceAgentServer.ts:1681-1695` suppresses only `42P01` (missing table); a pre-0024 DB where `workspaces` exists but the column doesn't yields `42703` and aborts startup before `/health` serves. Test only fabricates `42P01`.
2. **Medium — legacy NULL resolution ignores config-sourced application default.** Default resolved from `options.defaultAgentTypeId ?? rawConfig.defaultAgentTypeId` at `:1080-1097`, but workspace meta passes only the option at `:1744-1747` — config default can be bypassed for the first fleet seat.

Rebase state: **36 behind / 5 ahead** (no textual conflicts in a read-only merge test). Fix both, rebase, rerun CI.

hachej:
Closing as superseded by #1386, at the owner's direction.

#1386 states it directly: this PR rebased onto current main (it was 36 commits behind), plus fixes for both prior review findings — the 42703 pre-migration health path, and honoring the config-sourced default in legacy NULL resolution.

Verified before closing: both branches change **exactly the same 11 files**, and there is nothing touched here that #1386 does not also touch. #1386 is **CLEAN (14 green)**.

Reviewing both would be reviewing the same change twice, and landing both would guarantee a conflict. Reopen if you think a change here is missing from #1386.

PR #1324, opened 2026-08-17
[br-wt-391-forward-srj0] Make frozen-chat forks seamless

### LINKED ISSUE #1323: Sending a message is not smooth: no optimistic echo, chat visibly reloads before the message appears
## Symptom (owner, on the 0.1.101 RC)

Send a follow-up message → **nothing appears** → the chat visibly **reloads** → the sent message finally shows up. Every send feels like a page transition instead of a chat.

## Two suspected paths (verify both)

1. **General follow-up send**: the composer's optimistic echo either isn't rendered or is dropped and replaced by a full state refetch (`clear-optimistic-followups` exists in `remotePiSession.ts` — check when the optimistic entry is cleared vs when the server echo arrives; any gap = blank period). A send should append locally, instantly, and reconcile in place — never refetch-and-rerender the whole timeline.
2. **Fork-on-send (#1315 v1)**: typing into a frozen chat creates a NEW session and re-mounts the panel (`PiChatPanel` fork flow) — structurally a reload. Smoothing: keep the rendered history in place, swap the session binding underneath, append the new message optimistically; the fork should be invisible except for the runtime notice updating.

## Expected

- Sent message appears in the timeline the moment Enter is pressed (optimistic, marked pending).
- Streaming reply attaches under it without any timeline re-mount.
- Reconciliation with server state replaces the optimistic entry in place; no flicker, no scroll jump, no blank chat.
- Fork path: same, plus the session-id swap is invisible.

## Notes

- Related: #1295 established queue/optimistic handling around Stop; same machinery.
- This is core daily-feel: it multiplies with #1303 (bundle) since every re-mount re-runs heavy render paths.

hachej:
Closes #1323

Keeps the mounted transcript and stable timeline identity while a frozen chat forks. The prompt is shown optimistically at Enter, the remote session binding swaps beneath the pane, and server hydration reconciles the nonce in place before streaming continues.

Proof, independent review, and RC-fixture demo will be attached before owner validation.

### DIFF
--- a/packages/agent/src/front/chat/PiChatPanel.tsx
+++ b/packages/agent/src/front/chat/PiChatPanel.tsx
@@ -1227,7 +1227,7 @@ export function PiChatPanel<
   const disabled = !policy || sessionsLoading || composerBlocked
   const isStreaming = isPiBusyStatus(status)
   const submitStatus = initialHydrationPromptAllowed ? 'ready' : toPromptSubmitStatus(status)
-  const submitDisabled = !policy || sessionsLoading || modelSelectionBlocked || (composerBlocked && !isStreaming)
+  const submitDisabled = !policy || sessionsLoading || modelSelectionBlocked || Boolean(forkTransition?.forking) || (composerBlocked && !isStreaming)
   const mergedToolRenderers = useMemo(() => mergeShadcnToolRenderers(toolRenderers), [toolRenderers])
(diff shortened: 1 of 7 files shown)


hachej:
Closing per owner simplification ruling on #1330: the runtime-scope pin machinery this PR polishes is being removed entirely rather than smoothed (sessions will always continue, matching pi semantics — no trigger left for fork-on-send). The work here was sound; the premise underneath it was retired after review. See #1330.

Issue #1348, opened 2026-08-21
Pending owner gates are marked abandoned when the hub restarts — the review queue is wiped


hachej:
## Symptom (observed 2026-08-21, 0.1.101 anchor hub)

A hub restart marked **all 8 pending `ask_user` questions `abandoned`** in one shot: four starred strategic PR gates (#1211, #1220, #1164, #1145), the task-system Gate 1 r4, the issue-buckets purge gate, the workspaces-mode scope gate, and a PR #1288 gate. The owner's entire review queue vanished from the inbox; the store shows `abandoned` for every one.

## Why this is P1

A gate is a **durable decision request**, not a property of a live session. The whole factory model is "workers block, the owner decides when he decides" — that contract breaks the moment a restart (crash, upgrade, deploy) can silently void every pending decision. Cost so far: hours of worker time re-raising gates, and the owner repeatedly asking "why is my inbox empty?" (see also #1337, the projection defect — different mechanism, same felt symptom).

## Cause

Pending questions are tied to the asking agent session's liveness. When the process restarts, sessions do not resume, so the runtime transitions their questions `ready → abandoned` rather than leaving them answerable.

## Expected

1. A `ready` question **stays `ready`** across a hub restart. It is answerable from the durable store regardless of whether its asking session exists.
2. Answering a question whose asking session is gone must still record the decision durably (and, where the answer drives an action, be replayable by whoever picks it up).
3. `abandoned` should mean the *owner* dismissed it or the question was explicitly superseded — never "the process bounced".
4. Recovery for the existing 9 abandoned records: a way to restore/re-raise without hand-rebuilding each one.

## Related
#1337 (inbox reads streams instead of the store), #1295 (Stop deletes queued messages) — same family: durable state treated as session-scoped.

hachej:
Triage 2026-08-26: PR #1384 (restart-safe gates + decision record + store hardening) was deliberately closed — this capability ships inside the durable-stream feature instead. This issue stays open as the tracking bug until that feature lands; do not resurrect #1384.

Issue #1314, opened 2026-08-17
Store the request ledger outside the user workspace


hachej:
## Problem

`createWorkspaceAgentServer()` and `createStandaloneAgentHostApp()` hardcode the durable request ledger under the user workspace:

```text
<workspaceRoot>/.boring/agent-request-ledger.sqlite
```

The workspace Files API consequently exposes `.boring/` and the SQLite database, WAL and SHM files as read/write user files. Agents can discover and mutate them. `.gitignore` only prevents commits and does not provide a security boundary.

The request ledger is host application state, not workspace content. It coordinates request idempotency across all sessions and deployments.

## Requested change

- Add an explicit `requestLedgerPath?: string` host option to standalone and workspace server options.
- Prefer host-owned durable storage. When `sessionRoot` is supplied, default the ledger beside session state rather than under `workspaceRoot`.
- Preserve a documented compatibility fallback only when neither host path is supplied.
- Ensure workspace filesystem bindings never expose the ledger.
- Add tests proving a configured ledger is created outside the workspace and `.boring` is absent from workspace file/tree/search APIs.

## Healio target

Healio should configure the ledger under host state, e.g. `/data/agent-state/request-ledger.sqlite` in containers and `apps/analytics/bi-playground/.data/agent-state/request-ledger.sqlite` locally.

## Current source

- `packages/workspace/src/app/server/createWorkspaceAgentServer.ts`
- `packages/agent/src/server/createStandaloneAgentHostApp.ts`

hachej:
Fixed on main by PR #1358: all hosts route through resolveRequestLedgerPath, honoring host-owned BORING_AGENT_SESSION_ROOT (AGENTS.md rule 9) with legacy in-workspace fallbacks. PR #1404 follows up with cutover-hazard documentation (relocating the ledger strands legacy idempotency keys; copy/drain at cutover if exactly-once must survive the upgrade) and API cleanup.

Issue #1253, opened 2026-08-13
ui-review tooling leaks a mktemp -d per run (thousands of /tmp dirs, exhausts tmpfs inodes)


hachej:
Every `tools/ui-review` run creates temp directories (`boring-ui-review.*`, `boring-ui-review-{desktop,mobile}.*`, `ui-review-bounds.*`, `ui-review-priority.*`, `ui-review-trace.*`, `ui-candidate.*`, `ui-baseline.*`, `boring-ui-review-replay-*`) and does not remove them. Measured on the dev VM today: **1,825 leaked ui-review dirs** in `/tmp`.

The hung-Chromium path makes it worse — when cleanup is interrupted (observed in a real run: "Chromium cleanup subsequently hung and was interrupted"), the dirs are guaranteed to leak.

**Impact:** `/tmp` is tmpfs with a 1,048,576 inode cap. Accumulated leakage reached 100% inode usage and **killed three running factory workers** plus blocked the harness entirely (ENOSPC on every command, including tool bookkeeping). Cost ~1h of orchestration time today.

**Fix:** register cleanup on process exit (including SIGINT/SIGTERM and the Chromium-hang path) for every mktemp-d directory the tooling creates; or root them all under one run-scoped parent dir removed in a `finally`. Also honour `TMPDIR` so CI/dev can redirect off tmpfs.

Related: the deterministic-replay paint race (separate PR) and the /tmp aging rule (separate issue).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

hachej:
Fixed on main by PR #1358: run-scoped temp dirs removed on exit and signals. Pre-existing /tmp residue was manually purged 2026-08-26.

PR #1358, opened 2026-08-21
Overnight bugfix sweep: #601 #1028 #857 #1233 #1253 #1177 #1348 #1314 #1171 #1196

hachej:
## Overnight bugfix sweep — 2026-08-21

Six independent low-hanging fixes from tonight's backlog sweep, each with behavior tests and a proof run. Every commit stands alone and can be dropped without affecting the others.

| # | Fix | Proof |
| --- | --- | --- |
| #601 | Remote chat sessions no longer require workspace provisioning — new additive `remoteSessionsEnabled?: boolean` prop, default behavior unchanged | 92/92 workspace front tests, typecheck clean; new test proven to fail without the fix |
| #1028 | Remove unused `MessageTimeline` renderer (−297 lines) | typecheck clean; `test:chat-baseline --unit` 508 tests pass; zero remaining references |
| #857 | Dev servers: serialize shared-package builds (`build:app-deps`) before parallel app `dev:app` | two playgrounds booted concurrently; 528/528 `dist/` files byte-identical after |
| #1233 | Scaffolded plugins: agent and front templates now agree on the panel id (`.panel`) | new golden test scaffolds a real plugin via the built CLI; fails on un-fixed templates; 18/18 pass |
| #1253 | ui-review: run-scoped temp dirs removed on exit and on SIGINT/SIGTERM/SIGHUP | 81 tests pass incl. a spawned child killed by SIGTERM leaving no root; /tmp entry count unchanged across a full run |
| #1177 | Restore four `docs/visual/*.md` S2 pages that never reached main (stacked-PR merge trap, not a history rewrite) | diffstat matches source commit exactly; all 38 referenced paths resolve |

### Notes for review

- **#601 naming is open**: the new prop is `remoteSessionsEnabled?: boolean`; the issue also floated a `sessionMode` enum. Unreleased, so renaming is cheap.
- **#1028 found a second, pre-existing break**: `test:chat-baseline` has been silently dead on main since `292fa550c` renamed a test file without updating the script's hardcoded list. Fixed in the same commit — worth asking why CI never surfaced it.
- **#857 changes root `dev` semantics**: it no longer selects packages' watch-mode `dev` scripts. That path was already dead (`packages/core` `tsup --watch` never exits, so the apps were never reached), but anyone relying on package watchers from root `dev` must now invoke them directly.
- **#1253 residue**: ~203 leaked ui-review temp dirs already exist in `/tmp` (89% inodes). This prevents new ones; reclaiming the old ones is an operator action and was deliberately not done.
- **#883 and #1060, #1106, #1083, #1323** were closed separately with proof — they needed no code.


hachej:
## Added: #1348 — pending owner gates no longer wiped on hub restart

Commit `8b1d9c0` (6 files, +71/−33, all under `plugins/ask-user/src/server`).

**Root cause confirmed on both counts.** The Fastify lifecycle hook at `askUserServerPlugin.ts:37-38` swept every pending question through `abandonOrphanedPending()` on each boot, judging orphan-ness by `InProcessAskUserCoordinator`'s in-memory waiter map — which is empty by construction at boot, so one restart abandoned the entire queue. Separately, `submitAnswer()` converted an answer into an abandonment whenever the waiter was gone, surfacing as a 409.

**Fix:** boot sweep deleted (lifecycle now only starts the publisher and registers `onClose`); `abandonOrphanedPending(sessionIds[])` narrowed to `supersedeSessionPending(sessionId)` — same body, single session, reachable only from `ask()`, with the "never run this as a sweep" invariant carried in the name and doc comment; `submitAnswer()` always persists the answer plus transcript event and then best-effort resolves the waiter, so the dead `"abandoned"` return branch is gone. `cancelQuestion()` is untouched — cancel-without-waiter still marks abandoned, which matches "the owner dismissed it".

**Proof:** typecheck clean; `vitest run src/server` → 61 passed / 1 skipped. Stash check is clean-cut: with the three source files stashed, exactly the four #1348 tests fail and nothing else.

**Two notes.** A pre-existing flake in `askUserWorkflow.test.ts` (transcript ordering when an answer lands between the queued `created`/`ready` writes) failed once under load average ~100-180; it fails identically on un-fixed sources, so it is not from this change — worth its own bead. And issue expectation #4 (recovering the 9 already-`abandoned` records) is **not** covered: that needs a one-shot over the hub's `.boring/ask-user.json` plus your confirmation of which of the 9 are still wanted.

**Collisions:** none in source — neither #1325 nor #1343 touches `askUserRuntime.ts`, `askUserServerPlugin.ts`, or `questionsBridge.ts`. Both make additive edits to the same test files, so expect mechanical test-file merges only.

hachej:
## 🔴 Fable final review

Independent verification of all six commits: read every diff, re-ran the suites myself, re-did two stash-checks by reverting the source hunks, and tried to break the #1253 containment guard adversarially. One commit needs a fix before the owner tests; nothing needs pulling.

### Per-commit verdicts

**#601 `09cb18a95` (remoteSessionsEnabled) — SHIP-WITH-FIX (blocking, trivial)**
- **This commit is why Lint CI is red.** The new JSDoc (`WorkspaceAgentFront.tsx:341`) and the dev warning string (`:927`) both contain the literal `/api/v1/agent/pi-chat/*`, which `scripts/check-agenthost-cutover-matrix.mjs` flags as an unclassified Agent route literal. Reproduced locally: `invalid #1029 final route matrix: unclassified Agent route literals: packages/workspace/src/app/front/WorkspaceAgentFront.tsx`. Fix: reword both strings to not spell out the route (e.g. "the pi-chat agent API"), or classify the literal in the matrix. One-line-each change.
- Minor: the dev warning fires spuriously when the app passes `remoteSessionsEnabled={false}` together with `provisionWorkspace={false}` — the guard at `:923` doesn't check that `remoteSessionsEnabled` was actually omitted, so an explicit opt-out is told to "Pass remoteSessionsEnabled={true}". Add `remoteSessionsEnabled !== undefined` to the early return while touching the file.
- The fix itself is correct: both front coupling points (`remoteSessionHookEnabled`, `hydrateMessages`) go through the single `remoteSessionsResolved`, default behavior unchanged, no other `provisionWorkspace !== false` sites remain in the front. Stash-check re-verified by me: reverse-applied the source hunk, new test fails (1 failed / 91 skipped); restored, 92/92 pass. On naming: keep the boolean; it matches `hotReloadEnabled`, and a `sessionMode` enum is speculative surface.

**#1028 `7ca3ff86f` (delete MessageTimeline) — SHIP**
- Verified zero remaining references outside `docs/plans/archive/`; the kept helpers (`noticeStyles`, `bareToolRenderers`, `tool-call-group`) genuinely have live consumers incl. public `front/index.ts` exports — the worker was right to refuse the issue body's step 3.
- The scope addition (fixing the stale `activeSessionStorage.test.ts` entry in `run-chat-baseline.mjs`) is justified: same file, same bug class, and the requested proof gate literally could not run without it. `test:chat-baseline --unit`: 35 files / 508 tests green in my rerun. Worth a separate look at why CI never noticed the gate was dead since `292fa550c`.

**#857 `9a46889ca` (build:app-deps + dev:app) — SHIP**
- Root-cause analysis is right (`tsup clean:true` races); the fix is Tier-A only, no new tooling, docs match. I verified nothing in `.github/workflows/` or `scripts/` invokes the old `pnpm -r run dev` or app `dev` scripts, so the semantics change strands nobody.
- Owner notes for tomorrow: (1) `pnpm dev` now actually starts all three apps — including full-app, whose `dev:app` runs a Postgres migration first; if the DB is down that lane fails (others keep running). (2) First `pnpm dev` pays a ~13-min cold serialized build. (3) The 17-package filter list is hand-maintained and will drift when an app adds a dep; a future simplification is `--filter './apps/*^...'` and letting pnpm compute the closure — not needed now.

**#1233 `9b3f30571` (scaffold panel-id) — SHIP**
- Minimal and correct; `.panel` matches every first-party plugin. Stash-check re-verified by me: reverted the template to `.page`, golden test fails with exactly the expected assertion; restored, passes. Full plugin-cli suite 21/21 green on my rerun (the earlier `samplePlugin.test.ts` import failure was an unbuilt-dist artifact, not real).

**#1253 `a6c47260d` (ui-review temp root) — SHIP (two non-blocking hardening notes)**
- I attacked the containment guard: it refused `/tmp` itself, `/home/...`, `/tmp/boring-ui-review-run.x/../../home`, and traversal-composed paths; the symlink-escape test genuinely leaves the outside target intact; the SIGTERM test kills a real spawned child (exit 143) and the root is gone. The `rmSync` target can only ever be a path mkdtemp'd under `tmpdir()` with the owned prefix. It cannot escape.
- Hardening note 1: in `createUiReviewTempDir` (`tempRoot.ts:57-60`) the containment assert runs **after** `mkdtemp`, so a prefix containing `../` creates the directory outside the run root and then throws, leaking it (I demonstrated `/tmp/tmpleak-probe-*`). All 19 call sites are static literals today, so unreachable — but validate the joined path before `mkdtemp` next time someone touches this file.
- Hardening note 2: the signal handlers call `process.exit()` directly, which preempts any signal listeners registered after them (e.g. Playwright worker teardown, since the e2e spec imports this module). Acceptable; just know Ctrl+C in e2e may skip other teardown.
- The `UI_REVIEW_OUTPUT_DISCARDED` behavior change (bare runs without `UI_REVIEW_OUTPUT_DIR` now delete the report they produced, with a warning): justified — every `ui:review:*` package script sets the output dir, README documents it, and the alternative is re-leaking. Deliberately not reclaiming the ~203 pre-existing leaked dirs was the right call; that's an operator action (and worth doing soon — /tmp is at 89% inodes).
- Suite green: 81/81 incl. the 5 new tempRoot tests.

**#1177 `6a7953266` (restore S2 docs) — SHIP**
- I diffed all four pages against source commit `5a2374a9`: byte-identical, plus the four README index rows. The forensics (shallow-clone artifact, stacked-PR merge trap, no history rewrite) hold up. Docs-only, zero runtime risk. Side effect to know: the primary checkout was unshallowed (2868 commits) during investigation.

### My actual test results (this branch, this worktree)
- Root `pnpm run typecheck` (build:packages + all workspaces): **clean, exit 0**.
- `WorkspaceAgentFront.test.tsx`: **92/92**. Chat-baseline `--unit`: **35 files / 508 tests**. plugin-cli: **21/21**. ui-review tools: **81/81**.
- Earlier failed runs (vitest forks-pool startup timeouts, samplePlugin import error) were machine-load / unbuilt-dist artifacts — all disappeared on a clean sequential rerun. No pre-existing failures remain in the affected packages.
- CI at review time: **Lint red — caused by #601 as above**; Invariants/Action Pins/Dependency Age/Runtime Refactor pass; Typecheck/Unit/E2E/UI Review still pending.

### Overall
**Safe for the owner to test tomorrow after one small fix commit on #601** (reword the two route-literal strings; optionally tighten the spurious-warning guard in the same commit). Without it the branch runs fine but Lint stays red and the PR can't land. Nothing should be pulled from the sweep. Manual-test checklist worth hitting: `pnpm dev` from root (all three apps, full-app needs Postgres), a scaffolded plugin's `/open-<name>` command, and one interrupted (`Ctrl+C`) ui-review run to see `/tmp` stay clean.

hachej:
## Added: #1314, #1171, #1196 — and a correction to the #1196 diagnosis

Three more fixes landed on this branch since the last update.

**#1314 — request ledger path** (`9eda4ec`, 6 files, +259/−3). New shared `resolveRequestLedgerPath` used by both `createStandaloneAgentHostApp` and `createWorkspaceAgentServer`, each gaining a `requestLedgerPath?: string` option. Chain: explicit option → `sessionRoot` → `BORING_AGENT_SESSION_ROOT` → legacy `<workspaceRoot>/.boring/…`. One deliberate deviation from core: core's last step is `<workspaceRoot>/.agent-request-ledger.sqlite`, but the legacy `.boring/` path is kept last so today's default is genuinely unchanged — documented as a compatibility fallback. Tests boot the real host and assert the file on disk; 3 of 4 fail per call site without the change. The issue's other half (hardening filesystem bindings so the ledger can't be listed) is out of scope and stays open.

**#1171 — Reload agent affordance** (`762b273`, 3 files, +141/−4). Capability-gated header action in `AgentDetailsOverlay`, mirroring `PluginsOverlay`'s "Reload plugins" action end to end — same ghost `IconButton`, same `RefreshCw`/`animate-spin`, same placement, same generation-stamping. Hidden (not disabled) when the host passes no handler. Success copy carries the runtime-pin semantics: new chats use the updated instructions, already-open chats keep the composition they started with. Fail-without-fix: the dispatch and error tests fail when the source is stashed.

**#1196 — bad shared-skill symlink no longer 500s every agent route** (`00b40ab`, 5 files, +235/−9). **The filed diagnosis was half wrong and the fix is correspondingly larger.** A *dangling* symlink never reaches `packageResources.ts:317` — `enumerateExternalSkillFiles`'s `add()` already drops entries whose `realpath` fails, and the first repro built against that theory passed on unfixed code. The actual `PATH_SYMLINK_ESCAPE` originates at `piResourceDigest.ts:458` (`assertContainedWithoutSymlinks`), reached via `hashResolvedPiInventory` → `hashResourceCollection:238` → `digestPiResourceInputs` → `modeApps.ts:1156`: one symlinked entry aborts the whole digest. Fixing only the workspace layer would not have cleared the route, so both layers are fixed — per-entry admission with a `PackageResourceDiagnostic` in workspace, and a new `UnadmittedEntryPolicy = 'reject' | 'skip'` in the digest applied **only** to the two ambient inventory collections. Host-declared inputs still reject outright. `FORMAT_VERSION` v5→v6, which invalidates in-flight reload fences on deploy (one extra reload).

Security invariant: a skipped entry is framed by path and code only — never `realpath`'d, opened, or mounted. The framing keeps the reload fence honest, asserted explicitly (the digest still moves when the link disappears).

Stash-check evidence, fail-before / pass-after: CLI `workspacesModeAmbientSkills` `expected 403 to be 200` → 3 passed; workspace `packageResources` throws at `:317` → 19 passed; agent `createHarness` digest case `Pi resource symlinks are not allowed: …/.agents/skills/linked-skill` → 1 passed. Suites: workspace `src/server` 332 passed, agent `src/server` 1042 passed; typecheck green on workspace, agent and cli.

Note: Pi's ambient loader still surfaces the aliased skill from the user's own HOME — unchanged by design. What changed is that it no longer kills every route, and the test asserts that rather than pretending otherwise. This unblocks epic #1187's last open child.

hachej:
## 🔴 Fable final review — second pass (late commits)

Covers the four commits added after the first Fable pass, plus the small #601 follow-up at HEAD. All source read in full; worker reports verified rather than trusted (two stash-claims re-executed by me); targeted suites + typechecks re-run in the sweep worktree.

### Per-commit verdicts

**`031a6df` #1348 ask-user boot wipe — SHIP**
- Boot sweep deleted; repo-wide grep confirms no remaining caller of `abandonOrphanedPending` (only a stale untracked dist d.ts). `supersedeSessionPending(sessionId)` is reachable solely from `ask()`, with the invariant in the name and doc.
- `submitAnswer()` now persists answer + `answered` transcript event unconditionally, then best-effort resolves the waiter (`resolveAnswered` returns false on no waiter — no throw). `questionsBridge` 409 branch is genuinely dead code, correctly removed. `cancelQuestion` without a waiter still abandons, matching issue expectation #3.
- The inverted tests assert the *new* behavior, not just flipped expectations: answer persisted, transcript event present, hint republished into `hintsBySession`, restart runtime answerable. **Re-verified the stash claim myself**: with the three source files reverted to the parent commit, exactly the 4 #1348 tests fail (4 failed / 34 passed); with sources restored, 61 passed / 1 skipped.
- Not in this diff (worker flagged, agreed): the 9 already-`abandoned` records are not resurrected. Owner should not expect old gates to reappear — that needs a separate one-shot.

**`99f9526` #1314 request ledger path — SHIP** (one heads-up)
- Resolver precedence is correct and the legacy `<workspaceRoot>/.boring/agent-request-ledger.sqlite` fallback keeps the nothing-configured default byte-identical — the deliberate deviation from core's `sessionRoot ?? workspaceRoot` is right, since matching core would have silently moved the default.
- **Heads-up:** "default unchanged" is literally true only when neither `sessionRoot` nor `BORING_AGENT_SESSION_ROOT` is set. A deployed host that already sets the env var (the normal deployed configuration per AGENTS.md rule 9) will have its ledger relocate on upgrade — a one-time dedup-ledger reset, old file left behind in the workspace. That relocation is the point of the fix and matches core semantics, but it is a behavior change for env-configured hosts; expected during owner testing, not a bug.
- Tests boot real hosts and assert the SQLite file on disk; typecheck green (agent, workspace front+server).

**`190346457` #1171 Reload agent action — SHIP**
- Mirrors `PluginsOverlay`'s reload idiom exactly, including the spin-while-loading treatment (verified against `PluginsOverlay.tsx:197`). Hidden-not-disabled is the right call: nothing on this page can ever grant the capability.
- Generation-stamping verified in source: `load()` and the unmount cleanup both bump `generationRef`, so an agent switch or unmount mid-reload trips `generationRef.current !== generation` and drops both the toast and the refresh. Stale toast after an agent switch is genuinely prevented.
- Wiring targets the overlay's agent (not the addressed chat agent) and suppresses the shared helper's "Extensions reloaded." wording — correct. 42 front tests green (overlay + packageResources + ledger batch).

**`d2042e8` #1196 symlink skill degrade — SHIP** (highest-risk; actively tried to break it)
- **Scope of `'skip'` is exactly the two ambient collections**: `hashResolvedPiInventory`'s `resolved-package-manifest` / `resolved-resource` calls (piResourceDigest.ts:326-327). Settings, `additionalSkillPaths`, extensions, and local package sources (lines 150/154/159/179) keep the default `'reject'` — confirmed all call sites.
- **Only `PATH_ESCAPE`/`PATH_SYMLINK_ESCAPE` are skippable**; limit errors, changed-file errors, and IO errors still throw. No error-swallowing widening.
- **A skipped entry is never opened, hashed, or mounted**: in both `hashResourceCollection` and `hashLocalResource` the skip fires at the containment assertion, before `lstat`/`open`; the recursion propagates the policy; the registry layer skips before `resolveSkillRecord`, and the diagnostic carries only the id, no host paths. One nit: in the escaping-*directory* branch, `realpath` runs before the skip decision — metadata resolution only, nothing opened; not exploitable.
- **The fence still moves**: skips are framed into the hash (`${kind}-skipped code:path` / `unadmitted-skipped`), and inventory descriptors are framed independently. Test asserts the digest changes when the skipped link is removed. **Re-verified the fail-before claim myself**: with `piResourceDigest.ts` reverted to the parent commit, the new createHarness test fails; restored, 65 agent tests pass and the CLI end-to-end (`workspacesModeAmbientSkills`, 3 tests incl. the 403-before case) passes.
- **Known, accepted blind spot** (record it so nobody assumes otherwise): a skipped ambient symlink's *target* content changing, or the link retargeting, does not move the digest — while Pi's own ambient loader still surfaces that skill. The digest cannot hash outside authorized roots, and the pre-fix behavior was a hard 403 on every agent route, so this is the right trade — but the reload fence does not cover symlinked ambient skills' content.
- Registry layer: per-entry probes can't catch combination-level conflicts (e.g. duplicate shared ids), but the ambient enumerator dedupes ids and that failure class predates this change. `FORMAT_VERSION` v6 costs one extra reload per open session on deploy — no data impact.

**`b43706f` #601 follow-up — SHIP**
- Prose-only doc/warning change plus the `remoteSessionsEnabled !== undefined` guard (warning now only fires on the implicit coupling, as intended). Its motivating gate passes: `check:agenthost-cutover-matrix` → "zero forbidden compatibility references".

### Test evidence (this worktree, load avg ~4–9, no forks-pool artifacts)
- ask-user `src/server`: **61 passed / 1 skipped** (skip is the gated eval test)
- workspace targeted (packageResources + AgentDetailsOverlay + requestLedger): **42 passed**
- agent targeted (createHarness + standalone requestLedger): **65 passed**
- cli `workspacesModeAmbientSkills`: **3 passed**
- Typecheck exit 0: agent, workspace (front + server configs), cli, ask-user
- `check-agenthost-cutover-matrix`: pass
- Revert spot-checks: #1196 (1 failed on reverted source), #1348 (4 failed on reverted sources); tree restored clean after both.

### Overall recommendation
**Safe for the owner to test.** Nothing needs to be pulled from the sweep. Three things to know while testing:
1. If `BORING_AGENT_SESSION_ROOT` is set in the test environment, the request ledger relocates there (expected, #1314).
2. Expect one extra agent reload after deploy from the digest v5→v6 bump (#1196).
3. Previously-abandoned ask-user gates stay abandoned — #1348 prevents future wipes, it does not resurrect the 9 lost records.

hachej:

Scope: all 10 commits on `fix/bugfix-sweep-2026-08-21` (diff vs `main`), read against the surrounding source. Findings are ordered by structural weight and tagged **MUST-FIX-BEFORE-MERGE** or **FOLLOW-UP-BEAD** so tonight's merge stays unblocked.

### Overall decision

**APPROVE with 1 must-fix and 5 follow-up beads.** No commit crosses the 1000-line boundary, several commits fix bugs *by deleting complexity* (#1348, #1028 are exemplary), and nothing here is a structural regression that must block the owner's testing. But three commits (#1314, #1196, #857) each add a near-duplicate of a fallback/selection chain that already exists somewhere canonical — that is exactly how spaghetti compounds, and each has a clear judo move.

---

### Findings

#### 1. #1314 — the new resolver duplicates a chain `createAgentHost` already owns — FOLLOW-UP-BEAD

`packages/agent/src/server/agent-host/requestLedgerPath.ts` implements *explicit path → sessionRoot → legacy workspace path*. But both call sites hand `createAgentHost` **both** `sessionRoot` and the pre-resolved `requestLedgerPath` — and `createAgentHost` already implements the *explicit → sessionRoot* half itself, twice:

- `packages/agent/src/server/agent-host/createAgentHost.ts:290` — `options.requestLedgerPath ?? join(options.sessionRoot!, '.agent-request-ledger.sqlite')`
- `createAgentHost.ts:664-665` — same chain again for the mkdir/validation pass

And core has a **third** independent variant with its own env read and a *different* legacy location:

- `packages/core/src/app/server/createCoreWorkspaceAgentServer.ts:1079-1080` (`sessionRoot ?? BORING_AGENT_SESSION_ROOT`) and `:1590` (`join(sessionRoot ?? workspaceRoot, '.agent-request-ledger.sqlite')` — no `.boring/`).

So after this PR the repo has **four** encodings of "where does the ledger live" (core inline, createAgentHost ×2, new resolver), differing only in env handling and the legacy tail. The `.boring/` deviation itself is justified — it preserves each host's historical on-disk ledger — but it belongs in *the* canonical resolver, not in a fourth chain.

**Remedy (judo):** make `resolveRequestLedgerPath` the single owner. Give it the legacy fallback as an input (`legacyPath`), have `createCoreWorkspaceAgentServer` call it (passing its own legacy path), and delete the inline chains in `createAgentHost` by requiring callers to pass a resolved path (or having `createAgentHost` itself call the resolver). One chain, N call sites. The fix as shipped is behaviorally right and well-documented; the consolidation is the bead.

#### 2. #1196 — per-entry admission by probe-resolution, with a swallow-all catch — **MUST-FIX-BEFORE-MERGE** (small) + FOLLOW-UP-BEAD

`packages/workspace/src/server/plugins/packageResources.ts:502-521`: each shared skill is validated by running the *entire* resolver on a singleton input (`resolveWorkspacePackageResources([], { sharedSkillPaths: [shared] })`), then survivors are resolved *again* in the real call. Two problems:

- **The bare `catch {}` swallows every error class.** Any genuine resolver defect — not just an inadmissible symlink — now degrades to `"shared skill … was not admissible and was skipped"`, with the underlying error discarded. This is the exact failure mode the bug was about (silent 500 → now silent skip), and debugging it later will mean reproducing by hand.
  **MUST-FIX (cheap, tonight):** carry the underlying error code/message into the diagnostic, and rethrow errors that are not admission-shaped (path-escape / symlink / not-found), mirroring `SKIPPABLE_RESOURCE_CODES` from `piResourceDigest.ts`.
- **N+1 sequential probe-resolutions.** This mirrors the pre-existing scanned-record probe loop directly above (`:487-500`), so it is pattern-consistent — but the pattern is the debt. **FOLLOW-UP-BEAD (judo):** have `resolveWorkspacePackageResources` return per-entry accept/reject outcomes in one pass and delete *both* probe loops.

#### 3. #1196 — `UnadmittedEntryPolicy` threading in `piResourceDigest.ts` — FOLLOW-UP-BEAD

The policy param is a defensible boundary (host-declared inputs fail closed, Pi-discovered ambient entries degrade — that is a real semantic distinction, not a mode hack), and sharing the walker beats duplicating the recursion. But the implementation scatters the decision across **three** inline skip sites (`piResourceDigest.ts` ~`:239-268`, `:396-404` symlink branch, directory-escape branch) with **two different frame labels and path kinds** (`` `${kind}-skipped` `` + `absolutePath` vs `'unadmitted-skipped'` + `logicalPath`).

**Remedy:** collapse admission into one guard — `admitPiEntry(state, absolutePath, logicalPath, policy): 'admitted' | <skip-code>` — so the walker has a single decision point and a single frame shape. Also drop the `= 'reject'` default and make every call site state its policy: at a security boundary, explicit beats defaulted (the fail-closed default is the right instinct, but there are only ~5 call sites).

#### 4. #857 — hand-maintained 17-package filter list will drift — FOLLOW-UP-BEAD

`package.json` `build:app-deps` hand-lists 17 `--filter` flags with `--workspace-concurrency=1`, while `apps/full-app` and `apps/agent-playground` keep their own partially-overlapping `build:deps` chains (full-app's 9-package chain, agent-playground's 3 with a `build:dev` variant the root list doesn't know about). Three lists, one dependency graph, guaranteed drift.

**Remedy (judo):** let pnpm compute the set and the order — `pnpm -r --filter './apps/*...' --filter '!./apps/*' run build` selects exactly the dependencies of the apps, topologically ordered, which also deletes the `--workspace-concurrency=1` serialization hack (pnpm parallelizes where the graph allows, so dev startup gets *faster*). The per-app `build:deps` chains should collapse onto the same selector scoped to that app.

#### 5. #601 — dev-warning trio inline in a 2826-line component — FOLLOW-UP-BEAD

`packages/workspace/src/app/front/WorkspaceAgentFront.tsx:913-931`: the prop + resolved value + one-shot warning effect is 20 lines of ref/effect plumbing in a file that is already 2.8× the 1k cap. Two concrete defects inside it:

- The guard chain is redundant: given `provisionWorkspace === false` and `remoteSessionsEnabled === undefined`, `remoteSessionsResolved` is `false` by construction, so the `remoteSessionsResolved` guard and the `remoteSessionsEnabled !== undefined` guard can never independently fire. Three conditions express what one plus the derivation already implies.
- The effect reads `remoteSessionsEnabled` but omits it from the dep array (`exhaustive-deps` violation; benign for a one-shot dev warning, but it's exactly the kind of latent staleness this file collects).

**Remedy:** extract `useRemoteSessionsResolution({ remoteSessionsEnabled, provisionWorkspace, shouldUseRemoteSessions })` into its own module owning the resolution *and* the warning; the component keeps one line. The consolidation of the two former `provisionWorkspace !== false` sites onto `remoteSessionsResolved` (`:913`, `:2110`) is good — it made the invariant single-sourced.

#### 6. #1253 — tempRoot: identity wrapper + signal handlers installed in processes it doesn't own — FOLLOW-UP-BEAD

`tools/ui-review/src/core/tempRoot.ts` is a reasonable ~40-line concept carrying ~48 lines of ceremony:

- `cleanupUiReviewTempRoot()` (`:76-78`) is an async identity wrapper over `cleanupUiReviewTempRootSync()`. Delete it; call the sync one from `finally`.
- `e2e/review.spec.ts` importing `createUiReviewTempDir` means the module's SIGINT/SIGTERM/SIGHUP handlers — which call `process.exit()` — get installed **inside Playwright worker processes** on first temp-dir creation. That module side effect seizes signal handling in a test runner that has its own teardown; a Ctrl-C during a run can now preempt Playwright's graceful worker shutdown. Scope handler installation to the CLI entrypoints (`run-ui-review.mjs`, `explore-review-spec.ts`) via an explicit `armUiReviewTempCleanup()` call, and let library/test consumers get only the shared root + `finally` cleanup.
- `assertWithinUiReviewTempRoot(root, await mkdtemp(resolve(root, prefix)))` (`:60`) asserts a property `mkdtemp` guarantees by construction — dead ceremony; keep the assert for externally-supplied paths only.

The core fix (one run-scoped parent, ownership-checked recursive removal, `UI_REVIEW_KEEP_TMP`) is the right shape and the discarded-output warning in `run-ui-review.mjs` is a thoughtful touch.

#### 7. #1171 — Reload agent action — PASS (one nit)

Capability-gated optional callback (absent ⇒ control not rendered), generation-guard reuse for the stale-response race, and the host-vs-chrome ownership comment are all clean. Nit: `onReloadAgent?: (agentTypeId: string) => Promise<string | undefined | void>` — the `undefined | void` union plus the `|| undefined` coercion at the call site is contract mush; make it `Promise<string | undefined>` and have void-shaped hosts return `undefined`.

---

### Per-commit verdicts

| Commit | Verdict |
| --- | --- |
| #601 remote sessions prop (2 commits) | PASS-WITH-DEBT-NOTED (finding 5) |
| #1028 delete MessageTimeline | **PASS** — pure deletion, zero dangling refs, even fixes a stale baseline entry |
| #857 build:app-deps | PASS-WITH-DEBT-NOTED (finding 4) |
| #1233 scaffold panel id + golden test | **PASS** — the golden test checks both ends string-for-string; exactly right |
| #1253 ui-review temp root | PASS-WITH-DEBT-NOTED (finding 6) |
| #1177 docs restore | **PASS** (docs only) |
| #1348 ask-user gates | **PASS** — the model fix: deletes the boot sweep, narrows `submitAnswer` to `"answered"`, and a compensating branch in `questionsBridge` disappears with it. This is what a thermo-clean bugfix looks like. |
| #1314 requestLedgerPath | PASS-WITH-DEBT-NOTED (finding 1) |
| #1171 Reload agent | **PASS** (nit in finding 7) |
| #1196 shared-skill admission | PASS-WITH-DEBT-NOTED + one **MUST-FIX** (findings 2–3) |

### MUST-FIX-BEFORE-MERGE

1. `packageResources.ts:502-521` — narrow the bare `catch {}` to admission-shaped errors and include the underlying error code in the skip diagnostic. Anything else rethrows. (~10 lines; this is the bug's own failure path and must not become silently undiagnosable.)

### Follow-up beads to open

1. Consolidate the four ledger-path chains onto `resolveRequestLedgerPath` (finding 1).
2. Per-entry outcomes from `resolveWorkspacePackageResources`; delete both probe loops (finding 2).
3. Single `admitPiEntry` guard + required policy param in `piResourceDigest.ts` (finding 3).
4. Replace the 17-name filter list with pnpm dependency selectors; collapse per-app `build:deps` (finding 4).
5. Extract `useRemoteSessionsResolution` from `WorkspaceAgentFront`; tempRoot wrapper/signal-scoping cleanup rides along (findings 5–6).

### File-size check (rule 1)

No file crosses 1000 lines because of this PR. Files already over the cap grew slightly:

| File | Before | After |
| --- | --- | --- |
| `packages/workspace/src/app/front/WorkspaceAgentFront.tsx` | **2793** | **2826** (+33; already 2.8× the cap — net growth here should require an extraction bead) |
| `packages/workspace/src/app/server/createWorkspaceAgentServer.ts` | 2177 | 2188 |
| `packages/agent/src/server/piResourceDigest.ts` | 510 | 555 |
| `packages/workspace/src/server/plugins/packageResources.ts` | 505 | 524 |
| `packages/workspace/src/front/chrome/agents/AgentDetailsOverlay.tsx` | 335 | 399 |
| `packages/agent/src/server/createStandaloneAgentHostApp.ts` | 371 | 381 |
| `plugins/ask-user/src/server/askUserRuntime.ts` | 280 | 289 |
| `plugins/ask-user/src/server/questionsBridge.ts` | 148 | 143 (−5) |
| `tools/ui-review/src/core/tempRoot.ts` | new | 88 |
| `packages/agent/src/server/agent-host/requestLedgerPath.ts` | new | 47 |
| `packages/agent/src/front/chat/components/MessageTimeline.tsx` | 195 | **deleted** |

*Review only — no approve/request-changes state set, no code touched.*

hachej:
## MUST-FIX closure + fresh thermo

Updated HEAD: `6f6b6c7eb` (includes latest `main`).

The prior `MUST-FIX-BEFORE-MERGE` was not fully closed: lower filesystem catches still laundered `EACCES`/unexpected IO failures into `PACKAGE_RESOURCE_INVALID`. Fixed every admission boundary to degrade only expected missing/symlink verdicts, rethrow unexpected failures, and preserve both code and message in diagnostics. Added an `ENAMETOOLONG` regression at the exact declared-skill `stat` boundary.

Fresh T1 thermo also found and fixed:
- resolved-inventory digest skip policy covered configured resources; now only Pi `source === 'auto'` ambient discovery may degrade, configured/package/explicit resources remain fail-closed (`FORMAT_VERSION` v7)
- Agent details could pair the overlay Agent with another Agent's active session; details reload is now agent-scoped and omits `sessionId`, with fleet integration coverage
- direct `createAgentHost` bypassed canonical blank/padded ledger-path normalization

Final Sol xhigh thermo verdict after fixes: **CLEAN**.

Local proof after fixes / main merge:
- workspace package-resource tests: **22/22**
- agent targeted suites: **83/83**
- addressed fleet reload integration: **pass**
- workspace + agent typechecks: **clean**
- AgentHost cutover matrix: **pass**
- playground: `http://localhost:5175` (PR worktree, HTTP 200)

PR #1359, opened 2026-08-21
#1123 chore(bash): delete duplicated buildBwrapArgs, use canonical sandbox implementation (extracted from #1166)

hachej:
## What

Deletes `packages/boring-bash/src/agent/runtime/buildBwrapArgs.ts` — a 141-line **byte-identical** duplicate of the canonical `packages/boring-sandbox/src/providers/bwrap/buildBwrapArgs.ts`.

Purely subtractive. No behavior change. **No mount-contract code.**

| | |
|---|---|
| Files | 3 |
| Diff | +23 / −151 |

## Why this is a separate PR

This deletion was one change inside **#1166** ("#1123 slice 1 — environment mount contract + provider substrate"). Tonight's review of #1166 concluded:

> This is the textbook shape of speculative plumbing: a contract, a capability flag, a feature flag, a fail-closed guard, a validator, a resolver, an error-code pair, a doc amendment, and two hand-synced duplicate interfaces — with **zero** production callers […]
>
> **The one exception is genuinely valuable and stands on its own:** deleting `packages/boring-bash/src/agent/runtime/buildBwrapArgs.ts` […] removes a duplicated **security-critical argument builder**. […] Clean, zero-risk deletion. It has a consumer today, unlike everything else in this PR.

#1166 additionally carries a standing **owner request-changes** (2026-08-19) asking for the mount substrate to be re-derived in light of #1317, and is 136 commits behind main with material semantic conflicts. So the dedupe was being held hostage by ~900 lines of unreachable substrate. This PR extracts it, cut fresh from current `main`.

#1166 is untouched and stays open.

## Independent re-verification on current main

Not taken on trust from the review — re-checked against `origin/main` in a fresh worktree:

1. **Still byte-identical.**
   ```
   $ diff packages/boring-bash/src/agent/runtime/buildBwrapArgs.ts \
          packages/boring-sandbox/src/providers/bwrap/buildBwrapArgs.ts
   $ echo $?
   0
   $ wc -l <both>
   141 / 141
   ```
   No divergence since the review.

2. **The deleted value exports have zero importers**, repo-wide (`*.ts/tsx/js/mjs/json/md`, excluding `node_modules` and `dist`), including dynamic imports and public barrels:
   `buildBwrapArgs`, `BWRAP_TIMEOUT_SECONDS`, `KILL_GRACE_SECONDS`, `RO_BIND_DIRS`, `RO_BIND_TRY_DIRS`.
   The two real consumers — `packages/agent/host/sandbox.ts` and `packages/agent/test-host/sandbox.ts` — already import from `@hachej/boring-sandbox/providers/bwrap`.
   `packages/boring-bash/src/__tests__/dualTargetParity.contract.test.ts` imports the `@hachej/boring-bash/agent` barrel but uses none of these symbols.

## The one piece of rewiring

The only in-repo importer of the deleted module was a **type-only** import in `packages/boring-bash/src/agent/runtime/types.ts`, used by the host injection point:

```ts
export interface RuntimeHostOperations {
  buildBwrapArgs(workspaceRoot: string, options?: BwrapArgsOptions): string[]
  ...
}
```

### Correction: boring-bash is *not* forbidden from importing boring-sandbox types

The first commit on this branch copied `BwrapArgsOptions` verbatim into `types.ts`, justified by "boring-bash must not import boring-sandbox (enforced by `check:invariants`)". **That justification was factually wrong**, and the follow-up commit fixes it.

`packages/boring-bash/scripts/check-invariants.mjs` filters on `!typeOnly`:

```js
violations: parsed.references.filter(({ specifier, typeOnly }) =>
  !typeOnly && isForbiddenAgentRuntimeSpecifier(specifier)),
```

It forbids only **value** imports. Its own negative fixtures explicitly assert that `import type { Sandbox } from '@hachej/boring-sandbox/shared'` is a **non-violation**, and `types.ts` already type-imports from `@hachej/boring-agent/shared` a few lines above where the copy landed. A type-only import would have been legal all along; the 22-line mirror was never required.

### What it does instead: narrow the contract rather than mirror it

Swapping the mirror for a type-only import would still add a package edge for one interface. The better move: boring-bash's **only** call site, `tools/harness/bashToolOptions.ts`, populates exactly **1 of the 6** fields — `readonlyPaths`. The mirror typed an object that only ever has one key.

So the injected contract is narrowed to what boring-bash actually supplies:

```ts
buildBwrapArgs(
  workspaceRoot: string,
  options?: { readonly readonlyPaths?: readonly string[] },
): string[]
```

matching the canonical field's exact optionality and readonly-ness. The copied interface and its barrel re-export are **deleted**.

`AgentRuntimeHostOperations` (in `packages/agent`, which legally depends on boring-sandbox) re-declares the member with the real `BwrapArgsOptions`, so the injection site in `packages/agent/host/sandbox.ts` keeps full typing. Method bivariance accepts the widening **with no casts** — both packages typecheck clean, and a probe confirms the narrowed signature still rejects `extraArgs` at boring-bash call sites, so the narrowing is real rather than silently widened.

Net effect: **no mirrored interface, no hand-sync hazard**, and `types.ts` goes from +22/−1 to +10/−2.

## Public API surface: this removal *is* externally breaking

The original body said the deleted exports have "zero importers repo-wide". That is true **in-repo**, but incomplete, and worth stating plainly:

`@hachej/boring-bash` is **not private** and **is published to npm** (latest `0.1.103`), and `./agent` is a **public subpath export** in `package.json`. The published `0.1.103` artifact does contain all five removed value exports in `dist/agent/index.js` / `dist/agent/index.d.ts`:

```
dist/agent/index.d.ts:3: export { B as BWRAP_TIMEOUT_SECONDS, b as BwrapArgsOptions, K as KILL_GRACE_SECONDS, ... }
dist/agent/index.js:1741-1750: BWRAP_TIMEOUT_SECONDS, KILL_GRACE_SECONDS, RO_BIND_DIRS, RO_BIND_TRY_DIRS, buildBwrapArgs
```

So an external consumer importing `buildBwrapArgs`, `BWRAP_TIMEOUT_SECONDS`, `KILL_GRACE_SECONDS`, `RO_BIND_DIRS`, or `RO_BIND_TRY_DIRS` from `@hachej/boring-bash/agent` **would break**. The follow-up commit additionally drops the `BwrapArgsOptions` **type** export from that barrel.

"No behavior change" holds for every in-repo consumer and for the runtime behavior of the builder itself (the canonical implementation is byte-identical). It does **not** mean the published API surface is unchanged. On a `0.x` line this is a permitted minor-version break, but it should be an explicit, reviewed decision rather than an unnoticed side effect — external consumers should import from `@hachej/boring-sandbox/providers/bwrap`, which is where the implementation has always lived.

## Proof

All run in the branch worktree.

| Command | Result |
|---|---|
| `pnpm --filter @hachej/boring-bash run build` | ✅ Build success |
| `pnpm --filter @hachej/boring-bash run typecheck` | ✅ clean (`tsc --noEmit`, 0 errors) |
| `pnpm --filter @hachej/boring-bash run check:invariants` | ✅ all invariants PASS, incl. the no-boring-sandbox-value-imports scan over 530 files |
| `pnpm --filter @hachej/boring-bash run test` | ✅ **17 files / 86 tests passed** |
| `pnpm --filter @hachej/boring-agent run typecheck` | ✅ clean — the downstream barrel consumer, incl. the re-declared `AgentRuntimeHostOperations.buildBwrapArgs` |
| `pnpm --filter @hachej/boring-sandbox exec vitest run src/providers/bwrap` | ✅ **3 files / 41 tests passed** — canonical builder + snapshot unchanged |

### One pre-existing failure, disclosed

The full `@hachej/boring-sandbox` suite reports **2 failed / 574 passed**, both in
`src/providers/blaxel/__tests__/blaxelProvisioning.test.ts` ("resumes an owned transaction after interruption during top-level publication").

**This is pre-existing on `origin/main` and unrelated to this change** — this PR does not touch `packages/boring-sandbox` at all. Confirmed by `git stash`-ing the change and re-running that file on the pristine tree: same 2 failures.

## Follow-up (not in this PR)

The mount substrate from #1166 stays on hold pending the #1317 stress-test the owner asked for. Per the review it must return rebased onto main's `upload`-era catalog, with the `stat` TOCTOU wrapped, the `globalToolMountArgs` init-cache resolved, and — critically — **bundled with slice 2**, so `context.mounts` finally has a producer.


hachej:

**Verdict: PASS-WITH-DEBT-NOTED** — conditional on §1, which is a ~5-line change. Everything else here is clean and I would not re-litigate it.

The core of this PR is correct and I verified it independently rather than taking the description on trust: the five deleted value exports (`buildBwrapArgs`, `BWRAP_TIMEOUT_SECONDS`, `KILL_GRACE_SECONDS`, `RO_BIND_DIRS`, `RO_BIND_TRY_DIRS`) have zero importers repo-wide outside the deleted file itself and boring-sandbox's own canonical copy. Deleting 141 lines of duplicated **security-critical** argument builder is exactly the right move, and splitting it out of #1166 was the right call. `.agents/skills/present-pr/SKILL.md:99` still records this duplication as accepted debt ("Two `buildBwrapArgs` copies kept in sync rather than deduplicated; accepted with rationale") — that line should be retired in this PR, since the debt it names is now paid.

But the PR does not finish the job. It deletes the duplicated *code* and replaces it with duplicated *types* across the same boundary — and the stated reason for doing so is not correct.

---

### 1. MUST-FIX-BEFORE-MERGE — the justification for the type copy is false, and the judo move deletes the type entirely

`packages/boring-bash/src/agent/runtime/types.ts:8-15` states:

> *"boring-bash never imports the provider package, so hosts inject the builder through `RuntimeHostOperations`. This interface is the structural mirror of that builder's options and must stay in sync with it."*

and the PR body says boring-bash **"must not"** import boring-sandbox, "enforced by `check:invariants`".

That is not what the invariant enforces. `packages/boring-bash/scripts/check-invariants.mjs:150-157`:

```js
violations: parsed.references.filter(({ specifier, typeOnly }) =>
  !typeOnly && isForbiddenAgentRuntimeSpecifier(specifier)),
```

Type-only imports are **explicitly exempt**, and the script's own negative fixtures at `:187-189` assert precisely this case as a non-violation:

```js
"import type { Sandbox } from '@hachej/boring-sandbox/shared'",
"import { type ProviderCapabilities } from '@hachej/boring-sandbox/shared'",
"export { type SandboxProvider } from '@hachej/boring-sandbox/providers'",
```

So `import type { BwrapArgsOptions } from '@hachej/boring-sandbox/providers/bwrap'` would pass `check:invariants` today. The architecture deliberately carves out exactly this case: it forbids **value** coupling (which is what would drag the provider into boring-bash's runtime graph), not type coupling. Baking a false constraint into a docstring is worse than the duplication itself, because the next person to touch this will believe it.

boring-bash already does this exact thing elsewhere: `types.ts:1-5` type-imports `FileSearch`, `Sandbox`, `Workspace` from `@hachej/boring-agent/shared` rather than mirroring them. The precedent for the correct move is four lines above the copy.

**But there's a better answer than the type-only import, and it's the one I'd push for.** Look at what actually crosses this boundary. The only call site in boring-bash is `packages/boring-bash/src/agent/tools/harness/bashToolOptions.ts:26-28`:

```ts
const args = runtimeHost.buildBwrapArgs(workspaceRoot, {
  readonlyPaths: bundle.readonlyWorkspacePaths,
})
```

**One field.** boring-bash never populates `extraArgs`, `postWorkspaceArgs`, `network`, `newSession`, or `dropAllCapabilities` — five of the six fields in the interface it just copied. (The only other injected-path consumer, `packages/agent/src/server/runtime/modes/provisioningAdapter.ts:250-256`, populates `extraArgs` only — and that file lives in `packages/agent`, which imports boring-sandbox legally and can use the canonical type directly.)

So the whole 22-line hand-synced mirror exists to type an argument object that only ever has one key. Narrow the injection point to the contract that is actually crossed and the duplication evaporates:

```ts
// packages/boring-bash/src/agent/runtime/types.ts
export interface RuntimeHostOperations {
  /** Canonical impl: `@hachej/boring-sandbox/providers/bwrap#buildBwrapArgs`. */
  buildBwrapArgs(
    workspaceRoot: string,
    options?: { readonlyPaths?: readonly string[] },
  ): string[]
  withWorkspacePythonEnv(input: WorkspacePythonEnvOptions): Record<string, string | undefined>
}
```

`AgentRuntimeHostOperations` (`packages/agent/src/server/runtime/runtimeHost.ts:18`) can then re-declare `buildBwrapArgs` with the real `BwrapArgsOptions` imported from boring-sandbox for its own `extraArgs` call site — method-shorthand params are bivariant, so the extension is legal, and `packages/agent` is the layer that legitimately owns that dependency.

Net: −22 lines instead of +22, no mirrored type, no "must stay in sync" hazard, and the boundary now documents *exactly* what boring-bash is allowed to ask for instead of leaking the provider's full option surface into a package that is forbidden from knowing about the provider. That is the structural point — the duplication isn't the disease, the over-wide injected contract is. Right now `RuntimeHostOperations` advertises five capabilities to boring-bash that boring-bash cannot and does not use.

If for some reason the full option surface really is wanted at this boundary, then take the type-only import (permitted, precedented) plus a `@hachej/boring-sandbox` **devDependency** in `packages/boring-bash/package.json` — the package lists it nowhere today, and `import type` is fully erased by tsup so this adds zero runtime coupling. Either way, do not ship the copy.

---

### 2. MUST-FIX-BEFORE-MERGE (only if §1's copy is kept) — nothing enforces the sync the docstring demands

If the mirrored interface survives, note that **no test, invariant, or type assertion checks that the two definitions agree.** I looked: `packages/boring-bash/src/__tests__/dualTargetParity.contract.test.ts` is a runtime behavioural suite with no type-level assertion and no `buildBwrapArgs` reference; the only test doubles (`packages/agent/src/server/__tests__/createStandaloneAgentHostApp.test.ts:114-117`) are `vi.fn(() => [])`, which assert host precedence, not shape.

The one place that *does* provide partial coverage is the injection site, `packages/agent/host/sandbox.ts:61-79`, where the canonical `buildBwrapArgs` is assigned into `agentSandboxRuntimeHostOperations: AgentRuntimeHostOperations`. That catches gross signature breakage — but it does **not** catch the drift that actually matters: because these are all-optional properties and method params are bivariant, the canonical builder silently dropping or renaming `readonlyPaths` would still typecheck at the injection site, while boring-bash keeps happily passing a field the builder ignores.

`readonlyPaths` is the field that re-binds protected workspace paths readonly inside the sandbox so spawned shells cannot mutate what the Operations layer refuses to mutate. Silent loss of that is a security regression that compiles clean. "Must stay in sync" with nothing enforcing it, on a security-relevant field, is not an acceptable resting state. §1's narrowing removes this problem rather than needing a test to police it — which is why §1 is the preferred remedy.

---

### 3. FOLLOW-UP-BEAD — `runtimeHost?` optionality and the duplicated guard

`packages/boring-bash/src/agent/runtime/types.ts:77` declares `runtimeHost?: RuntimeHostOperations`, and `bashToolOptions.ts:22-23` / `:51-52` each open with the same guard under different message strings:

```ts
const runtimeHost = bundle.runtimeHost
if (!runtimeHost) throw new Error('local sandbox runtime requires injected host operations')
...
if (!runtimeHost) throw new Error('direct runtime requires injected host operations')
```

Both bash strategies that reach a spawn hook require it unconditionally, so the optionality buys nothing except two runtime throws for a condition the type system could have made unrepresentable (rule 5). Either make `runtimeHost` required on the `RuntimeBundle` variants whose `bash` strategy needs it, or extract a single `requireRuntimeHost(bundle, strategy)` helper. Pre-existing, not caused by this PR — but this PR is editing the interface that declares it, so it's the natural moment to file it.

---

### 4. FOLLOW-UP-BEAD — this is a breaking change to the published package surface, and the description says otherwise

The PR body calls this "purely subtractive. No behavior change." That is true **in-repo**, and I verified it. It is not true of the published artifact: `@hachej/boring-bash/agent` is a real export (`package.json:28-31`) and this removes five value exports from it. At `0.1.x` that's low-stakes, but per the honesty rule it should be stated plainly in the description rather than covered by "no behavior change," and it should carry whatever changeset/version note the repo uses for a removed export.

---

### Checks that came back clean

- **File size (rule 1):** `types.ts` 116 lines, `agent/index.ts` 87 lines. Nowhere near 1k, and the +22 does not push anything toward a decomposition threshold.
- **Spaghetti (rule 2):** no new branches, no conditionals bolted into existing flows. The diff is a delete plus a type move.
- **Zero-importer claim:** independently confirmed by grep across `origin/main` for all five deleted value exports.
- **Byte-identity claim:** confirmed — the moved interface body is byte-identical to `packages/boring-sandbox/src/providers/bwrap/buildBwrapArgs.ts:55-67`.
- **Deletion of the duplicate itself:** unreservedly correct, and the right thing to have extracted from #1166.

Fix §1 and this is a clean, genuinely valuable subtraction. Please also drop the now-stale `present-pr/SKILL.md:99` debt entry.

PR #1256, opened 2026-08-13
[Sandbox Alias Cleanup] Centralize sandbox source resolution

hachej:
## Owner Review

**Bead / PR / issue:** `wt-391-forward-o56o.3` · [PR #1256](https://github.com/hachej/boring-ui/pull/1256) · [issue #1240](https://github.com/hachej/boring-ui/issues/1240)  
**Historical lineage (read-only):** `wt-391-forward-pr-1256-r2-research-reconcile-e9t9`

**What changed / why:** The repaired PR keeps only the sandbox alias/source-resolution cleanup. `@hachej/boring-sandbox` now owns `boring-source` targets for its nine declared public subpaths; TypeScript consumes those exports and Vitest/Vite consumers share one exact-match helper instead of maintaining repeated private path tables. The rejected runtime descriptor registry, host-policy coupling, runtime-mode widening, and remote-worker behavior changes are absent.

**Why you are needed (matched protected boundary):** **PROTECTED** — `packages/boring-sandbox/package.json` changes supported package export conditions and therefore public package resolution. Package production churn is **151 additions + deletions** (`+50/-101` across 12 production files under `packages/`; the focused `sourceResolution.test.ts` is excluded from the numerical count). This is below the >500 size trigger, but the package export contract independently requires owner judgment, and the active rollout preserves owner merge handoff.

**Recommendation / alternatives:** Approve the single exact-revision merge card for artifact head `7e0c5a5ae678c34980b8e4b3edce52e2966292cd`, whose only change after reviewed code SHA `67f17095ec0ff068c86ed247dd2d5d627a610aab` is this owner-facing show-me document. Alternatives: request a narrower export list, defer, or reject and retain the duplicated consumer aliases.

**Scope / base / head SHA:** Base/current main `68dcb7db8822f721c6b45d0731e01a46fa364f28`; independently reviewed code head `67f17095ec0ff068c86ed247dd2d5d627a610aab`; docs-only artifact/PR head `7e0c5a5ae678c34980b8e4b3edce52e2966292cd`. The reviewed SHA is an ancestor of the artifact head; `67f17095e..7e0c5a5ae` adds only `docs/issues/1240/show-me-67f17095e.md`.

**What approval authorizes:** Merge exact PR head `7e0c5a5ae678c34980b8e4b3edce52e2966292cd` into `main`. It does not authorize deployment, package publication/release, further source changes, branch deletion, or a later revision.

[...]

Issue #1240, opened 2026-08-12
Provider registry: adding a sandbox provider should touch boring-sandbox (+agent) only


hachej:
Owner directive 2026-08-12. Adding a provider today requires 1-line manual mode-enumeration edits in packages/core, packages/workspace, packages/cli, and apps/full-app plus config-schema unions in packages/agent — shotgun surgery paid three times now (bwrap → vercel-sandbox → blaxel, see #1236's file spread).

Goal: each provider package exports a self-describing mode descriptor (id, capabilities, error-code namespace, workspace+sandbox factory pair); hosts iterate a registry instead of hand-enumerating. The runtime-mode-pair invariant (Workspace+Sandbox swap together) is unchanged — the descriptor carries the pair, so the guarantee is preserved structurally.

Acceptance: adding a hypothetical provider requires changes ONLY in packages/boring-sandbox (+ the agent registry import list), proven by a test or a scripted dry-run; all existing providers migrated; zero behavior change for hosts; invariants + full CI green.

Sequencing: start only after #1236 (blaxel) merges — base includes all three real providers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1360, opened 2026-08-21
#1240 chore: centralize source aliases (extracted from #1256)

hachej:
## What

Extracts the clean, purely-subtractive half of PR #1256 and re-implements it from
current `main` (`99db0fb47`). No descriptor/registry/runtime-mode code is
included.

Sandbox source resolution is now declared once instead of being hand-maintained
in every consumer config:

- **new** `tsconfig.base.json` — `compilerOptions.customConditions: ["boring-source"]`
- **new** `scripts/vite-sandbox-alias.ts` — one regex alias entry
  (`@hachej/boring-sandbox/<subpath>` → `packages/boring-sandbox/src/<subpath>/index.ts`)
- `packages/boring-sandbox/package.json` — a `"boring-source"` export condition on
  each of the 9 existing subpath exports (`./shared`, `./providers`, and the 7
  provider subpaths). Node and normal bundlers never request this condition, so
  published/runtime resolution is unchanged.

That lets the hand-maintained mappings be deleted:

- **42 `paths` entries removed** from 7 tsconfigs
  (`apps/full-app`, `packages/{agent,cli,core}`,
  `packages/workspace/tsconfig.{front,server}.source.json`,
  `plugins/boring-automation`), each of which now `extends` `tsconfig.base.json`.
- **40 alias entries removed** from 7 vitest configs
  (`apps/{agent-playground,full-app}`, `packages/{agent,cli,core,workspace}`,
  `plugins/boring-automation`), replaced by the single shared
  `sandboxSourceAlias`. Three of those configs used the object alias form and
  were converted to the equivalent array form (Vite preserves order and match
  semantics) so the shared entry can be spliced in at the same position.

Net: **+87 / -143** across 18 files, zero runtime surface.

## Why

Tonight's review of #1256 concluded it is "two PRs wearing one coat":

- **Half A (this PR)** — the alias cleanup — "genuinely good, low-risk,
  mechanical, and solves a real recurring pain. It should land."
- **Half B** — the `SandboxRuntimeModeDescriptorV1` registry — is blocked:
  #1317 (merged 2026-08-17) ratified R-33-14/R-33-15, which reject the
  authority-mixing central type in that half. The author of #1256 reached the
  same conclusion in the PR body and stopped mid-reconciliation. #1256 is also
  59 commits behind `main` with a *semantic* conflict in the remote-worker path.

Rather than rebase 94 files, the review's recommendation was to cut the clean
subtraction fresh from current `main` and let the registry design go back through
#1317 reconciliation on its own schedule. That is exactly this PR.

Re-implemented against today's layout, not cherry-picked: `apps/agent-playground`
has no sandbox `paths` on main (only vitest aliases), `packages/boring-sandbox`
has no self-referencing imports, and the `./runtime-modes` and
`./providers/registry` export entries from #1256 are omitted because those source
directories do not exist on `main`.

Relates to #1240. Extracted from #1256 (which stays open for the registry half).

## Proof

Run in `.worktrees/alias-cleanup-extract` off `origin/main`:

```
pnpm install --prefer-offline
pnpm typecheck        # = pnpm build:packages && pnpm -r run typecheck
```
→ **EXIT=0**, 29 packages `typecheck: Done`, zero `error TS`.

Test suites for every package whose vitest config was touched:

```
pnpm --filter agent-playground        run test
pnpm --filter full-app                run test
pnpm --filter @hachej/boring-agent    run test
pnpm --filter @hachej/boring-ui-cli   run test
pnpm --filter @hachej/boring-core     run test
pnpm --filter @hachej/boring-workspace run test
pnpm --filter @hachej/boring-automation run test
```
Fully green:

| package | result |
|---|---|
| `agent-playground` | 1/1 files, 1/1 tests **pass** |
| `full-app` | 5/5 files, 46/46 tests **pass** |
| `@hachej/boring-agent` | 223 passed / 2 skipped files, 2047 passed / 17 skipped tests **pass** |

Four packages had failures, **all of which reproduce identically on pristine
`origin/main`**. Each was re-run in isolation on the branch and then on a
detached `origin/main` checkout in the same worktree, back to back:

| package | branch | baseline (`origin/main`) | verdict |
|---|---|---|---|
| `@hachej/boring-core` | 16 FAIL | **same 16 FAIL, identical test names** | pre-existing — shared Postgres (`raceConditions`, `postSignupHook`, `authHook`, `createAuth`, `deleteUserCompletely`, `PostgresModelBudgetStore`, `routes`) contended by other agents' concurrent suites on this VM |
| `@hachej/boring-ui-cli` | 2 files FAIL | **same 2 files FAIL** | pre-existing — `cli.integration` + `runtimePluginBrowser.integration` (browser 403s); baseline failed the whole `cli.integration` file at suite level |
| `@hachej/boring-workspace` | 1 file FAIL | **2 files FAIL** (baseline worse) | flaky — 5000ms `testTimeout` under load; a *different* test failed in each of the three runs |
| `@hachej/boring-automation` | 1 FAIL | **same 1 FAIL** | pre-existing — `migrations.test.ts` against the shared Postgres |

No failure anywhere in any run mentions `@hachej/boring-sandbox`, and there are
zero module-resolution errors — which is the shape a regression from this change
would take. Load average on the VM was ~140 during these runs (several other
agents running vitest concurrently).

Sanity check that nothing was left behind:

```
grep -rn 'boring-sandbox' --include='tsconfig*.json' --include='vitest.config.ts' .
```
→ no matches outside `node_modules`.

## Not in scope

- The descriptor registry, `runtimeModeCatalog`, host-policy plumbing, and the
  `'remote-worker'` mode-enum widening from #1256 — all deferred pending the
  #1317 R-33-15 split.
- Do not merge #1256 as a unit; it stays open for that rework.

### DIFF
--- a/apps/agent-playground/tsconfig.json
+++ b/apps/agent-playground/tsconfig.json
@@ -1,4 +1,5 @@
 {
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "target": "ES2022",
     "useDefineForClassFields": true,
--- a/apps/agent-playground/vitest.config.ts
+++ b/apps/agent-playground/vitest.config.ts
@@ -1,5 +1,6 @@
 import { resolve } from 'node:path'
 import { defineConfig } from 'vitest/config'
+import { sandboxSourceAlias } from '../../scripts/vite-sandbox-alias.ts'
 
 const repositoryRoot = resolve(import.meta.dirname, '..', '..')
 
@@ -12,12 +13,7 @@ export default defineConfig({
       { find: /^@hachej\/boring-agent\/shared$/, replacement: resolve(repositoryRoot, 'packages/agent/src/shared/index.ts') },
       { find: /^@hachej\/boring-bash\/server$/, replacement: resolve(repositoryRoot, 'packages/boring-bash/src/server/index.ts') },
       { find: /^@hachej\/boring-bash\/agent$/, replacement: resolve(repositoryRoot, 'packages/boring-bash/src/agent/index.ts') },
-      { find: /^@hachej\/boring-sandbox\/shared$/, replacement: resolve(repositoryRoot, 'packages/boring-sandbox/src/shared/index.ts') },
-      { find: /^@hachej\/boring-sandbox\/providers\/direct$/, replacement: resolve(repositoryRoot, 'packages/boring-sandbox/src/providers/direct/index.ts') },
-      { find: /^@hachej\/boring-sandbox\/providers\/bwrap$/, replacement: resolve(repositoryRoot, 'packages/boring-sandbox/src/providers/bwrap/index.ts') },
-      { find: /^@hachej\/boring-sandbox\/providers\/node-workspace$/, replacement: resolve(repositoryRoot, 'packages/boring-sandbox/src/providers/node-workspace/index.ts') },
-      { find: /^@hachej\/boring-sandbox\/providers\/blaxel$/, replacement: resolve(repositoryRoot, 'packages/boring-sandbox/src/providers/blaxel/index.ts') },
-      { find: /^@hachej\/boring-sandbox\/providers\/vercel-sandbox$/, replacement: resolve(repositoryRoot, 'packages/boring-sandbox/src/providers/vercel-sandbox/index.ts') },
+      sandboxSourceAlias,
     ],
   },
   test: { environment: 'node' },
--- a/apps/full-app/tsconfig.json
+++ b/apps/full-app/tsconfig.json
@@ -1,4 +1,5 @@
 {
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "target": "ES2022",
     "module": "ESNext",
@@ -23,12 +24,6 @@
       "@hachej/boring-agent/server": ["../../packages/agent/src/server/index.ts"],
       "@hachej/boring-bash/agent": ["../../packages/boring-bash/src/agent/index.ts"],
       "@hachej/boring-bash/server": ["../../packages/boring-bash/src/server/index.ts"],
-      "@hachej/boring-sandbox/shared": ["../../packages/boring-sandbox/src/shared/index.ts"],
-      "@hachej/boring-sandbox/providers/direct": ["../../packages/boring-sandbox/src/providers/direct/index.ts"],
-      "@hachej/boring-sandbox/providers/bwrap": ["../../packages/boring-sandbox/src/providers/bwrap/index.ts"],
-      "@hachej/boring-sandbox/providers/node-workspace": ["../../packages/boring-sandbox/src/providers/node-workspace/index.ts"],
-      "@hachej/boring-sandbox/providers/vercel-sandbox": ["../../packages/boring-sandbox/src/providers/vercel-sandbox/index.ts"],
-      "@hachej/boring-sandbox/providers/blaxel": ["../../packages/boring-sandbox/src/providers/blaxel/index.ts"],
       "@hachej/boring-core/app/front": ["../../packages/core/src/app/front/index.ts"],
       "@hachej/boring-core/app/server": ["../../packages/core/src/app/server/index.ts"],
       "@hachej/boring-core/app/vite": ["../../packages/core/src/app/vite/index.ts"],
--- a/apps/full-app/vitest.config.ts
+++ b/apps/full-app/vitest.config.ts
@@ -1,6 +1,7 @@
 import path from 'node:path'
 import { defineConfig } from 'vitest/config'
 import { createBoringAppViteAliases } from '../../packages/core/src/app/vite/index.ts'
+import { sandboxSourceAlias } from '../../scripts/vite-sandbox-alias.ts'
 
 const appRoot = import.meta.dirname
 const boringAliases = createBoringAppViteAliases({ appRoot })
@@ -17,12 +18,7 @@ export default defineConfig({
       { find: /^@hachej\/boring-bash\/server$/, replacement: path.resolve(repoRoot, 'packages/boring-bash/src/server/index.ts') },
       { find: /^@hachej\/boring-agent\/shared$/, replacement: path.resolve(repoRoot, 'packages/agent/src/shared/index.ts') },
       { find: /^@hachej\/boring-agent\/server$/, replacement: path.resolve(repoRoot, 'packages/agent/src/server/index.ts') },
-      { find: /^@hachej\/boring-sandbox\/shared$/, replacement: path.resolve(repoRoot, 'packages/boring-sandbox/src/shared/index.ts') },
-      { find: /^@hachej\/boring-sandbox\/providers\/direct$/, replacement: path.resolve(repoRoot, 'packages/boring-sandbox/src/providers/direct/index.ts') },
-      { find: /^@hachej\/boring-sandbox\/providers\/bwrap$/, replacement: path.resolve(repoRoot, 'packages/boring-sandbox/src/providers/bwrap/index.ts') },
-      { find: /^@hachej\/boring-sandbox\/providers\/node-workspace$/, replacement: path.resolve(repoRoot, 'packages/boring-sandbox/src/providers/node-workspace/index.ts') },
-      { find: /^@hachej\/boring-sandbox\/providers\/vercel-sandbox$/, replacement: path.resolve(repoRoot, 'packages/boring-sandbox/src/providers/vercel-sandbox/index.ts') },
-      { find: /^@hachej\/boring-sandbox\/providers\/blaxel$/, replacement: path.resolve(repoRoot, 'packages/boring-sandbox/src/providers/blaxel/index.ts') },
+      sandboxSourceAlias,
       { find: /^@hachej\/boring-core\/app\/server$/, replacement: path.resolve(repoRoot, 'packages/core/src/app/server/index.ts') },
       { find: /^@hachej\/boring-core\/server$/, replacement: path.resolve(repoRoot, 'packages/core/src/server/index.ts') },
       { find: /^@hachej\/boring-workspace\/app\/server$/, replacement: path.resolve(repoRoot, 'packages/workspace/src/app/server/index.ts') },
--- a/packages/agent/tsconfig.json
+++ b/packages/agent/tsconfig.json
@@ -1,4 +1,5 @@
 {
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "target": "ES2022",
     "module": "ESNext",
@@ -22,12 +23,6 @@
       "@hachej/boring-agent/front": ["src/front/index.ts"],
       "@hachej/boring-agent/server": ["src/server/index.ts"],
       "@hachej/boring-agent/shared": ["src/shared/index.ts"],
-      "@hachej/boring-sandbox/shared": ["../boring-sandbox/src/shared/index.ts"],
-      "@hachej/boring-sandbox/providers/direct": ["../boring-sandbox/src/providers/direct/index.ts"],
-      "@hachej/boring-sandbox/providers/bwrap": ["../boring-sandbox/src/providers/bwrap/index.ts"],
-      "@hachej/boring-sandbox/providers/node-workspace": ["../boring-sandbox/src/providers/node-workspace/index.ts"],
-      "@hachej/boring-sandbox/providers/vercel-sandbox": ["../boring-sandbox/src/providers/vercel-sandbox/index.ts"],
-      "@hachej/boring-sandbox/providers/blaxel": ["../boring-sandbox/src/providers/blaxel/index.ts"],
       "@agent-test-host": ["test-host/sandbox.ts"],
       "@/*": ["src/*"]
     },
--- a/packages/agent/vitest.config.ts
+++ b/packages/agent/vitest.config.ts
@@ -1,24 +1,20 @@
 import { fileURLToPath, URL } from "node:url";
 import { defineConfig } from "vitest/config";
+import { sandboxSourceAlias } from "../../scripts/vite-sandbox-alias.ts";
 
 export default defineConfig({
   resolve: {
-    alias: {
-      "@": fileURLToPath(new URL("./src", import.meta.url)),
-      "@hachej/boring-bash/agent": fileURLToPath(new URL("../boring-bash/src/agent/index.ts", import.meta.url)),
-      "@hachej/boring-bash/server": fileURLToPath(new URL("../boring-bash/src/server/index.ts", import.meta.url)),
-      "@hachej/boring-agent/core": fileURLToPath(new URL("./src/core/index.ts", import.meta.url)),
-      "@hachej/boring-agent/front": fileURLToPath(new URL("./src/front/index.ts", import.meta.url)),
-      "@hachej/boring-agent/server": fileURLToPath(new URL("./src/server/index.ts", import.meta.url)),
-      "@hachej/boring-agent/shared": fileURLToPath(new URL("./src/shared/index.ts", import.meta.url)),
-      "@agent-test-host": fileURLToPath(new URL("./test-host/sandbox.ts", import.meta.url)),
-      "@hachej/boring-sandbox/shared": fileURLToPath(new URL("../boring-sandbox/src/shared/index.ts", import.meta.url)),
-      "@hachej/boring-sandbox/providers/direct": fileURLToPath(new URL("../boring-sandbox/src/providers/direct/index.ts", import.meta.url)),
-      "@hachej/boring-sandbox/providers/bwrap": fileURLToPath(new URL("../boring-sandbox/src/providers/bwrap/index.ts", import.meta.url)),
-      "@hachej/boring-sandbox/providers/node-workspace": fileURLToPath(new URL("../boring-sandbox/src/providers/node-workspace/index.ts", import.meta.url)),
-      "@hachej/boring-sandbox/providers/vercel-sandbox": fileURLToPath(new URL("../boring-sandbox/src/providers/vercel-sandbox/index.ts", import.meta.url)),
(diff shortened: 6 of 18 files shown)


hachej:

**Verdict: BLOCK.**

The goal is right — 82 hand-maintained mappings is real debt and deleting them is worth doing. But the mechanism chosen is not equivalent to what it replaces, and the CI failure is not incidental flakiness: it is the mechanism's central premise being wrong. Beyond that, the PR trades one explicit idiom for a second implicit idiom while leaving ~80 equivalent entries untouched, so the repo ends up with *more* concepts, not fewer.

---

### 1. MUST-FIX — the core premise is false: `paths` had a third consumer, and it was esbuild

`scripts/vite-sandbox-alias.ts:8-11` states the contract:

> *"They now spread this one entry instead; TypeScript gets the equivalent mapping from the `boring-source` export condition."*

That is only true for `tsc`. The deleted `paths` blocks were also being read by **esbuild**, which auto-discovers the nearest `tsconfig.json` and honours `compilerOptions.paths` — and which has **no support for `customConditions`**.

Concretely, `packages/agent/scripts/assert-build-artifacts.mjs:160-171` runs:

```js
const result = await esbuild({
  absWorkingDir: packageRoot,
  bundle: true,
  packages: 'bundle',   // <- must really resolve @hachej/boring-sandbox/*
  ...
})
```

With no `tsconfig` option passed, esbuild picks up `packages/agent/tsconfig.json`. On `main`, `paths` mapped `@hachej/boring-sandbox/providers/*` to `../boring-sandbox/src/...` and resolution succeeded. This PR deletes those six entries and esbuild has nothing left, so it falls through to node resolution → `./dist/providers/*/index.js` → which is never built in that job. Result:

```
build-artifacts: FAIL
Error: Build failed with 5 errors:
  dist/chunk-GMYXEZQT.js:2159:7: ERROR: Could not resolve "@hachej/boring-sandbox/providers/blaxel"
  ... bwrap, direct, node-workspace, vercel-sandbox
```

This cascades into `Typecheck`, `Invariants` (`[boring-bash invariant] FAIL could not build and verify @hachej/boring-agent artifacts`), `Unit Tests`, `E2E`, both Bundle Budgets, and Reference Images — 8 red jobs from one root cause. It is green locally only because a local tree has `packages/boring-sandbox/dist` lying around from a previous build.

This is the finding that matters most, and it's a design finding, not a CI chore: **the replacement is not a drop-in.** `paths` is one setting honoured by three independent resolvers (tsc, vitest via its own alias plumbing, esbuild/tsup). `customConditions` + a vite alias covers two of them and silently drops the third. Any future resolver that reads `paths` (rolldown, oxc, ts-node, jiti, ESLint's TS resolver) is dropped too.

**Remedy — and this is the right one, not "put the paths back":** the assertion is currently only passing because a dev-only path mapping papers over the fact that `packages/agent`'s *published* dist genuinely imports `@hachej/boring-sandbox/providers/*` and those imports only resolve against unbuilt source. That is pre-existing rot this PR usefully exposed. Fix it properly: either build `@hachej/boring-sandbox` as a topological dependency before the agent build assert, or mark `@hachej/boring-sandbox/*` `external` in that esbuild call since it is a real runtime dependency and should never be bundled into the assert graph. Do that as a **separate, prior PR**, land it, and then this PR becomes the clean subtraction it wants to be.

---

### 2. MUST-FIX — the alias regex is over-broad and will silently resolve to files that do not exist

`scripts/vite-sandbox-alias.ts:12-15`:

```ts
export const sandboxSourceAlias = {
  find: /^@hachej\/boring-sandbox\/(.+)$/,
  replacement: resolve(import.meta.dirname, '../packages/boring-sandbox/src/$1/index.ts'),
}
```

`(.+)` matches **any** subpath and unconditionally appends `/index.ts`. It works today only because all nine export subpaths happen to be directories with an `index.ts`. Two problems that are live, not theoretical:

- `packages/boring-sandbox/src/providers/runtimeSupport.ts` and `src/providers/static.ts` are *files*. `@hachej/boring-sandbox/providers/runtimeSupport` now rewrites to `src/providers/runtimeSupport/index.ts` — a path that does not exist. Previously the import fell through to node resolution and either worked or failed loudly at the package boundary; now it fails as a mangled path that points at nothing, which is a materially worse debugging experience.
- The alias now covers `./providers`, `./providers/runsc` and `./providers/remote-worker`, **none of which were aliased before**. Every vitest config in this diff previously resolved those three to built `dist`; they now resolve to source. That is an undisclosed behaviour change in the test environment, and it is exactly the kind of thing that turns into a "why does this test only fail in CI" ticket six weeks from now.

This is rule 4 in its purest form: a generic mechanism hiding a simple data-shape assumption (`subpath === directory-with-index`) that nothing enforces.

**Remedy (the judo move):** stop hand-writing the regex and *derive* the aliases from the one place that already declares them. The `boring-source` condition in `packages/boring-sandbox/package.json` is already an exact subpath→source-file map. Read it:

```ts
import { createRequire } from 'node:module'
const pkg = createRequire(import.meta.url)('@hachej/boring-sandbox/package.json')

export const sandboxSourceAliases = Object.entries(pkg.exports)
  .filter(([, c]) => typeof c === 'object' && 'boring-source' in c)
  .map(([sub, c]) => ({
    find: new RegExp(`^@hachej/boring-sandbox${sub.slice(1)}$`),
    replacement: fileURLToPath(new URL(c['boring-source'], pkgUrl)),
  }))
```

That is one source of truth for both tsc and vitest, cannot drift, only matches subpaths that actually exist, and does not encode the directory convention. It is also strictly fewer concepts than what's here now.

Minor but related: `resolve()` is being applied to a string containing a regex backreference (`'$1'`). It happens to survive normalisation, but path-normalising a template is semantically wrong and would silently mangle the moment the template shape changes.

---

### 3. MUST-FIX — a dev-only resolution hint is now baked into a published artifact, pointing at files that aren't shipped

`packages/boring-sandbox/package.json:12-14` still declares `"files": ["dist"]`, but the exports map now contains nine `"boring-source": "./src/**"` entries. The published tarball therefore ships an exports map referencing paths that do not exist inside it. `publint` / `arethetypeswrong` will flag this, and any downstream consumer that happens to set that condition gets a hard resolution failure.

This is the strongest argument *against* the whole approach, and it deserves an explicit answer in the PR description rather than being left implicit: `paths` was dev-only and could never leak. `exports` is public API. Moving monorepo dev wiring into the published contract of a package is a boundary regression (rule 6), and the "82 lines deleted" win does not obviously pay for it.

If the condition approach is kept, `src` must be added to `files` (bloating the tarball) or the condition must be stripped at publish time (a new build step — more moving parts, which cuts against the PR's own thesis).

---

### 4. MUST-FIX — `tsconfig.base.json` is not a base config; as written it is pure indirection

`tsconfig.base.json` is five lines carrying exactly one option:

```json
{ "compilerOptions": { "customConditions": ["boring-source"] } }
```

Nine tsconfigs now gain an `"extends"` line to inherit one setting. That is one added line per config either way — so the file buys **zero** line reduction and costs a new file, a new extends chain, and a new precedence question. Rule 4: a thin wrapper that adds indirection without buying clarity.

Two ways out, and both are better than what's here:

- **Make it an actual base.** All nine configs independently repeat `"target": "ES2022"`, `"module": "ESNext"`, `"moduleResolution": "Bundler"`, plus strictness flags. Hoisting those is the change that justifies the file's existence and its name, and it deletes far more than this PR does.
- **Or delete the file** and inline `"customConditions"` in the nine configs. Same line count, one fewer concept.

Also note `packages/workspace/tsconfig.front.source.json:1` and `tsconfig.server.source.json:1` now use array `extends` where **later entries win**, so the file named "base" sits in the lowest-precedence slot while being listed first. It reads backwards, and it means any future `customConditions` in `tsconfig.front.json` silently overrides the base with no error. If array extends is kept, put the base last or add a comment.

---

### 5. FOLLOW-UP-BEAD — this creates a second idiom without retiring the first

The PR centralises source aliases for exactly one package. After it lands, the repo still hand-maintains roughly eighty equivalent mappings:

| config | remaining hand-written alias entries |
|---|---|
| `packages/cli/vitest.config.ts` | 25 |
| `packages/workspace/vitest.config.ts` | 12 |
| `apps/full-app/vitest.config.ts` | 11 |
| `plugins/boring-automation/vitest.config.ts` | 10 |
| `packages/core/vitest.config.ts` | 9 |
| `packages/agent/vitest.config.ts` | 8 |
| `apps/agent-playground/vitest.config.ts` | 5 |

plus ~25 `paths` entries in `packages/cli/tsconfig.json`, 22 in `apps/full-app/tsconfig.json`, and so on — for `boring-core`, `boring-workspace`, `boring-agent`, `boring-bash`, `boring-ui-kit`.

So a reader now has to know *two* rules: "sandbox subpaths resolve via a magic condition + a shared regex" and "everything else resolves via literal entries you can grep for." That is a net increase in concepts held in the head, which is the opposite of the stated goal.

If the `boring-source` condition is genuinely the right mechanism, it should be applied to **every** workspace package, with one derived `scripts/vite-source-aliases.ts` that walks the workspace and builds the full alias array from each package's `boring-source` entries. That deletes ~160 lines instead of 82 and leaves the repo with exactly one rule. If it is not worth doing repo-wide, that is strong evidence it is not worth doing for one package either.

Please split this: land the repo-wide version, or don't introduce the special case.

---

### 6. FOLLOW-UP-BEAD — the root export is asymmetric, leaving the package half-source, half-dist for tsc

`packages/boring-sandbox/package.json:15-18`: the `"."` export is the only one that did **not** get a `boring-source` condition, even though `packages/boring-sandbox/src/index.ts` exists. So under `customConditions`, `@hachej/boring-sandbox` resolves to `dist/index.d.ts` while `@hachej/boring-sandbox/shared` resolves to `src/shared/index.ts`.

Any type re-exported through both paths now has two distinct declarations in the same program, which produces the classic "Type X is not assignable to type X" monorepo failure. The old `paths` blocks had the same gap, so this is inherited rather than introduced — but this PR is the moment to close it, since it is codifying the mapping into a durable mechanism rather than a scratch list.

---

### What would make this a PASS

1. Land the `assert-build-artifacts` / sandbox-build-ordering fix first, as its own PR, so this diff is not carrying an unrelated pre-existing bug.
2. Derive the vite aliases from the `boring-source` entries instead of hand-writing a regex (§2).
3. Resolve the published-artifact question (§3) explicitly — `files`, or publish-time strip, or an argument for why it's fine.
4. Either make `tsconfig.base.json` a real base config or delete it in favour of inlining (§4).
5. State the intent on §5: repo-wide, or not at all.

Items 1–4 are **MUST-FIX-BEFORE-MERGE**. Items §5 and §6 are acceptable as **FOLLOW-UP-BEAD** provided §5 gets an explicit answer in the PR description.

The instinct here is good and the debt is real. The execution needs the mechanism to actually be equivalent to what it replaces, and right now CI is telling you it isn't.

hachej:
## CI repair investigation: confirmed root cause — recommend this stays draft pending an upstream prerequisite

All 10 failing checks on this PR (Typecheck, Unit Tests, Unit Tests Changed, Bundle Budget, Core Bundle Budget, E2E, Invariants, Reference Images, Runtime Refactor P8, PR Fast Summary) trace to **one** failure. It is not flake, not a missing/gitignored file — the commit is complete. It is the central premise of the mechanism.

### Root cause

`paths` had a **third consumer** beyond `tsc` and vitest: **esbuild**.

`packages/agent/scripts/assert-build-artifacts.mjs:160-171` runs `esbuild({ bundle: true, packages: 'bundle', ... })` with no tsconfig override, so esbuild auto-discovers `packages/agent/tsconfig.json` and honors its `paths`. **esbuild has no support for `customConditions`** — it is a `tsc`-only compiler option. So `tsconfig.base.json` replaces only the `tsc` half of what `paths` was doing; the bundler half has no replacement.

With the 6 `@hachej/boring-sandbox/*` paths entries deleted, `packages/agent`'s tsup output no longer inlines sandbox source — it emits bare external specifiers, and `build:assert` then has to resolve them against `packages/boring-sandbox/dist`, which does not exist in CI.

### CI evidence (run `32538523567`, job Typecheck; identically in Workflow Invariants `32538523656`)

```
packages/agent build: build-artifacts: FAIL
packages/agent build: Error: Build failed with 5 errors:
packages/agent build: dist/chunk-GMYXEZQT.js:2159:7: ERROR: Could not resolve "@hachej/boring-sandbox/providers/blaxel"
packages/agent build: dist/chunk-GMYXEZQT.js:2163:7: ERROR: Could not resolve "@hachej/boring-sandbox/providers/bwrap"
packages/agent build: dist/chunk-GMYXEZQT.js:2166:7: ERROR: Could not resolve "@hachej/boring-sandbox/providers/direct"
packages/agent build: dist/chunk-GMYXEZQT.js:2181:7: ERROR: Could not resolve "@hachej/boring-sandbox/providers/node-workspace"
packages/agent build: dist/chunk-GMYXEZQT.js:2187:7: ERROR: Could not resolve "@hachej/boring-sandbox/providers/vercel-sandbox"
```

Three things confirm the diagnosis:

1. **The 5 failing specifiers are exactly the 5 provider entries deleted from `packages/agent/tsconfig.json`.** `./providers/runsc` and `./providers/remote-worker` got the `boring-source` condition too but never had a `paths` entry, and they do not fail.
2. **New rollup warnings appear in the same job**, absent on `main`, showing the specifiers became external where they were previously inlined:
   `"createBlaxelSandboxProvider" and "BLAXEL_WORKSPACE_ROOT" are imported from external module "@hachej/boring-sandbox/providers/blaxel" but never used in "host/sandbox.ts".`
3. **`packages/boring-sandbox` never builds in CI.** Grepping the full Typecheck job log for `boring-sandbox` returns only the 4 rollup warnings and the 5 esbuild errors — zero `packages/boring-sandbox build` lines. Its `dist/` is never produced, because until now every consumer resolved it from source via `paths`. `build:packages` reaches `packages/agent build` before sandbox has built.

### Local reproduction

`pnpm typecheck` was green locally only because a stale `packages/boring-sandbox/dist` existed in the worktree. Moving that `dist` aside and re-running `pnpm --filter @hachej/boring-agent run build:assert` reproduces the CI failure **byte-for-byte, including the identical offsets** (2159 / 2163 / 2166 / 2181 / 2187). Restoring `dist` makes it pass again.

I also probed whether `esbuild`'s `conditions` option could stand in for `customConditions`: passing `conditions: ['boring-source']` (alone, and combined with `import`/`default`/`node`) does **not** resolve these specifiers in this pnpm layout. There is no drop-in bundler-side equivalent.

### Additional blocking findings (independent of CI)

- **The alias regex is over-broad.** `scripts/vite-sandbox-alias.ts` uses `/^@hachej\/boring-sandbox\/(.+)$/` → `src/$1/index.ts` — an unconditional `/index.ts` suffix. It mangles the non-directory sources `src/providers/runtimeSupport.ts` and `src/providers/static.ts`, and it silently re-points `./providers`, `./providers/runsc` and `./providers/remote-worker` from dist to source in vitest, which the hand-written entries it replaces deliberately did not do.
- **The `boring-source` condition ships dev-only wiring in a published artifact.** `packages/boring-sandbox/package.json` declares `files: ["dist"]`, but its `exports` now reference `./src/**` — paths absent from the tarball. That is a publint/attw failure surface on release.
- **`tsconfig.base.json` saves nothing net** at its current size relative to what it costs to keep correct across three resolvers.

### Recommendation

Keep this PR **draft / closed** as-is. The deleted `paths` entries are not hand-maintained boilerplate — they are the load-bearing build configuration by which `boring-sandbox` is consumed as a source-only package that CI never builds. They cannot be removed until that is no longer true.

The prerequisite must land as its **own PR first**, one of:

- make `packages/boring-sandbox`'s build topologically ordered ahead of `packages/agent` in `build:packages` so its `dist` exists, or
- mark `@hachej/boring-sandbox/*` external for the `assert-build-artifacts.mjs` esbuild step.

Once sandbox no longer needs source resolution at bundle time, the alias cleanup can be reconsidered — with the regex narrowed to the exact exported subpaths and the `boring-source` condition reconciled against `files: ["dist"]`.

No code changes were pushed to this branch; it is left exactly as the author committed it.

hachej:
## Closing — the premise is wrong, and the prerequisite is now filed as #1361

Two independent passes (a CI diagnosis and a thermo-nuclear quality review) reached the same conclusion: the `paths` entries this PR deletes are **not** hand-maintained boilerplate. They are the load-bearing build configuration by which `boring-sandbox` is consumed as a source-only package that CI never builds — `build:packages` reaches `packages/agent` before sandbox's `dist` exists.

`customConditions` is a `tsc`-only option, so `tsconfig.base.json` replaces only the `tsc` half of what `paths` did. **esbuild** — the third consumer nobody accounted for — auto-discovers `packages/agent/tsconfig.json` and honors `paths`, but ignores `customConditions`. `assert-build-artifacts.mjs:160-171` bundles with no tsconfig override, so the five deleted provider specifiers become unresolvable, and that single failure cascades into all 10 red checks. Probing `conditions: ['boring-source']` on the esbuild side (alone and combined with `import`/`default`/`node`) does not resolve them in this pnpm layout — there is no drop-in bundler-side equivalent.

It was green locally only because a stale `packages/boring-sandbox/dist` existed. Moving that dist aside reproduces CI byte-for-byte, including identical error offsets.

Three further blockers, independent of the above:
- The alias regex `/(.+)$/` plus an unconditional `/index.ts` mangles `providers/runtimeSupport.ts` and `providers/static.ts` (files, not directories), and silently re-points `./providers`, `./providers/runsc` and `./providers/remote-worker` from dist to source.
- `files: ["dist"]` while `exports` now reference `./src/**` breaks publint/attw on release — dev-only wiring moved into a published artifact.
- The change covers one package while ~80 equivalent hand-written mappings remain elsewhere, leaving two idioms rather than fewer concepts.

**The prerequisite is now #1361**: make `boring-sandbox` build topologically ahead of `packages/agent`, or mark `@hachej/boring-sandbox/*` external for that esbuild step. Once that lands, this cleanup becomes viable — and should then be done repo-wide with aliases derived from the exports map rather than a hand-written regex.

The branch `chore/1240-alias-cleanup-extract` is preserved, so this can be reopened once #1361 is done. Closing so it isn't mistaken for reviewable work.

Issue #1338, opened 2026-08-19
Session inventory scans and parses the whole transcript store per request (16s list, 'Preparing workspace' hang)


hachej:
## Symptom (0.1.101 anchor hub)

- \"Preparing workspace\" hangs; session history takes tens of seconds.
- Measured: `GET /api/v1/agents/boring-worker/sessions` = **16.5 s**; `default` = 12.3 s **returning empty**. The front calls this per seat at boot → the hang.

## Cause

The shared session root held **1,316 transcripts / 194 MB across 549 session dirs**. The inventory (touched by #1331's cross-hub lineage listing, `sessionInventory.ts`) walks and parses the store per request — O(files × size), no cache, no pagination, cost paid even for an empty result.

## Aggravator worth its own fix

**E2E tests write into the production session root** — hundreds of `--dev-shm-…-e2e-…` session dirs polluted the store. Tests must use an isolated `BORING_AGENT_SESSION_ROOT`.

## Mitigation applied (ops)

Archived 460+ dirs untouched since Aug 17 to `~/.factory/pi-sessions-archive` (moved, restorable). List time: 16.5 s → **1.8 s**. This is a workaround; growth restores the problem.

## Required

1. Incremental inventory: summaries from cheap metadata (first/last lines + mtime), persistent cache invalidated per file mtime, never a full-store parse on request.
2. Pagination: recent N by default, rest on demand.
3. Boot: workspace must not gate on full inventories of every seat.
4. Tests: isolated session root.

hachej:
Ops mitigation completed: active store reduced to 15 files / 4.9 MB (rest archived). Worker/orchestrator seat listings: 16.5s → **0.5s**. Remaining anomaly for the fix: `GET /api/v1/agents/default/sessions` still costs **4.7s while returning ~empty** — its inventory cost is independent of the transcript store size, so the default-seat path scans something else (legacy/path-derived locations?). Profile that path specifically.

PR #1363, opened 2026-08-22
Git branch in workspace chrome + session status on session-list rows

hachej:
Two independent Roadmap board drafts (board items, no issue numbers). Kept as two commits so either can be dropped without touching the other.

---

## 1. Show the current Git branch for Git-backed workspaces

`059dc8e` — `packages/boring-bash`, `packages/cli`

**What the investigation found.** Nothing in the workspace model records that a workspace is Git-backed: `LocalWorkspace`, `toCoreWorkspace()`, and the core workspace runtime have no `git`/`repo`/`vcs` field, and there is no clone-a-repo provisioning path anywhere. Git-ness is discovered on demand by shelling out — that is already how `resolveGitFileUrl` works, and how `detectFolderModeTaskProviders` probes for a GitHub remote at boot.

**So it reuses that, exactly.** `resolveGitBranch()` sits next to `resolveGitFileUrl` in `packages/boring-bash/src/server/git/` and goes through the same `__gitTestUtils.runGit` seam, so:
- no new git dependency (the repo has none — no simple-git, no isomorphic-git);
- tests stub git without spawning a process;
- the invariant that routes stay free of `node:child_process` is preserved.

It is two plumbing commands (`rev-parse --show-toplevel`, `symbolic-ref --quiet --short HEAD`), with `rev-parse --short HEAD` only when HEAD is detached. No log walk, no `git status`.

**Nothing to wire per host.** `GET /api/v1/git/branch` rides the existing `gitRoutes` registration, so it lights up wherever `/api/v1/git/file-url` already does: CLI folder mode, CLI workspaces mode, standalone agent host, and core. Runtimes without a strong workspace fs capability have no host root and get `{ enabled: false }` without git ever being invoked.

**Refresh — why not a poll.** The badge fetches on mount, on workspace switch, and on window focus. A branch changes only through work done outside the tab (a terminal, an agent run), so regaining focus is precisely when the value can be stale; a timer would spend requests to learn nothing while you are typing. This mirrors the `refreshWorkspaces()` focus listener already in `CliWorkspaceShell`. Cost per refresh: one `git symbolic-ref`.

**Presentation.** `CliGitBranchBadge` reuses `CliVersionBadge`'s pill styling verbatim (same border, background, `text-[10px]`, radius) and sits next to it in `topBarRight`, with a lucide `GitBranch` glyph at `size-3` — the icon set already used in `WorkspaceSwitcherControl`. Long branch names truncate inside `max-w-40`. **Non-Git workspaces render nothing** — the component returns `null`, so there is no error text and no empty chrome slot, exactly like `CliVersionBadge` with no version. Detached HEAD shows the short sha with a "Detached HEAD at …" title.

**Proof**
- `packages/boring-bash`: `src/server/git/gitBranch.test.ts` (branch, detached HEAD, not-a-repo, unborn branch — and asserts the branch path stays at two git calls and that not-a-repo bails after one), plus two route tests in `src/server/routes/git.test.ts`.
- `packages/cli`: `App.test.tsx` gains "shows the checked-out branch in the top bar for a Git-backed workspace" and "renders no branch chrome for a workspace that is not Git-backed" (asserts absence *after* the probe resolved, so it cannot pass vacuously).
- **Revert check:** with only the implementation removed and the tests kept, the two route tests fail `expected 404 to be 200` and the CLI badge test times out waiting for `[data-boring-cli-part="git-branch"]`.

---

## 2. Session status visibility in the session list

`cb96d70` — `packages/agent`, `packages/workspace`

**The taxonomy is the server's, not invented.** `AgentSessionActivityIndex` tracks exactly four values — `idle | running | aborting | error` — in a process-lifetime in-memory `Map`. `embeddedGateway.listSessions()` already stamps it onto every summary, and `usePiSessions.toAddressedSessionSummary` already normalizes it into `SessionSummary.status`. The rows simply never rendered it. There is no "awaiting input" state anywhere in the session model, so none is shown.

**Perf: this adds nothing per row (gh-1338).** Status is a `Map.get` server-side — it never touches the transcript store, and the expensive part of inventory (title/turnCount/updatedAt) is untouched here. In the browser it rides the `session-activity` SSE stream that `WorkspaceAgentFront` already subscribes to. No new request, no extra round trip, no per-row work.

**One honesty fix was needed.** The activity stream previously only triggered a refresh for *unknown* sessions, so a listed row's `status` was frozen at whatever the last inventory fetch saw — a row would have sat on "running" forever after its turn ended. `usePiSessions.applyActivity(id, status)` patches local list state only (no request, no refetch) and returns the same array identity when nothing changed, so idle heartbeats cannot churn every consumer of the list.

**Presentation is deliberately quiet.** `idle` renders nothing at all — it is the resting state of nearly every row, and a column of dots would carry no information while costing the list its calm. The three remaining states reuse the tool-call group's existing dot vocabulary verbatim (`bg-amber-500/70` working, `bg-muted-foreground/55` stopping, `bg-destructive ring-2 ring-destructive/20` failed) in the same `size-1.5 rounded-full` inside a `size-4` centering wrapper, so chat speaks one status language. The dot sits in the trailing slot before the delete affordance, which keeps titles from shifting when a status appears. Screen readers get the label; the dot itself is `aria-hidden`. No new component, no new dependency, no badge chrome.

**Proof**
- `packages/agent`: `SessionList.test.tsx` — "conveys live session status on the row, and stays silent when idle" covers all four states plus a status-less summary; `usePiSessions.test.tsx` — "applies live activity to a listed session without refetching" asserts the status flips, that the fetch count does **not** move, and that an unknown session id leaves the array identity untouched.
- **Revert check:** with only the implementation removed and the tests kept, `expected undefined to be 'running'` and `TypeError: result.current.applyActivity is not a function`.

---

## Coordination

`fix/1338-session-inventory-perf` was read, not touched. It currently carries no inventory-perf commits and its diff touches `PiChatPanel.tsx` and `server/harness/pi-coding-agent/sessions.ts`; this branch touches neither. `SessionList.tsx`, `usePiSessions.ts`, `sessionInventory.ts`, `embeddedGateway.ts`, and `httpProjection.ts` are untouched by that branch, so there is no overlap.

## Visual verification

Pending — see the note added before merge.

## Review fixes (thermo gate round 1)

- **CRITICAL — visible list never received live updates**: `PiChatPanel` now owns the session-activity SSE stream beside its rendered `SessionList` (`useSessionListActivity`), so standalone rows track running→idle→error transitions. Integrated transition test added to `PiChatPanel.test.tsx`. Disposition note: workspace-side rows keep their existing working-chip path; full terminal-status rendering on workspace chrome rows belongs to #1364's surface, not this branch.
- **CRITICAL — `applyActivity` had no source guard**: now fails closed via `sourceIsCurrent(requestScopeKey)` like every other mutation; a callback retained from a previous source can no longer mutate a colliding session id. Extended the saved-callback source-switch test.
- **SHOULD-FIX — `CliGitBranchBadge` focus race**: fetch commits are sequenced; a late mount response can no longer overwrite a newer focus response. Focus-race regression test added.

### DIFF
--- a/packages/agent/src/front/chat/PiChatPanel.tsx
+++ b/packages/agent/src/front/chat/PiChatPanel.tsx
@@ -45,7 +45,7 @@ import {
   writePiComposerThinking,
   type ActiveSessionStorageLike,
 } from './session'
-import { SessionList, usePiSessions, type UsePiSessionsOptions } from './session'
+import { SessionList, usePiSessions, useSessionListActivity, type UsePiSessionsOptions } from './session'
 import {
   type ComposerBlocker,
   type ComposerBlockerAction,
@@ -307,6 +307,14 @@ export function PiChatPanel<
       if (sessionListRefreshRef.current === refreshSessionList) sessionListRefreshRef.current = undefined
     }
   }, [externalSessionId, sessions.refresh])
+  // Live-activity ownership sits beside the rendered list: a visible
+  // SessionList must see status transitions (running → idle → error), not a
+  // hidden one. Local state only — no request, no transcript read (gh-1338).
+  useSessionListActivity({
+    apiBaseUrl,
+    enabled: showSessionSidebar,
+    onActivity: (sessionId, status) => sessions.applyActivity(sessionId, status),
+  })
   const externalPiSession = useExternalRemotePiSession({
     sessionId: externalSessionId,
     agentTypeId,
--- a/packages/agent/src/front/chat/__tests__/PiChatPanel.test.tsx
+++ b/packages/agent/src/front/chat/__tests__/PiChatPanel.test.tsx
@@ -220,6 +220,54 @@ describe('PiChatPanel sandbox shell', () => {
     await waitFor(() => expect(fetchMock).toHaveBeenCalledWith('/api/v1/agents/default/sessions', expect.objectContaining({ method: 'POST' })))
   })
 
+  test('live session-activity transitions reach the rendered SessionList rows', async () => {
+    class FakeEventSource {
+      static last: FakeEventSource | undefined
+      closed = false
+      listeners = new Map<string, Array<(event: MessageEvent) => void>>()
+      constructor(public url: string) { FakeEventSource.last = this }
+      addEventListener(type: string, listener: (event: MessageEvent) => void) {
+        const list = this.listeners.get(type) ?? []
+        list.push(listener)
+        this.listeners.set(type, list)
+      }
+      removeEventListener() {}
+      close() { this.closed = true }
+      emit(type: string, data: unknown) {
+        for (const listener of this.listeners.get(type) ?? []) listener({ data: JSON.stringify(data) } as MessageEvent)
+      }
+    }
+    const previousEventSource = (globalThis as { EventSource?: unknown }).EventSource
+    ;(globalThis as { EventSource?: unknown }).EventSource = FakeEventSource
+
+    const fetchMock = vi.fn().mockResolvedValue(jsonResponse([session('pi-1')]))
+    try {
+      const view = render(<PiChatPanel showSessions serverResourcesEnabled={false} storageScope="scope-a" fetch={fetchMock as unknown as typeof fetch} createRemoteSession={remoteFactory(new FakeRemotePiSession(remoteState()))} />)
+      const row = await screen.findByText('Session pi-1')
+      expect(row.closest('[data-boring-agent-part="session-row"]')).toBeTruthy()
+
+      const source = FakeEventSource.last!
+      expect(source.url).toBe('/api/v1/agents/session-activity/events')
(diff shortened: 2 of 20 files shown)


hachej:
Closing instead of splitting after reevaluating the two slices against current `main`:

- **Git branch visibility:** dropped per owner direction.
- **Session status visibility:** the shipped workspace session surfaces already show live `working` and attention/needs-input state through `useWorkingSessionIds`, the `session-activity` SSE stream, and `SessionBrowser`/`AppLeftPane` badges (including the live-sync fix merged in #1112).
- The remaining unique delta in this PR is limited to status dots in the standalone `PiChatPanel` session sidebar. Workspace chat explicitly uses `showSessions: false`, so the added workspace `applyActivity`/fleet plumbing does not create a new primary workspace capability. It also does not satisfy the original fuller backlog acceptance (completed + needs-input on that standalone surface).

A second PR would therefore preserve duplicate/incomplete scope rather than produce a clean independent slice. No replacement PR opened. Remote branch retained; no branch deletion or history rewrite.

PR #1364, opened 2026-08-22
fix(workspace): render completed and failed session states

hachej:
## Which case this was: the state already existed

I verified the triage claim before writing anything. `completed` and `failed` genuinely never render — but **not** because the outcome is unknown. It was known and projected the whole way down, and then discarded at the last step.

The chain that already exists on `main`:

- `SessionSummary.status` — `idle | running | aborting | error` (`packages/agent/src/shared/session.ts:39`), projected from AgentHost.
- It survives onto `WorkspaceAgentSession.status` and `SessionItem.status` (`packages/workspace/src/front/components/SessionList.tsx:23`), so **every row in the list already carries it**.
- The same status streams live on the session-activity SSE channel (`startSessionActivityStream`, `AgentSessionActivity`).

`SessionBrowser` consumed that status through exactly one path: `useWorkingSessionIds`, which collapses it to `running || aborting` and throws everything else away. `error` reached the row and was dropped on the floor. So this is a **presentation + projection** change, not a derivation problem — no new server state, no new computation.

The one genuinely absent piece is `completed`, because `idle` is ambiguous: it is both "the run finished" and "this session never ran". Marking every idle row `completed` would tag the entire list with a chip that carries no information. The cheap correct signal is the **transition**: a key leaving the working set the browser already maintains. That is a set difference over data already in hand.

## Performance argument (gh-1338)

Nothing here scans or parses a transcript, and nothing adds a request — per row or otherwise.

- `failed` reads `session.status`, a field already on the row the list already fetched, plus the existing `boring:chat-session-status` event. No new fetch.
- `completed` is a set difference between the previous and current `workingSessionIds` — the set `SessionBrowser` already computes for the `working` chip. No new fetch.
- The only wire change is **additive**: the workspace front now includes the `status` it already has in the status event's detail, so `failed` appears immediately instead of waiting for the next list fetch. Same event, same frequency, one more field.

Net added work is O(rows) over in-memory data, inside the `useMemo` that already walks the rows. This does not interact with the transcript-store scan being fixed on `fix/1338-session-inventory-perf`, and touches none of its files.

## UI

No new visual language. The chip reuses the existing attention-badge shape (`rounded-full px-1.5 py-0.5 text-[10px] font-medium`) and the existing `sessionBadgeToneClassName` / `sessionBadgeDotClassName` helpers rather than introducing colors:

- `failed` → the existing `danger` tone: `bg-destructive/12 text-destructive`. A 12% tint at 10px on only the row that actually errored — noticeable, not a wall of red.
- `completed` → the existing `neutral` tone, the same `bg-foreground/[0.07] text-muted-foreground` the `working` chip already uses. Label `done`, matching the lowercase register of `working` / `needs input`.

Both use a **static** dot instead of the live `animate-pulse`, which is the honest distinction: these are settled, not live. Precedence is attention badge → working → terminal → timestamp, so a live state always beats a settled one and the row never gains a second chip.

## Proof

- `packages/workspace` — `SessionBrowser.test.tsx`: **28/28 pass**, covering 5 new behaviors and every pre-existing badge test unchanged.
  - `completed` appears on the working → finished transition, is quiet (neutral tone, no pulse), and retires after the window.
  - idle sessions that never ran are **not** marked done (the regression this design avoids).
  - `failed` renders from the server-reported `error` row status, and only on that row.
  - `failed` renders from a live activity frame before the list refetches, and a failure is never also reported as done.
  - a previously-failed session running again shows `working`, not `failed`.
- `pnpm run typecheck` (workspace, front + server projects) — see the comment below for the run against fully built dependency dists.
- `git stash list` checked; working tree carries only the four files in this diff.

No screenshots: driving the real workbench needs the hub plus built dists on a VM that is currently saturated with concurrent builds from other worktrees. Stating that plainly rather than faking it — the chip markup and tone classes are asserted directly in the tests instead.

## Overlap check

- `origin/feat/git-branch-and-session-status` does not exist on the remote yet. The recovered stash carrying that work (`stash@{0}`) is a `gitBranch` route plus CLI App changes only — no session status presentation, and no terminal states. Inspected read-only; not touched.
- `gh pr list --search "session status"` surfaces no open PR covering terminal states. The nearest, #1112 (merged), added the live status sync this change consumes.
- Scope kept to `SessionBrowser` as triaged. The second session surface, `AppLeftPaneSessionRow`, has its own badge vocabulary and is deliberately left for a follow-up rather than widened into here.

## Review fixes (sol-xhigh thermo gate round 1)

- **CRITICAL (fixed):** cancelled runs rendered as completed — `agent-end{status:'aborted'}` now maps to a new `'aborted'` activity outcome end-to-end (shared types, activity index, SSE projection, session lists). The terminal-state hook skips `completed` for aborted finishes and retires the chip when the aborted status lands late. Regression: SessionBrowser abort test (both orderings) + `sessionActivityOutcomes.test.ts`.
- **CRITICAL (fixed):** terminal caches survived workspace switches — caches are scoped by workspace id (`activityWorkspaceId` → `scopeKey`); a source switch resets them and live events tagged with a foreign workspace are ignored. Regression: colliding-id workspace-switch test + foreign-workspace event filter test.
- **SHOULD-FIX (dispositioned, not taken):** consolidating the second global listener into one scoped reducer — deferred; the two listeners have different lifecycles (panel-local optimistic vs server-projected outcomes), and merging them would widen this fix's blast radius. Can follow as its own refactor PR.

## Review fixes (round 2 — verification pass)

- `useWorkingSessionIds` now takes the same workspace `scopeKey` as
  `useTerminalSessionStates`: live events tagged with a foreign workspace are
  ignored and source switches reset the optimistic set, closing the last
  cross-workspace bleed (a foreign finish can no longer manufacture a
  `completed` chip on a colliding id). Regression added.
- CI: all 15 checks green on commit c3fe2caea.


hachej:
CI fix pushed (6088d4e): the branch's only failing check was `Unit Tests Changed` — the committed test still asserted the pre-change activity payload shape. The worker's unfinished update asserting the new `status` field was sitting uncommitted in the worktree; finished, verified locally (`vitest run WorkspaceAgentFront.test.tsx SessionBrowser.test.tsx` — 119 passed), and pushed.

hachej:
Closing as over-scoped. The original presentation change expanded into a general run-lifecycle state machine (+2.1k lines) with unresolved semantic edge cases. The branch and review history are preserved. Minimal explicit-failure rendering is tracked in #1421; inferred semantic completion will be designed separately.

Issue #1197, opened 2026-08-10
[epic #1187] Workspaces mode publishes no persona instruction links — every seat loses its Open button


hachej:
Found running gh-1187 **S0/S1** on the factory hub.

## Symptom

Booting the CLI hub in **workspaces mode** with `BORING_AGENT_FLEET=1` logs this once per seat, at every boot:

```
warn agent-fleet-loader "fleet seat instructions not linkable"
  seat=triage        code=AGENT_FLEET_SEAT_INSTRUCTIONS_PATH_UNPUBLISHABLE
  seat=orchestrator  code=AGENT_FLEET_SEAT_INSTRUCTIONS_PATH_UNPUBLISHABLE
  seat=worker        code=AGENT_FLEET_SEAT_INSTRUCTIONS_PATH_UNPUBLISHABLE
  message="this host resolves a workspace root per request, so persona instructions
           have no single \"user\" path to be addressed against"
```

`instructionFiles` is therefore empty for **all** seats. The behaviour is deliberate and correct as written (`loadConfiguredAgentFleet`: a well-formed path the workbench cannot open is worse than no link, and the CLI hub passes `workspaceRoot: null` — `packages/cli/src/server/modeApps.ts:963`). The problem is the consequence, not the logic.

## Why it matters

"Read this seat's actual instructions" is a load-bearing part of the migration's observability story — the Agent details overlay (#1176) is how the owner verifies a seat is what `fleet.yaml` + `policy.yaml` say it is, and gh-1187's S6 proof is literally "the overlay reflects a seat's changed skill set". In the multi-workspace hub — **the** deployment shape the factory runs on — that link is structurally absent for every seat, so the overlay can never show it here.

Note this is orthogonal to #1196: fix the symlink 500 and the overlay loads, but the instructions row is still empty.

## Proposed (one of)

1. Resolve `instructionFiles` **per request**, once the request's workspace root is known, instead of once at composition time.
2. Publish personas on a dedicated non-`user` filesystem (the agent-resources filesystem already exists for package resources) so the ref is addressable regardless of which workspace is being served.
3. Serve instruction text through `/describe` itself when no openable path exists — read-only, but not blank.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QtPPbToGDvkARQZuYKFyWN

hachej:
Audit 2026-08-19: duplicate of #1189 (instruction-link resolution in CLI hub mode).

PR #1365, opened 2026-08-22
fix(agent): resolve Agent instruction refs per request so CLI hub mode gets working links

### LINKED ISSUE #1189: Resolve Agent instruction refs per request so CLI hub mode gets working links
## Problem
`GET /api/v1/agents/:agentTypeId/describe` publishes `instructionFiles` only when the fleet loader is given the workspace root that backs the `user` filesystem. The CLI hub cannot supply one at boot, because it serves a **different root per registered workspace**, so it passes `workspaceRoot: null` (`packages/cli/src/server/modeApps.ts:959`):

```
// The hub serves a DIFFERENT root per registered workspace, so there is no
// single one persona instruction refs could be addressed against.
agents: await agentServer.resolveDefaultAgentFleet({ workspaceRoot: null }),
```

Consequence: **CLI hub users never see a persona-instructions link**, on any Agent. Single-workspace CLI mode is fine (it goes through `createWorkspaceAgentServer` with a real root), as is the workspace server. `null` is the honest answer today — a withheld link beats one that opens nothing — but it is not the good answer.

## Ask
Resolve instruction refs **at request time**, against the workspace actually being served for that request, instead of once at composition time:

- The Host knows the seat's persona directory; the request knows the workspace scope and its served root.
- Publish a ref only when the persona source is reachable through that request's filesystem; otherwise withhold with the existing stable diagnostic `AGENT_FLEET_SEAT_INSTRUCTIONS_PATH_UNPUBLISHABLE`.
- Preserve the invariant established in #1176: **a rendered link always resolves.**

## Why canonicalization matters here specifically
CLI users run against arbitrary directories, and dotfiles managers (stow, chezmoi) make a symlinked `.agents` ordinary rather than exotic. A lexical containment check would call `~/project/.agents -> ~/dotfiles/agents` "inside the workspace" and publish a dead link. #1176 added `realpath`-based confinement (`admitPersonaSource`) precisely so the check follows symlinks; per-request resolution should reuse it rather than re-implement path math.

## Relationship to #1186
Same shape: a binding resolved per request/scope rather than fixed at composition time. #1186 does it for agent-scoped knowledge sources; this does it for instruction refs. Worth designing the two together so there is one mechanism, not two.

## Out of scope
Not part of #1176, which is a UI PR. Raised while reviewing it.

https://claude.ai/code/session_016qY6FajM6TUYpT8nxwfEXm

hachej:
Closes #1189 (and #1197, closed as a duplicate of it).

## Root cause

Two facts were fused into one decision taken at boot:

1. **seat → persona directory** — a composition fact. Only `loadConfiguredAgentFleet` knows it.
2. **is that persona reachable through the `user` filesystem?** — a *request* fact. It depends on which workspace root is being served.

The loader answered both at once: it took a single `workspaceRoot`, did the containment check there, and froze a workspace-relative `instructionFiles` ref onto the compiled spec. `describe` then just echoed `spec.instructionFiles`.

The CLI hub serves a **different root per registered workspace** off **one shared fleet**, so it had no single root to give and honestly passed `workspaceRoot: null` (`packages/cli/src/server/modeApps.ts:963`). Consequence: `AGENT_FLEET_SEAT_INSTRUCTIONS_PATH_UNPUBLISHABLE` for every seat at every boot, and **no persona-instructions link for any agent in hub mode** — the deployment shape the factory actually runs on (#1197 S0/S1). Core had the same shape.

## Fix

Split the two facts back apart.

- The loader records `instructionSources` — canonical **host absolute paths** (`realpath`'d persona root + `instructions.md`). No root, no path algebra, no link decision at composition time. `AGENT_FLEET_SEAT_INSTRUCTIONS_PATH_UNPUBLISHABLE` leaves the loader's diagnostic union.
- New `resolveAgentInstructionFileRefs` (`packages/agent/src/server/agentDefinition/instructionFileRefs.ts`) addresses those sources against a served root: `realpath` **both sides**, canonical containment, then the same publishable-path guard the browser applies to every openable resource. It returns refs plus withheld entries carrying the unchanged stable code.
- `describeAgent` calls it with `resolved.environment.workspaceRoot` — the root **this** request is served from — and logs each withheld ref with the stable code on `request.log`.

`authorizeAgentAccess` already resolves the request's environment scope for its per-agent authorization check, so the served root costs **no extra resolution**; the only added work is the confinement check itself (two `realpath`s).

Symlink handling is deliberately preserved from #1176: canonicalizing both sides is what stops a stowed/chezmoi'd `.agents` from being called "inside the workspace" and published as a dead link. The #1176 invariant holds unchanged — **a rendered link always resolves** — it is now enforced per request instead of per boot.

`workspaceRoot` is removed from `LoadConfiguredAgentFleetOptions` / `ResolveDefaultAgentFleetOptions` rather than left accepted-and-ignored; the three hosts and the playground are updated.

## What changes for each host

| Host | Before | After |
| --- | --- | --- |
| CLI hub (multi-workspace) | link withheld for every seat, always | link resolves when the served workspace contains the personas tree; withheld (with the code) otherwise |
| Core (multi-workspace) | same | same |
| Workspace server (single root) | link when persona ⊂ root | unchanged behaviour, decided per request |
| Playground | link only with `BORING_AGENT_WORKSPACE_ROOT=<repo>` | unchanged behaviour, decided per request |

## Proof

**Fail-without-fix (isolated).** With only `runtimeCapabilityProjection.ts` reverted to `main` and all tests kept:

```
PASS (7) FAIL (3)
  AssertionError: expected [] to deeply equal [ { filesystem: 'user', …(2) } ]
  AssertionError: expected [] to deeply equal [ ObjectContaining{…} ]
```

— i.e. exactly the three per-request behaviours fail (empty `instructionFiles` = the bug), the other 7 describe tests pass.

**Tests added/updated (both hosts covered).**
- `agentDefinition/__tests__/instructionFileRefs.test.ts` (new): same source resolves to a *different* ref per served root; withheld outside the root, with no root, on an unsafe composed path, through an escaping symlink, and for a vanished source.
- `agent-host/__tests__/describeAgent.test.ts`: one compiled spec, three served roots — ref for the workspace containing the persona, `[]` for a sibling workspace, hub-root-relative ref from the parent — plus the stable-code log assertion.
- `agentDefinition/__tests__/resolveDefaultAgentFleet.test.ts`: the **real** repository factory personas record absolute sources and address cleanly from a repo-served request, withheld from a `tmpdir()`-served one.
- `apps/workspace-playground/.../factoryAgents.test.ts`: the single-root host path, checked against both the repo root and the default playground workspace root.

**Checks** (see comment below for the final run log).

**No live hub run.** The VM was saturated (three concurrent package builds, `/tmp` inodes exhausted mid-session); I did not boot the hub and click a link, and I am not claiming I did. The hub path is covered by the per-request unit/behaviour tests above rather than by a live click.

## Final CI proof (wave-3 verification)

Fix commit `2c6bcc163` (`test(agent): import error code from canonical enum to satisfy invariants`) — the Invariants and Runtime Refactor P8 failures from the first run were caused by a literal error-code import; now imported from the canonical enum.

Run `32592231971` settled **all green, zero failures**: E2E 7m45s ✅ · Typecheck 8m32s ✅ · Unit Tests + Unit Tests Changed ✅ · Invariants 2m10s ✅ · Runtime Refactor P8 ✅ · UI Review ✅ · Reference Images & Remote Worker Smoke 8m16s ✅ · Lint ✅ · Bundle Budget / Core Bundle Budget ✅ · Dependency Age ✅ · PR Fast Summary ✅.

## Review fixes (thermo gate, round 2)

| Finding | Severity | Disposition |
| --- | --- | --- |
| `ERROR_CODES.md` assigned HTTP 500 to `AGENT_FLEET_SEAT_INSTRUCTIONS_PATH_UNPUBLISHABLE`, but describe() only logs a warning and returns success (`runtimeCapabilityProjection.ts` warns per withheld ref, projection served normally) | should-fix | **Fixed** — commit `3785847`: doc row corrected to `200 (log-only)` with an explicit "NOT an HTTP error" note; no runtime code touched. Docs-parity test green (code-name sync unaffected). |


hachej:
CI repair: all four red checks shared a single root cause — the raw error-code string `'AGENT_FLEET_SEAT_INSTRUCTIONS_PATH_UNPUBLISHABLE'` at describeAgent.test.ts:147 violated the stable-error-code-enum invariant. The Invariants job failed directly on it, and both Unit Tests jobs fail via `invariants.test.ts` (which shells out to check-invariants.sh). Runtime Refactor P8 runs the same invariant script.

Fix (2c6bcc1): import ErrorCode from shared error-codes and assert via `ErrorCode.enum.AGENT_FLEET_SEAT_INSTRUCTIONS_PATH_UNPUBLISHABLE`.

Local proof:
- `bash scripts/check-invariants.sh packages/agent` → OK
- `vitest run src/server/agent-host/__tests__/describeAgent.test.ts src/__tests__/invariants.test.ts` → 11 passed, 0 failed

PR #1370, opened 2026-08-22
fix(agent): externalize @hachej/boring-sandbox in bundling (#1361)

### LINKED ISSUE #1361: boring-sandbox is never built in CI — source-only consumption depends on tsconfig paths that esbuild also reads
## Problem

`packages/boring-sandbox` is never built in CI. `build:packages` reaches `packages/agent build` before sandbox's `dist` exists, so sandbox is effectively consumed as a **source-only** package — and the only thing making that work is a set of hand-written `paths` entries in `packages/agent/tsconfig.json` (and siblings) that redirect `@hachej/boring-sandbox/*` to source.

That makes those `paths` entries load-bearing build configuration rather than developer-convenience aliases, which is not obvious from reading them. PR #1360 attempted to replace them with a `boring-source` custom export condition and went red across 10 CI checks; see that PR for the full diagnosis.

## Why it resists the obvious fix

`customConditions` is a **tsc-only** option. The `paths` entries have a third consumer beyond tsc and vitest: **esbuild**, which auto-discovers `packages/agent/tsconfig.json` and honors `paths` but ignores `customConditions`. `packages/agent/scripts/assert-build-artifacts.mjs:160-171` runs `esbuild({ bundle: true, packages: 'bundle' })` with no tsconfig override, so removing the paths makes these five specifiers unresolvable:

```
@hachej/boring-sandbox/providers/{blaxel,bwrap,direct,node-workspace,vercel-sandbox}
```

Probing `conditions: ['boring-source']` on the esbuild side — alone and combined with `import`/`default`/`node` — does not resolve them in this pnpm layout. There is no drop-in bundler-side equivalent.

Note that `providers/runsc` and `providers/remote-worker` never had `paths` entries and do **not** fail, which confirms the mechanism.

## Ask

Make the source-only consumption explicit and bundler-independent. Either:

1. **Build `boring-sandbox` topologically** ahead of `packages/agent` in `build:packages`, so a real `dist` exists and no source redirection is needed; or
2. **Mark `@hachej/boring-sandbox/*` external** for the `assert-build-artifacts` esbuild step, so the bundler stops trying to inline it.

Option 1 is the more honest fix — it removes the special case rather than teaching one more tool about it — but it costs build time and needs the cyclic `agent`↔`boring-bash` workspace dependency checked, since that already makes build order non-deterministic.

## Why it matters beyond the cleanup

A stale local `dist` masks the problem completely: PR #1360 was green locally and red in CI purely because of one. Moving `packages/boring-sandbox/dist` aside reproduces CI byte-for-byte, including identical error offsets. Any contributor touching these tsconfigs will hit the same trap.

## Unblocks

- Reopening `chore/1240-alias-cleanup-extract` (PR #1360, branch preserved) — and once unblocked, that cleanup should be done repo-wide with aliases **derived from the exports map**, not a hand-written regex, since ~80 equivalent mappings exist across cli/workspace/full-app/automation/core/agent.
- Part of #1240's broader goal of not hand-maintaining provider wiring in five places.

hachej:
## Summary
- \`packages/agent\` depends on \`@hachej/boring-sandbox\` while sandbox declares \`@hachej/boring-agent\` as a peerDependency — a genuine workspace cycle pnpm cannot topologically order, so \`packages/boring-sandbox/dist\` never exists when the build-artifact assertion runs.
- Until now the assertion's esbuild call resolved sandbox specifiers only by accident: esbuild auto-discovers \`packages/agent/tsconfig.json\` and honors its \`paths\`, six of which redirect at sandbox *source*. That made dev aliases load-bearing build config (#1361) and is what made #1360 red in CI while green on machines with a stale sandbox dist.
- Fix: mark \`@hachej/boring-sandbox\`(+`/\*`) external in the closure analysis — matching the published shape, where sandbox is a real runtime dependency never inlined — and add a hard assertion that \`dist/core\` does not import sandbox at all, so the Fastify closure proof keeps full coverage.
- Fixture extended to pin the new sandbox-external detector (negative coverage included).

## Scope
Single file: \`packages/agent/scripts/assert-build-artifacts.mjs\`. No runtime code changes.

## Proof
- \`node --check packages/agent/scripts/assert-build-artifacts.mjs\` → OK
- CI: build-artifact assertion + full check suite on this PR

Closes #1361 (unblocks reopening the alias-cleanup PR that was blocked on it).

### DIFF
--- a/packages/agent/scripts/assert-build-artifacts.mjs
+++ b/packages/agent/scripts/assert-build-artifacts.mjs
@@ -166,7 +166,6 @@ function assertFastifyDetectorFixture() {
         imports: [
           { external: true, kind: 'import-statement', path: '@fastify/static' },
           { external: true, kind: 'dynamic-import', path: 'fastify' },
-          { external: true, kind: 'import-statement', path: '@hachej/boring-sandbox/providers/blaxel' },
         ],
         inputs: {},
       },
--- a/packages/agent/scripts/assert-build-artifacts.mjs
+++ b/packages/agent/scripts/assert-build-artifacts.mjs
@@ -171,7 +171,6 @@ function assertFastifyDetectorFixture() {
       },
     },
   }
-  // The sandbox specifier must not leak into the Fastify result.
   const actual = findFastifyClosureViolations(fixture)
   const expected = {
     externalSpecifiers: ['@fastify/static', 'fastify'],
--- a/packages/agent/scripts/assert-build-artifacts.mjs
+++ b/packages/agent/scripts/assert-build-artifacts.mjs
@@ -110,6 +110,44 @@ function isFastifySpecifier(specifier) {
     || specifier.startsWith('@fastify/')
 }
 
+// `@hachej/boring-sandbox` cannot be bundled by the closure analysis below.
+//
+// `packages/agent` depends on `@hachej/boring-sandbox`, and
+// `packages/boring-sandbox` declares `@hachej/boring-agent` as a peer
+// dependency (its published `.d.ts` files reference agent types). That is a
+// genuine two-node workspace cycle, so pnpm cannot order sandbox's build ahead
+// of agent's -- not in `pnpm -r run build`, and not in any of the
+// `pnpm --filter <pkg>... run build` invocations CI uses. When this script
+// runs, `packages/boring-sandbox/dist` reliably does not exist yet.
+//
+// Until now these specifiers resolved anyway, but only by accident: esbuild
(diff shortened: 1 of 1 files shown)


hachej:
**Review fixes (thermo gate round 2):** commit ac8e2b1d1 — sandbox is now externalized ONLY in the server closure analysis; the core analysis stays fully bundled (a bare sandbox import now fails resolution loudly at assertion time). Deleted `isSandboxSpecifier`, `findExternalSandboxSpecifiers`, their fixture, and the compensating core check, per reviewer remedy. Local esbuild probe previously confirmed core bundles without the external.

PR #1377, opened 2026-08-22
fix(agent): factory generic seat for dispatched agents (#1296)

### LINKED ISSUE #1296: Factory workspace shows a generic 'Agent' seat and an Agent nav tab alongside the real fleet
## Symptom

A workspace running the composed fleet lists **four** agents:

```
default            → labelled "Agent"
boring-triage
boring-orchestrator
boring-worker
```

The first is the legacy single-agent fallback. In a fleet workspace it is noise: an unnamed generic "Agent" sitting above the named seats, plus an "Agent" tab in the left nav that only makes sense when there is exactly one agent.

## Where it comes from

`packages/agent/src/server/agentDefinition/resolveDefaultAgentFleet.ts:85`:

```ts
return Object.freeze([...LEGACY_DEFAULT_AGENT_FLEET, ...configuredAgents])
```

The legacy default is **always prepended**, even when `BORING_AGENT_FLEET=1` composes a real fleet. Its label is hardcoded in two places:

- `packages/agent/src/server/agent-host/createAgentHost.ts:746`
- `packages/agent/src/server/agent-host/embeddedGateway.ts:254`

both as `label: 'legacyDefault' in agent ? 'Agent' : agent.definition.label`.

## Asks

1. **Left-nav "Agent" tab should appear only when there is exactly one agent.** With a fleet composed, the seats are the navigation; a separate generic tab is redundant.
2. **The legacy entry should not present itself as a peer of the real seats.** Either hide it from agent listings when a configured fleet is present, or label it `default` so it reads as the fallback identity it is, not as an agent someone authored.

## Constraint

Do **not** simply drop the `default` agentTypeId. Existing sessions are bound to it and hosts normalize `default → primary` (DECISIONS, Decision 30). This is about presentation and navigation, not removing the identity:

- keep `default` resolvable for existing sessions and API callers
- stop advertising it as a listed agent when a configured fleet exists
- if it must remain listed, label it `default`

## Note

Third UI-truthfulness defect from the same session, with #1290 (composer shows a model the session is not using) and #1295 (Stop deletes queued messages). Each is the interface asserting something the system is not doing.

hachej:
## Problem

A factory fleet of configured agents still showed a fourth entry: the legacy `default` fallback, labelled `Agent`, sitting above the seats someone actually wrote. Nothing distinguished an authored seat from the compatibility identity, while removing `default` would strand sessions already bound to it.

## Change

**Host (`packages/agent`)**
- `listAgents` marks the legacy fallback `legacy: true` and labels it `default` whenever at least one configured agent shares the runtime.
- The legacy single-agent boot remains unchanged: the lone default keeps the `Agent` label and no `legacy` marker.
- `host.describe()` remains within its declared schema; the presentation marker is exposed only by the Gateway catalog.

**Clients (`agent front` + `workspace`)**
- The marker is presentation-only, not a creation policy.
- Empty legacy fallback entries are hidden only from authored-seat chrome and are never counted as authored seats.
- The complete catalog remains available to explicit New chat selection. If the fallback owns chats, or its session source fails, its route/card stays visible.
- Fleet session sourcing keeps a source for every listed agent, including `default`, so old sessions remain listed, scoped, and addressable.
- Explicit owners are preserved exactly; split creation is not silently retargeted to another agent.

## Owner-ratified Option B disposition

The owner ratified **Option B** for this loop. Decisions 28 and 29 are preserved: legacy `agentTypeId: default` remains creation-capable in mixed fleets and fallback-only mode across the Gateway, UI, CLI, MCP, and dispatcher. Direct default creation and owner-preserving split behavior are compatible behavior, not bugs. The prior mixed-fleet Gateway rejection and history-only claims have been removed.

## Reachability guarantee

Sessions bound to `default` stay listed fleet-wide, listed under their own agentTypeId, and readable. Partial source failures keep the fallback route and error chrome visible while healthy authored sources continue loading.

## Proof

- Agent focused regressions: 5 files, 60 tests passed (Gateway compatibility, fallback-only creation, selection, dispatcher, MCP, `host.describe()`).
- Workspace focused regressions: 2 files, 132 tests passed (history reachability, partial errors, authored seat count, picker/card creation, explicit owner-preserving split).
- CLI mixed-fleet HTTP creation regression: 1 test passed.
- Typecheck: `@hachej/boring-agent`, `@hachej/boring-workspace`, and `@hachej/boring-ui-cli` passed.
- Invariants: agent, boring-bash, boring-sandbox, workspace plugin, alignment, and skill digest checks passed.
- `git diff --check`: passed.

Closes #1296

### DIFF
--- a/packages/agent/docs/AGENT_GATEWAY_V0.md
+++ b/packages/agent/docs/AGENT_GATEWAY_V0.md
@@ -72,12 +72,12 @@ Note `listAgents` takes `ListAgentsInput`, not a bare `AuthorizedAgentScope`.
 
 ### Output DTOs
 
-`AgentSummary` (`agentTypeId`, `label`, `description?`, `legacy?`,
-`definition?{version,digest}`), `AgentSessionSummary` (`ref`, `title`,
-`status`, `createdAt`, `updatedAt`), `AgentSessionPage` (`sessions`,
+`AgentSummary` (`agentTypeId`, `label`, `description?`, `pluginIds?`,
+`legacy?`, `definition?{version,digest}`), `AgentSessionSummary` (`ref`,
+`title`, `status`, `createdAt`, `updatedAt`), `AgentSessionPage` (`sessions`,
 `nextCursor?`), and the receipts (`CommandReceipt`, `AgentSendReceipt`,
 `QueueClearReceipt`, `StopReceipt`) are unchanged from §6, field for field,
-apart from `AgentSummary.legacy`.
+apart from `AgentSummary.pluginIds` and `AgentSummary.legacy`.
 
 `legacy: true` marks the `default` fallback identity, listed beside a
 configured fleet so sessions bound to it stay addressable (gh-1296). It is not
--- a/packages/agent/docs/AGENT_GATEWAY_V0.md
+++ b/packages/agent/docs/AGENT_GATEWAY_V0.md
@@ -80,16 +80,16 @@ Note `listAgents` takes `ListAgentsInput`, not a bare `AuthorizedAgentScope`.
 apart from `AgentSummary.legacy`.
 
 `legacy: true` marks the `default` fallback identity, listed beside a
-configured fleet only so the sessions bound to it stay addressable (gh-1296).
-It is not an authored seat: with configured Agents in the fleet it is labelled
-`default` rather than `Agent`, and clients keep listing and routing its
-sessions while leaving it out of seat chrome unless it owns chats or its
-history source fails. Decision 28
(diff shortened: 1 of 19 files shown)


hachej:
Superseded by #1386 after the owner removed the legacy fallback concept entirely. #1386 keeps existing `agentTypeId: default` sessions reachable by making `default` a real configured Agent and preserving transcript placement only at the storage boundary. The legacy marker, hidden-card, and fallback UI machinery in this PR must not merge.

PR #1301, opened 2026-08-14
Rolling small fixes (owner flush)

hachej:
## Summary

Rolling owner-flush branch; **never auto-merge**. This PR remains draft.

Current effective scope:

- Present Codex context-window exhaustion consistently through structured, non-retryable chat errors while keeping Pi's canonical model metadata intact.
- Leave context compaction policy entirely to Pi; this branch does not add provider-specific reserves or alter model context metadata.
- Generate the agent front CSS once from `packages/agent/tsup.config.ts`.
- Show generic **Agent** navigation only for a genuine one-agent catalog.
- Make agent-driven `exec_ui openFile` honor plural requests with persistent tabs: one successful call per requested path, in requested order; every file remains open and each call auto-focuses its tab, leaving the final file active. Direct user preview opens retain the existing replaceable-preview behavior.

## Ownership boundaries

- **Queue/Stop behavior is excluded.** PR #1368 is its sole implementation owner.
- **Configured/persisted default-agent semantics are excluded.** PR #1386 owns identity, selection, persistence, migration, routing, and fleet-default behavior.
- **Ask-user refresh behavior is excluded from this PR's effective diff.** #1325 landed on `main` and is now canonical. During the latest `main` merge, its authoritative UI-state invalidation/reconnect implementation and tests were retained; the superseded branch-local `agentData` refresh test was dropped.
- GH #1298 archive behavior is now canonical on `main` and is not branch-owned scope. GH #1283 remains outside this rolling bugfix scope.

### Persistent, auto-focused agent file tabs

- Manual rebuilt-playground/browser smoke after three agent-style UI commands confirmed:
  1. `config.json` remained open with `__workbenchPreview: false`.
  2. `schema.sql` remained open with `__workbenchPreview: false`.
  3. `utils.ts` remained open with `__workbenchPreview: false` and was the active tab/file.
- Added frontend coverage proving agent UI commands request persistent opens and three persistent file paths create three distinct tab identities.
- The live-agent regression in `packages/workspace/src/server/__tests__/openFileMultiTurn.test.ts` asserts exactly one successful `exec_ui/openFile` call per path, separate sequential messages, and requested order.
- The deterministic contract assertion in `packages/workspace/src/server/__tests__/uiTools.test.ts` requires all requested tabs to remain open with the final tab active.
- Focused workspace tests after provenance/focus hardening: **112 passed; 3 live-LLM tests skipped without provider credentials**.
[...]

PR #1386, opened 2026-08-23
fix: reconcile persisted default Agents safely (supersedes #1311)

### LINKED ISSUE #1296: Factory workspace shows a generic 'Agent' seat and an Agent nav tab alongside the real fleet

hachej:
## Summary

Canonical owner of configured and persisted Workspace default-Agent semantics, including complete removal of the former pseudo-Agent fallback.

- `agentTypeId: default` is now a normal configured `AgentHostAgentSpec` with a real definition
- every initialized Workspace must persist a real configured default Agent
- Core config, routes, signup, governance, and both stores require that non-null identity
- historical NULL rows exist only as rolling-migration input and are CAS-reconciled before Agent routes mount
- historical `default` transcripts retain their exact storage placement through a storage-only adapter
- configured fleet/policy failures fail startup; only genuinely unseated discovered packages remain nonfatal diagnostics
- every regular Agent remains visible, including the configured Workspace default

## Ownership boundaries

- **#1386 owns:** default identity, selection, persistence, startup reconciliation, Core/meta propagation, execution gating, initial UI selection, and removal of the pseudo-Agent fallback.
- **#1377 is superseded.** Its required history-reachability guarantee is retained here without its legacy marker, hidden-card, or fallback UI machinery.
- **#1301 must remove its #1296 default-identity/listing assertions.** It does not own default identity, selection, persistence, migration, or fleet membership.
- **#1368 exclusively owns Queue/Stop behavior.** This branch preserves its `queueAction: hold|resume` and Stop semantics.

## Behavior

```text
application fleet = [default] + authored Agents       // additive regular fleet
workspace.defaultAgentTypeId = one required fleet ID
fleet listing = every regular Agent, including the selected default
historical default sessions = same ID + same transcript path
invalid requested fleet/policy = startup failure
```

- Generic hosts start with the platform-owned regular `default` Agent; adding Orchestrator produces `[default, orchestrator]`, not replacement.
- Applications may configure another real fleet member, such as `boring-orchestrator`, as the Workspace default.
- Unknown persisted non-NULL IDs remain authoritative and fail closed; they are never silently rewritten.
- New Workspace creation without a valid default is rejected by the store contract.
- Plugins are ordinary Agent bindings; host-owned defaults are materialized from host provenance, never fleet cardinality or Agent identity tricks.
- Environment skill inheritance is an explicit provisioning grant on the configured Agent spec.
- Primary New-chat/Agent `+` actions collapse prior splits; only explicit split actions create or retain a split.
- Cross-Agent session selection stays pane-local: Workspace transport, plugin, and provisioning identity remain pinned to the configured default, so switching chats does not reboot the filesystem/workbench shell.

## Exact head

`75bc2c0bba`

## Proof

- Simplified from **110 files, +3,242/−1,217** to **103 files, +2,140/−887**, including the rolling-write guard, latest-main rebase, and owner-found shell regressions.
- Agent Host + definition focus: **26 files, 179 passed**.
- Core default/composition/store focus: **7 files, 195 passed**.
- Workspace server composition focus: **2 files, 64 passed**.
- Agent/Core/Workspace/CLI typechecks: **passed**.
- `pnpm lint:invariants`: **passed**.
- `pnpm check:agenthost-cutover-matrix`: **passed**.
- `git diff --check origin/main...HEAD`: **passed**.
- GitHub required checks: **17 passed, 0 failed**.

## Review history

- Initial Sol xhigh thermo: **REVISE** — production propagation, deleted-workspace meta, transient policy classification, and shutdown receipt gaps fixed.
- Second Sol xhigh: **REVISE** — deterministic access denial fixed to terminal 403.
- Removal review: **REVISE** — non-null Core/store contract, ordinary plugin/skill composition, and fail-loud fleet policy fixed.
- Focused re-review after both original blocker rounds: **CLEAN**.
- Fresh simplification thermo: **REVISE** — generic effect lifecycle ownership, singleton pseudo-modes, dual default authority, cohort inventory, and duplicate integration coverage removed.
- Independent standards/spec review before the owner’s removal amendment: **PASS**.

## Architecture alignment

- Decision 28 now records `default` as a normal configured fleet member and preserves historical transcript placement only at the storage boundary.
- Decision 29’s fail-loud fleet-validation requirement is enforced for malformed policy and every seated-Agent failure.
- The three-seat factory ruling remains three authored factory personas plus one platform-owned regular default Agent for generic hosts and historical reachability.

## Risk and rollback

- Requested malformed/missing fleet policy or any invalid seated Agent now fails startup instead of silently dropping seats.
- Callers creating Workspaces must supply a real default identity; old NULL rows are migration data only.
- Existing `default` session IDs and transcript paths are unchanged.

## Refs

- Closes #1296.
- Supersedes #1311 and #1377.
- Follow-up recovery UX for retired persisted IDs: #1402.


hachej:
## Owner review: approved direction, one follow-up filed

Reviewed with the owner. The **fail-closed** change is the right call — silently running a workspace on a substitute Agent (the old fail-open chain) hid a real misconfiguration, and a configured non-NULL value should be authoritative.

Three things that stood out as well-built:
- `classifyWorkspaceDefaultAgentTypeCohorts()` **counts without mutating**; repair is a separate **NULL-only compare-and-set**, so it can never clobber a user's stored choice.
- The backfill runs **after** the fleet is compiled and validated, so it writes a value known to exist.
- The `42P01`/`42703` guard is scoped to the **first inventory only** — narrow, and it's what lets `/health` come up before schema deployment. Every CAS or convergence error after that stays fatal.

**Follow-up, not a blocker: #1402.** Failing closed is only half the story — the owner of an affected workspace currently gets a 409 with no way forward, and the implied recovery is editing the database. It is reachable without anyone doing anything wrong: rename a seat in `fleet.yaml` or drop an Agent package and every workspace pinned to it becomes unopenable. #1402 covers surfacing that state and letting the user re-pick a default in place, while keeping this PR's guarantee that nothing rewrites a non-NULL value without an explicit choice.

Also confirmed before the owner closed **#1311**: both branches change exactly the same 11 files, with nothing in #1311 missing here.

https://claude.ai/code/session_018SqFadabn7hUwqfcs7AUbW

hachej:
Proof of work — exact head `15fcf1bcb`

What changed:
- configured Workspace defaults now resolve only to regular fleet Agents and remain visible/initially selected
- `legacyDefault` remains compatibility-only; legacy mode persists NULL
- deleted/unknown/unavailable defaults fail closed deterministically
- merged #1368 Queue/Stop behavior is preserved
- transient policy failures are sanitized/retryable; completed provider effects retain their receipts through drain

Automated verification:
- Agent Host + addressed selection: 28 files, 219 passed, 12 skipped
- Core focused: 287 passed; one unrelated 15s provisioning timeout passed immediately in isolation
- Workspace front: 95 passed; known async fleet-restore timing test remains flaky (same tree previously 96/96), no assertion weakened
- agent/core/workspace typechecks: passed
- `pnpm lint:invariants`: passed
- `git diff --check origin/main...HEAD`: passed

Review:
- Sol xhigh thermo round 1: REVISE, findings fixed
- Sol xhigh round 2: REVISE, deterministic access-denial classification fixed
- final Sol xhigh exact-head verification: CLEAN
- independent standards/spec review: PASS

Known gap:
- pre-existing Workspace async fleet-restore/catalog test flake remains; CI is authoritative for the pushed head.

hachej:
Owner amendment completed at exact head `038841bbb`:

- removed `legacyDefault` from production and test source
- `default` is a real regular configured Agent
- Core config and Workspace store-create contracts require a non-null real default
- NULL remains migration-fixture input only and is startup-reconciled
- historical `default` transcript placement is preserved only at the storage boundary
- plugin bindings are ordinary spec bindings or topology-based single-seat composition
- generated skill inheritance is an explicit provisioning grant
- missing/malformed policy and every invalid seated Agent now fail startup

Proof:
- Agent Host + definition: 326 passed, 12 skipped
- Core focused: 287 passed
- Workspace server: 60 passed
- store/config/routes: 263 passed
- all seven composition roots passed
- typechecks, invariants, route matrix, and diff check passed
- final Sol xhigh re-review: CLEAN

This supersedes #1377; only its history-reachability requirement is retained, without legacy presentation/fallback machinery.

hachej:
## Thermonuclear review — REVISE

Reviewed exact head: 1fc6a9645652524b4931c4c5e0d0de890a929925

### Ranked findings

1. **P1 — the one-shot NULL reconciliation is not safe against an overlapping legacy writer.** In `reconcileWorkspaceDefaultAgentTypes.ts:69-79`, startup performs one CAS and one convergence count. An old process can insert a legacy NULL after the final count, so the new process reports convergence and mounts routes while the row remains NULL indefinitely; an insert between the CAS and count instead aborts startup rather than retrying. Migration 0024 remains nullable and there is no recurring/transactional rollout barrier. That contradicts this PR's required invariant that every initialized Workspace durably persists a real default Agent. Please add a rollout-safe invariant (for example a compatible constraint phase, drain-aware/repeated convergence, or another durable barrier) and cover the post-count interleaving.

2. **P1 — the generic Workspace host dropped #1310's fail-closed configured-default guard.** #1310 validates `opts.defaultAgentTypeId` against the resolved fleet and throws `ConfiguredDefaultAgentError`. At this head, `createWorkspaceAgentServer.ts:1282-1328` has no equivalent guard and forwards an unknown configured ID into plugin composition at line 1327. Core correctly validates `rawConfig.defaultAgentTypeId` against `agentTypeIds` in `createCoreWorkspaceAgentServer.ts:1121-1125`; the generic Workspace seam does not. Restore the guard and its negative test.

### Consistency with #1310

**Not fully consistent.** #1386 agrees with #1310's fail-closed configured-default semantics in Core and preserves fail-loud configured fleet/policy validation, while deliberately replacing #1310's pseudo/legacy `default` Agent with a regular configured Agent. However, omission of #1310's generic Workspace `ConfiguredDefaultAgentError` check is a semantic regression, so the PR as a whole does not yet agree with #1310's configured-default contract.

### Verified vs. trusted

Verified directly: PR #1386 body/full diff at the head above; PR #1310 body/full diff at `ebb8895ccf7a9e85cdeaae1be68a021172a452f6`; Core resolver/store/reconciler/server paths; generic Workspace composition; the 42703/42P01 guard is narrowly limited to the initial inventory query; unknown persisted non-NULL IDs remain authoritative and fail closed; CAS updates only NULL rows; current GitHub checks are 17 passed / 0 failed. I found no weakened assertion concealing either issue.

Took on trust: the PR body's claimed local suite/typecheck counts and CI internals. I did not rerun tests because both blockers are control-flow/interleaving properties not resolved by the existing green suites.

hachej:
## Thermo review gate — findings resolved

Exact revision: `24b4eae51dc7fb105743ed38e3e3b406a04f7086`

1. **P1 rolling-NULL race — FIXED.** The finding was valid: a finite CAS/count sequence cannot exclude an old writer inserting NULL after the final zero count. Migration 0025 now adds `workspaces_default_agent_type_id_required_check` as `CHECK (default_agent_type_id IS NOT NULL) NOT VALID`. That preserves the legacy NULL cohort for the existing CAS reconciliation while PostgreSQL rejects every new NULL from an overlapping legacy process. The schema model carries the same invariant. The new PostgreSQL behavior test performs reconciliation, observes zero NULLs, then exercises the exact post-count legacy insert interleaving and proves it fails with `23514`.

2. **P1 generic Workspace configured-default guard — FIXED.** Restored PR #1310 semantics exactly at the pre-composition boundary: `ConfiguredDefaultAgentError`, stable `CONFIG_INVALID` code, `defaultAgentTypeId` field, public app/server export, and rejection when the configured default is missing or empty before Host creation. Existing Agent fleet coverage continues to prove invalid seated configuration fails hard while invalid unseated discovery is excluded.

Automated proof:
- `pnpm --filter @hachej/boring-workspace exec vitest run src/app/server/__tests__/createWorkspaceAgentServer.test.ts` — **62 passed**
- `pnpm --filter @hachej/boring-core exec vitest run src/app/server/__tests__/createCoreWorkspaceAgentServer.defaultAgentType.test.ts src/server/db/__tests__/defaultAgentTypeIdRequired.migration.test.ts` — **10 passed**
- `pnpm --filter @hachej/boring-agent exec vitest run src/server/agentDefinition/__tests__/loadConfiguredAgentFleet.test.ts` — **13 passed**, no type errors
- `pnpm --filter @hachej/boring-agent run typecheck` — **passed**
- `pnpm --filter @hachej/boring-workspace run typecheck` — **passed**
- `pnpm --filter @hachej/boring-core run typecheck` — **passed**
- `git diff --check` — **passed**

Per-finding disposition: **1 fixed; 2 fixed.**

Files changed: 8 — Core migration/journal/schema/reconciler plus PostgreSQL race test; Workspace server guard/public export plus behavior test.

Push status: **pushed** `24b4eae51` to `origin/weekend/pr1311-revive`. No merge performed.

hachej:
## Thermo re-verification — REVISE

Reviewed exact head: `24b4eae51dc7fb105743ed38e3e3b406a04f7086`

The two original findings are semantically closed:

- **Rolling-NULL race:** migration 0025's `CHECK (default_agent_type_id IS NOT NULL) NOT VALID` is a real write barrier, not another narrowed timing window. PostgreSQL preserves the pre-existing NULL cohort without scanning it, enforces the check on every later INSERT/UPDATE, and DDL locking prevents an overlapping legacy writer from slipping across constraint installation. The exact former post-count insert now fails with `23514`.
- **Workspace fail-closed guard:** the class, stable `CONFIG_INVALID` contract, public export, and pre-composition membership check match #1310's implementation. Invalid seated configuration fails startup; unseated discovery remains excluded/inert; selecting an excluded Agent as the configured default fails hard.

### Ranked remaining finding

1. **P1 — the new barrier leaves the touched PostgreSQL Workspace-store conformance suite red.** `packages/core/src/server/db/stores/__tests__/PostgresWorkspaceStore.test.ts:754` still manufactures a “legacy” cohort by updating `default_agent_type_id` to NULL *after all migrations have run*. Migration 0025 correctly rejects that post-barrier write, so both NULL/CAS conformance cases fail during setup with `23514`:

   ```text
   pnpm --filter @hachej/boring-core exec vitest run \
     src/server/db/stores/__tests__/PostgresWorkspaceStore.test.ts \
     -t 'NULL-default|NULL defaults'

   Test Files  1 failed (1)
   Tests       2 failed | 73 skipped (75)
   PostgresError: violates check constraint "workspaces_default_agent_type_id_required_check"
   ```

   Please make this an authentic pre-0025 fixture (or relocate/retire the obsolete post-barrier fixture while preserving store CAS coverage). Do not weaken or disable the production constraint.

Focused verification otherwise passed: Workspace 62/62; Core default/migration 10/10; Agent fleet 13/13 with no type errors; `git diff --check` clean.

**REVISE**

hachej:
## Thermo re-verification disposition — FIXED

Exact revision: `f076f0fedda06bc98cc9a1c75bb4bfcd7bca2153`

**Root-cause decision: fixture bug; migration 0025 remains unchanged.** Its `CHECK (default_agent_type_id IS NOT NULL) NOT VALID` correctly installs over genuine pre-existing NULL rows without scanning them, while enforcing the barrier on later INSERT/UPDATE writes. The conformance fixture was manufacturing legacy data after that barrier.

The PostgreSQL conformance fixture now batches the complete legacy NULL cohort while the constraint is absent, then reinstalls the identical `NOT VALID` barrier in the same transaction before exercising the store CAS behavior. PostgreSQL transactional DDL keeps the temporary barrier removal atomic. The Local fixture mirrors the plural cohort contract. No production constraint or assertion was weakened.

Reproduction at reviewed head `24b4eae51`:
- `pnpm --filter @hachej/boring-core exec vitest run src/server/db/stores/__tests__/PostgresWorkspaceStore.test.ts -t 'NULL-default|NULL defaults'` — **2 failed**, both setup writes SQLSTATE `23514`.

Proof at `f076f0fed`:
- same focused command — **2 passed**
- `pnpm --filter @hachej/boring-core exec vitest run src/server/db/stores/__tests__/LocalWorkspaceStore.test.ts src/server/db/stores/__tests__/PostgresWorkspaceStore.test.ts` — **146 passed**
- `pnpm --filter @hachej/boring-core exec vitest run src/server/db/__tests__/defaultAgentTypeIdRequired.migration.test.ts` — **1 passed**
- `pnpm --filter @hachej/boring-core run typecheck` — **passed**
- `pnpm typecheck` — **passed** across the monorepo
- `git diff --check` — **passed**
- independent thermo review of the fixture diff — **APPROVE, no findings**

Files changed:
- `packages/core/src/server/db/__tests__/storeConformance.ts`
- `packages/core/src/server/db/stores/__tests__/LocalWorkspaceStore.test.ts`
- `packages/core/src/server/db/stores/__tests__/PostgresWorkspaceStore.test.ts`

Push: `f076f0fed` pushed to `origin/weekend/pr1311-revive`.

hachej:
## Final thermo stamp — SHIP

Reviewed only fixture commit `184680fd582a8f5f0b4e5950dbebc773fc30922d` (the rebased successor of `f076f0fed`).

- **Legacy simulation is faithful.** The PostgreSQL fixture creates the complete legacy cohort in one transaction: it removes the barrier, sets the cohort to NULL, then reinstalls the identical `CHECK (default_agent_type_id IS NOT NULL) NOT VALID` constraint before commit. This yields the same observable state as migration 0025 encountering pre-existing NULL rows, while transactional DDL prevents an externally visible barrier-free window. The Local fixture mirrors the plural cohort.
- **No assertions were weakened.** No `expect`, rejection, or `assertionPassed` line changed. Inventory, NULL-only CAS, idempotence, non-NULL preservation, cross-app isolation, and concurrent single-winner assertions remain byte-for-byte intact; only fixture setup was batched before barrier reinstallation.
- **Proof:** Local + PostgreSQL Workspace-store conformance: **146/146 passed**. Current PR CI is fully green.

The original two findings are closed, and the SQLSTATE `23514` fixture regression is closed. Review chain complete.

**SHIP**

PR #1427, opened 2026-08-27
#1426 fix(workspace): keep new chats out of split panes

### LINKED ISSUE #1426: New chat should not create an additional split pane
## Problem
Creating a new chat from the active project while chats are already split adds a third split pane.

## Expected
New chat replaces the active pane; only the explicit split control creates a split pane.

## Proof
Regression coverage exercises project chat creation from a persisted two-pane layout.

hachej:
Closes #1426

## Summary
- route active-project New chat through the replacement transaction
- retain explicit split behavior only for the split control
- cover a persisted two-pane layout regression

## Validation
- `vitest run src/app/front/__tests__/WorkspaceAgentFront.test.tsx` (92 passed)
- `tsc --noEmit -p tsconfig.json`


hachej:
Superseded by #1428, which is rebased on current  and contains only this two-file fix.

hachej:
Correction: #1428 is the current PR. It is rebased on current main and contains only this two-file fix.

PR #1435, opened 2026-08-27
feat(workspace): recovery UI for unavailable default Agent (#1402)

### LINKED ISSUE #1402: Unavailable Workspace default Agent must be recoverable from the UI, not a bare 409
## Context

#1386 makes default-Agent resolution **fail closed**. When a workspace's persisted `default_agent_type_id` names an Agent that no longer exists (removed, renamed, or a fleet reshuffle), resolution now throws instead of silently substituting another Agent:

```ts
// packages/core/src/server/defaultAgentType.ts
throw new HttpError({ status: 409, code: DEFAULT_AGENT_TYPE_UNKNOWN_SEAT,
                      message: 'Workspace default Agent is unavailable' })
```

**The fail-closed decision is right.** Quietly running someone's workspace on a *different* Agent than the one they chose is worse than refusing — that was the old fail-open behavior and it hid the problem.

## The gap

Failing closed is only half of it. Today the owner of that workspace gets a **409 and no way forward**. There is no surfaced explanation and no in-product path back to a working state — the implied recovery is editing the database, which is not a recovery story.

Note this is reachable without anyone doing anything wrong: rename a seat in `fleet.yaml`, or drop an Agent package, and every workspace pinned to it becomes unopenable.

## What is needed

Surface it to the user and let them fix it:

1. **Show the state.** When `DEFAULT_AGENT_TYPE_UNKNOWN_SEAT` is raised, the workspace should render an explicit, readable state — which Agent id is missing, and that the workspace is pinned to it — rather than a failed request.
2. **Offer the fix in place.** Let the user re-pick a default Agent from the available fleet, which writes through to `workspaces.default_agent_type_id`. One action, no DB access.
3. **Keep the guarantee.** Re-picking is an explicit user choice, so it must not reintroduce silent substitution: nothing may rewrite a non-NULL persisted value without the user asking. #1386's NULL-only compare-and-set backfill already holds this line and should stay as-is.

## Acceptance

- A workspace whose persisted default names an unknown seat opens to a clear explanatory state naming the missing Agent id.
- The user can choose an available Agent from that state and the workspace becomes usable again.
- The persisted value changes **only** through that explicit choice; automated paths still touch NULL rows only.

## Provenance

Raised by the owner while reviewing #1386. #1386 itself is not blocked by this — the fail-closed change is the correct foundation and this is the follow-up that makes it humane.

https://claude.ai/code/session_018SqFadabn7hUwqfcs7AUbW

hachej:
Fixes #1402

**Stacked on #1386** (`weekend/pr1311-revive`) — this PR targets that branch, not `main`. **It must be retargeted to `main` once #1386 merges.**

## Why

#1386 makes default-Agent resolution fail closed: when a workspace's persisted `default_agent_type_id` names a seat that no longer exists (a renamed seat in `fleet.yaml`, a dropped Agent package), resolution throws instead of silently rehoming the workspace onto a different Agent. That decision is right — quiet substitution is worse than refusal. But on its own it leaves the workspace owner with a 409 and no way forward; the implied recovery was editing the database.

This is the UX half: see the state, understand it, fix it in place.

## Design decisions

**1. A diagnostic endpoint, not a parsed error.** `GET /api/v1/workspace/default-agent` deliberately does *not* call `resolveWorkspaceDefaultAgentTypeId` — that function throws precisely when the workspace is broken, which is exactly when the recovery surface needs an answer. It *reports* the state (`status`, `persistedDefaultAgentTypeId`, `availableAgents`) rather than enforcing it. The front therefore never re-implements resolution client-side, and never has to reverse-engineer a 409 body to learn what the valid seats are.

**2. The write can't reintroduce the bug.** `PUT /api/v1/workspace/default-agent` accepts only members of the same validated fleet (`agentTypeIds`) that resolution checks against — recovery can never pin another unavailable seat. Malformed ids are rejected 400 before any write; unknown-but-well-formed ids 409.

**3. The persisted value still changes only on explicit choice.** New `WorkspaceStore.setDefaultAgentTypeId(id, agentTypeId)` is the *only* write that may overwrite a non-NULL value, and it is reachable from exactly one place: the user pressing "Set as default Agent". #1386's NULL-only compare-and-set backfill is untouched and stays the automated path. A store-conformance test pins both halves: the explicit repin changes one row, and a subsequent `compareAndSetNullDefaultAgentTypeId` still returns 0 and leaves other stranded rows exactly as found.

**4. The probe is not a render gate.** `WorkspaceDefaultAgentRecoveryGate` renders its children immediately and swaps in the recovery surface only once the server has *confirmed* the broken state. Consequences, both deliberate:
- a healthy workspace pays zero boot latency and zero extra re-render for this feature;
- a probe that can't be reached (offline, proxy hiccup) degrades to the normal workspace rather than locking a user out of a workspace that is probably fine. The server, not this component, is what enforces the guarantee — the gate is an explanation, not a lock.

**5. Degrade honestly with an empty fleet.** If the deployment has no Agents at all there is nothing to pin to, so the surface says that plainly and offers Retry instead of an empty picker pretending to be actionable.

**6. Recovery reloads rather than patching cached state.** The workspace record that feeds `agentTypeId` down into `WorkspaceAgentFront` is provider-cached; after a successful repin the route re-derives from a fresh read. The gate itself only reports that the write succeeded (`onRecovered`), keeping the component reusable outside the core route.

## Surface

Uses the existing `ErrorState` from the UI kit and the workspace route's existing full-screen error shell — this is a recovery surface, not a new design language. The picker is a `radiogroup` of `min-h-[48px]` options; the confirm and Retry buttons are `min-h-[48px]` too (coarse-pointer hard gate).

## Proof

- `packages/core` typecheck: clean.
- New front behaviour tests (`WorkspaceDefaultAgentRecovery.test.tsx`, 6 passing): healthy state renders the workspace untouched; unavailable state names the missing id and lists replacements; confirm PUTs the chosen id and clears the state; empty fleet renders the honest fallback with no picker; a failed write surfaces an alert instead of pretending recovery happened; an unreachable probe does not lock the user out.
- New server tests in `createCoreWorkspaceAgentServer.defaultAgentType.test.ts` (file now 14 passing): unavailable read shape, healthy read with no write, successful repin, refusal to repin onto an unknown seat, malformed-id rejection before any write.
- Store conformance: explicit repin is scoped to one workspace and does not disturb the NULL-only backfill contract (runs against both `LocalWorkspaceStore` and `PostgresWorkspaceStore`; local suite 72 passing).
- Broader core suite (`src/app src/server/db src/server/routes src/server/__tests__`) run before and after: **baseline 12 failed / 5 files → with this change 9 failed / 4 files**. All remaining failures are pre-existing Postgres-dependent / under-load flakes (`raceConditions`, `PostgresModelBudgetStore`, and two tests in files that pass in isolation), none introduced here.

## Files

- `packages/core/src/shared/workspaceDefaultAgent.ts` (new) — route constant + transport types, shared by server and front
- `packages/core/src/app/server/createCoreWorkspaceAgentServer.ts` — GET/PUT recovery routes
- `packages/core/src/server/app/types.ts`, `LocalWorkspaceStore.ts`, `PostgresWorkspaceStore.ts` — `setDefaultAgentTypeId`
- `packages/core/src/app/front/WorkspaceDefaultAgentRecovery.tsx` (new) — gate + recovery surface
- `packages/core/src/app/front/CoreWorkspaceAgentFront.tsx` — wires the gate around the workspace route
- tests: `WorkspaceDefaultAgentRecovery.test.tsx` (new), `createCoreWorkspaceAgentServer.defaultAgentType.test.ts`, `createCoreWorkspaceAgentServer.testHarness.ts`, `storeConformance.ts`


hachej:
REVISE

Scope: reviewed only #1435 own recovery commit `7bf423f1` against its parent `9a9622a`, as requested. The current GitHub base has diverged and `gh pr diff 1435` includes inherited #1386 history; none of that inherited delta is graded here.

Ranked findings:

1. **HIGH — the PUT grants mutation authority to viewers.** `PUT /api/v1/workspace/default-agent` calls `resolveWorkspaceId`, whose production authorization checks authentication plus `workspaceStore.isMember` only (`packages/core/src/app/server/createCoreWorkspaceAgentServer.ts:792-809,1288-1293,1888-1914`). Any viewer can therefore change the workspace-wide default Agent. Canonical workspace mutation requires `editor` or higher (`packages/core/src/server/routes/workspaces.ts:154-157`; role hierarchy in `packages/core/src/server/auth/requireWorkspaceMember.ts:6-10,67-72`). Require at least editor, or owner if this is administrative policy, and add unauthenticated/nonmember/viewer/editor tests. The current harness globally defaults `isMember=true`, so this regression is untested.

2. **MEDIUM — recovery is an unconditional last-writer-wins repin, not a recovery-state transition.** The route checks only existence and candidate fleet membership, then both stores overwrite any active row by id (`createCoreWorkspaceAgentServer.ts:1888-1914`; `PostgresWorkspaceStore.ts:251-258`; `LocalWorkspaceStore.ts:129-135`). A stale recovery tab or two simultaneous confirms can overwrite a newer successful recovery, and a direct PUT can repin an already-healthy workspace. Send the observed persisted id/version and make the store update conditional (`WHERE id = ? AND default_agent_type_id = ?`); return 409 with fresh state when it changed. Add competing-confirm and healthy-state tests. The NULL backfill itself is CAS-only and completes before route mounting in this server lifecycle, so I did not find a direct in-process backfill race.

3. **MEDIUM — one failed probe can leave a genuinely broken workspace permanently on the unusable child UI for that mount.** The gate renders children before confirmation and silently ignores every non-2xx, malformed, rejected, or indefinitely pending probe (`packages/core/src/app/front/WorkspaceDefaultAgentRecovery.tsx:65-78`). That preserves access for a probably-healthy workspace, but when the workspace is actually broken the server refuses execution while the user gets neither recovery nor Retry. Preserve non-gating rendering, but expose a retry affordance or bounded re-probe after probe failure; add the broken-state/transient-probe-failure case. The initial child render is otherwise intentional, and the server-side effect admission remains fail-closed.

Assertions: no existing assertions were removed or weakened in the sole #1435 commit; it is additive. The new assertions do not cover role authorization, cross-tenant denial, stale-write conflicts, or recovery after a transient probe failure.

Verified:
- Request-scope/header mismatch and nonmembership block workspace A → workspace B; active workspace app ownership is checked before read/write.
- Candidate ids are parsed and checked against the server boot fleet before write.
- Focused new suites passed: 3 files, 92/92 tests (`WorkspaceDefaultAgentRecovery`, default-Agent server tests, LocalWorkspaceStore conformance).
- `@hachej/boring-core` typecheck passed; diff check passed.

Trusted/inherited rather than independently re-proven here:
- Production auth-hook population of `request.user`.
- The base branch fleet compiler guarantees that the boot `agents` array is the validated regular fleet.

hachej:
Disposition: **all three findings accepted and fixed** in `5237bf8f`. Branch also rebased onto `origin/main` (parent #1386 merged), so `gh pr diff 1435` is now the recovery work only — no inherited delta.

---

**1 (HIGH) — viewer mutation authority. Fixed.**

You were right that membership was the whole check. `PUT /api/v1/workspace/default-agent` now:
- refuses any `request.requestScope` outright (403 `agent_host_managed_workspace_mutation_forbidden`) — an agent-host scope is an automated caller by construction, and Decision 28 reserves automated paths for the NULL-only backfill;
- requires `request.user` (401 otherwise);
- reads `workspaceStore.getMemberRole` and refuses `viewer` (403 `forbidden`, "Requires editor role or higher"), matching the editor floor in `routes/workspaces.ts`.

The harness previously defaulted `isMember=true` and had no role mock at all; it now carries a `getMemberRole` mock so the floor is actually exercised. New tests: viewer denied, editor allowed (same request that the viewer was denied), non-member denied, unauthenticated denied — each asserting `setDefaultAgentTypeId` was never called.

**2 (MEDIUM) — last-writer-wins repin. Fixed, both halves.**

The request now carries the seat the client observed (`expectedDefaultAgentTypeId`), and:
- the store update is compare-and-set on it — `WHERE id = ? AND deleted_at IS NULL AND default_agent_type_id = ?` in Postgres, the same predicate in the local store — returning `null` when nothing matched;
- on `null` the route re-reads: missing row → 404, otherwise 409 "Workspace default Agent changed since it was read";
- the route additionally refuses when the *observed* seat is itself inside the fleet. That closes the second half you named: a healthy workspace can never be repinned through this route (its persisted seat is by definition in the fleet), and a NULL row cannot be touched either, so the endpoint cannot be repurposed as a settings write or race the backfill.

The front sends the seat it rendered and, on 409, re-probes rather than reporting success. New tests: stale repin conflicts (409, no false success), healthy-workspace repin refused with no write, store-conformance case where a second confirm presenting the old seat matches nothing and the newer recovery survives (runs against both stores).

**3 (MEDIUM) — stranded on unusable child UI after a probe failure. Fixed.**

Agreed this was the real gap: preserving access for a probably-healthy workspace is right, but when the workspace genuinely is broken the old code gave neither recovery nor retry. Now every probe failure — rejection, non-2xx, *and* a malformed body, which previously counted as "healthy" — re-probes a bounded number of times with linear backoff, and after that surfaces a non-blocking bottom bar: "This workspace's default Agent could not be verified… **Check again**" (48px target). Children still render throughout; the workspace is only probably broken at that point, so a network blip must not lock anyone out, but there is now always a way to ask again. The re-check uses a monotonic probe counter rather than resetting a retry count — resetting to a value it already held would have been a no-op React skipped, which is exactly the silent-stall shape the finding described.

New tests: bounded re-probe then explicit "Check again" revealing the broken state; a 500 probe response treated as failure rather than health.

---

**Proof**

- `@hachej/boring-core` typecheck: clean.
- Touched suites, 127/127: `WorkspaceDefaultAgentRecovery.test.tsx` 9, `createCoreWorkspaceAgentServer.defaultAgentType.test.ts` 19, `LocalWorkspaceStore.test.ts` (store conformance) 72, `CoreWorkspaceAgentFront.test.tsx` 27.
- Broader core suite (`src/app src/server/db src/server/routes src/server/__tests__`) re-baselined on this rebase: **`origin/main` alone → 9 failed / 4 files; with this branch → 9 failed / 4 files, same four**. All are pre-existing Postgres-dependent or under-load flakes (`raceConditions`, `PostgresModelBudgetStore`, plus two tests in files that pass in isolation). Nothing introduced here.
- No assertions removed or weakened; the earlier repin tests were updated to the new request shape and the coverage gaps you named — role authorization, non-member/unauthenticated denial, stale-write conflict, recovery after a transient probe failure — are now covered.

hachej:
REVISE

Findings 1 and 2 are closed:
- The PUT rejects any automated request scope before resolution/write, then requires an authenticated member with editor-or-owner role; viewers are refused.
- Local and Postgres writes now compare-and-set the observed persisted seat, and the route rejects an observed healthy fleet seat, so stale confirms cannot overwrite a newer healthy recovery.

[MEDIUM] Finding 3 remains open for the indefinitely pending probe case explicitly named in the prior review.

Evidence: `WorkspaceDefaultAgentRecovery.tsx:84-107` advances the bounded retry counter only after fetch rejection, non-2xx, or malformed data. The fetch at line 96 has no timeout or abort signal. If it never settles, `failed()` is never called, `probeFailed` never becomes true, and the `Check again` escape hatch at lines 119-124 is unreachable. The monotonic `setProbeRun(run => run + 1)` exists and works for settled failures, but does not bound a non-settling probe. Tests cover thrown failures and HTTP 500, not a never-settling promise.

Impact: a genuinely broken workspace can still remain indefinitely on unusable children with no recovery affordance, matching the unresolved portion of the original finding.

Disposition: bound each probe with a timeout/abort path and add a never-settling-fetch test that reaches `Check again` and can re-probe.

Verification: touched suites passed, 4 files / 127 tests.

hachej:
Disposition: **accepted and fixed** in `a877e0ef`. The finding was exactly right — the ladder only ever ran from a *settled* probe, so a hung connection reached none of it.

**Change**

Each probe now carries an `AbortController` bounded at `PROBE_TIMEOUT_MS = 10_000`, matching `DEFAULT_TIMEOUT` in the workspace filesystem fetch client (`packages/workspace/src/plugins/filesystemPlugin/front/data/fetchClient.ts:16`) — the existing fetch-timeout precedent in this codebase.

Two details worth naming, because the obvious implementation would have left the hole open:

1. **The timeout drives `failed()` directly**, rather than relying on the abort surfacing as a rejection. A request wedged below the abort signal may never settle either; hanging the recovery path off `controller.signal` rejecting would just move the same dependency one layer down. The abort still fires (so the connection is released), but the failure path does not wait on it.
2. **A `settled` latch** guards both paths, so a late response arriving after its probe timed out cannot double-count into the retry ladder or resurrect a stale payload.

Everything downstream is unchanged: timeout counts as one failure, the existing bounded ladder retries with backoff, and after `PROBE_RETRY_LIMIT` the non-blocking "Check again" bar appears. Children still render throughout.

**Test** — `bounds a probe that never settles and still reaches Check again`, with fake timers and a fetch that never settles *and ignores its abort signal* (so the test cannot pass by accident through the abort rejecting):

- nothing has fired at `PROBE_TIMEOUT_MS - 1` — one probe, no bar — which is the bug's shape;
- at the bound, `signals[0].aborted === true`;
- the ladder stops at exactly `PROBE_RETRY_LIMIT + 1 = 3` attempts;
- `workspace-shell` still rendered, and the `Check again` bar present;
- advancing a further 60s produces no additional probes — once it gives up, the user is in control;
- clicking `Check again` re-probes and reaches the recovery surface.

**Proof**

- Falsification check: with the bound widened (`PROBE_TIMEOUT_MS = 10_000_000`) the new test **fails**; restored, it passes. It is not vacuous.
- `@hachej/boring-core` typecheck: clean.
- Touched suites 128/128: recovery front 10, default-Agent server 19, store conformance 72, `CoreWorkspaceAgentFront` 27.
- Broader core suite (`src/app src/server/db src/server/routes src/server/__tests__`): same four pre-existing failing files as `origin/main` alone (`raceConditions`, `PostgresModelBudgetStore`, `provisioning`, `CoreWorkspaceAgentFront` — Postgres-dependent and under-load flakes). One run also tripped `defaultAgentType.test.ts` on a 30s timeout under parallel load; it passes 19/19 in isolation and did not recur on the following run. Nothing introduced here.

hachej:
SHIP

hachej:
Owner simplification applied in `cd6000f5`: the recovery surface is now **mounted reactively**, not probed for. Net **-49 lines**, with the recovery component itself at **-57** and its test file at **-100**.

## What was deleted

All of it lived in `WorkspaceDefaultAgentRecovery.tsx` purely to make a speculative per-boot request safe. With nothing speculative left, none of it is needed:

- the boot-time `GET` probe and its effect
- the bounded retry ladder (`PROBE_RETRY_LIMIT`, `PROBE_RETRY_DELAY_MS`)
- the `AbortController` timeout (`PROBE_TIMEOUT_MS`) added in the last round
- the `settled` latch guarding late responses
- the monotonic probe-run counter
- the "Check again" probe-failure notice bar

A healthy workspace now makes **no request at all** on this feature's behalf, and the previous round's whole class of problems (hung probe, retry bounds, stale latch) stops existing rather than being defended against.

## The seam

`WorkspaceBackgroundBoot` already reports terminal boot failures through `WorkspaceAgentFront`'s existing `onWorkspaceWarmupStatusChange` — the boot-error state the ruling pointed at. The one thing missing was discrimination: it threw the server's error code away and reported only a message, so one boot failure looked like any other.

- `workspacePreload.ts` gains `WorkspaceBootError` + `errorCodeFromPayload`, reusing the existing `parseWorkspaceReadyError` shapes (which already handle both `{code}` and `{error:{details:{code}}}`), and `WorkspaceWarmupStatus`'s `failed` variant gains `code?: string`.
- `CoreWorkspaceAgentFront` mounts `WorkspaceDefaultAgentRecovery` when — and only when — that code is `DEFAULT_AGENT_TYPE_UNKNOWN_SEAT`. Every other boot failure keeps the workspace's own error handling untouched.
- `/api/v1/workspace/meta`, the call that actually produces gh-1386's verdict, joins the host's **existing** boot preload set instead of being fetched separately.

The recovery page then fetches the one thing it cannot know on its own — which Agents can replace the missing seat — on **its** mount, i.e. only for a workspace the server has already refused.

## Unchanged

Recovery page UI (picker, confirm, zero-agents fallback, 48px targets), the `PUT` with its editor floor + compare-and-set + observed-seat refusal, the `GET` endpoint, and every server-side test.

## Tests

Deleted: probe, bounded-retry, timeout, and latch cases.

Added, pinning the discrimination the ruling asked for:
- recovery mounts on `{status:'failed', code:'default_agent_type_unknown_seat'}`, and the workspace front unmounts;
- a **generic boot failure** (`{status:'failed', message:'tree failed with 500'}`) leaves the workspace front mounted, shows no recovery, and — asserted explicitly — issues **no fetch at all**;
- boot preload paths include `/api/v1/workspace/meta`, proving the verdict rides an existing request.

Kept: picker listing, confirm/persist with the observed-seat body, zero-agents fallback, failed-write path, stale-repin 409 re-read. Added one honest case for the page's own fetch failing (retryable "Try again", no pretending the workspace is fine).

`WorkspaceBackgroundBoot`'s JSON-envelope failure test now pins the propagated code.

## Proof

- `@hachej/boring-core` typecheck clean; `@hachej/boring-workspace` typecheck clean (front + server projects).
- `packages/core`: `WorkspaceDefaultAgentRecovery` 6/6, `CoreWorkspaceAgentFront` 29/29, default-Agent server suite 19/19, store conformance 73/73 — **92/92** on the server+store pair, 35/35 on the front pair.
- `packages/workspace` `src/app/front`: **149/149**.

One honest note: `CoreWorkspaceAgentFront.test.tsx`'s pre-existing cold-composition test (`injects the routed workspace id…`) already carried a 15s timeout with a comment about cold Core composition; two more tests in the file pushed it over. I raised its budget to 30s — no assertion touched — and it passes with the suite at 16.5s total.

Known limitation, stated rather than hidden: boot preloads run as a `Promise.all`, so if another boot path fails first for an unrelated reason, its error wins the race and the user sees the normal boot failure instead of the recovery surface. That is a degradation to today's behavior, never a regression, and the recovery surface is reachable on reload.

hachej:
REVISE

[BLOCKER] The error discrimination is correct for a single workspace, but `bootFailureCode` is not scoped to the workspace that produced it. `WorkspaceRoute` survives `/workspace/:workspaceId` parameter changes. After workspace A reports `DEFAULT_AGENT_TYPE_UNKNOWN_SEAT`, navigating in the same mounted route to healthy workspace B retains the code from A. The recovery branch then returns before the front for B can mount and emit `preparing`, so recovery mounts and fetches for B even though B never produced the qualifying error. A healthy `/workspace/meta` response does not escape this state; the recovery component renders it as the recovery page.

Scope the stored failure as `{ workspaceId, code }` and require the id to match, or synchronously reset/key it on workspace change. Add a same-mounted-route A-to-B test proving B mounts its normal front and triggers no recovery fetch until B itself reports the exact code. The current generic-failure test covers only one workspace and therefore does not pin this lifecycle boundary.

The other requested checks are clean:
- `code?: string` is an additive failed-status field; existing overlay and agent-front consumers still work and the host callback remains chained.
- `cd6000f53` changes only nine front/front-test files; the reviewed PUT editor authz, observed-seat refusal, and Local/Postgres CAS core are unchanged from `5237bf8f`.
- The disclosed `Promise.all` limitation is accurate and acceptable: the first unrelated terminal rejection can yield the ordinary failure surface; reload retries but does not guarantee recovery if that unrelated failure deterministically wins again.
- Requested tests pass: recovery 6/6 and `CoreWorkspaceAgentFront` 29/29, 35/35 total. Supplemental warmup plumbing and consumer tests pass 17/17.

hachej:
Disposition: **accepted and fixed** in `9379333f`. The diagnosis was exact — `WorkspaceRoute` survives `:id` changes and the stored code outlived the workspace that raised it.

## Fix

The boot failure is stored as `{ workspaceId, code }` and honoured only when the id matches the routed workspace — the same scoping `WorkspaceAgentFront` already applies to its own warmup state (`{ workspaceId, status }`).

Writing the test surfaced a **second half of the same bug** that matching alone does not close: after A → B, the entry for A is still sitting there, so navigating *back* to A makes it match again and recovery mounts from the **previous visit's** verdict rather than the current boot. So the state is also reset on workspace change — **during render, not in an effect**, because an effect leaves one frame in which the stale verdict is live and the recovery branch returns before the new boot can speak. A returning user now waits for the current boot, which is the only verdict that means anything.

```
if (bootFailure.workspaceId !== workspaceId) setBootFailure({ workspaceId, code: null })
const activeBootFailureCode = bootFailure.workspaceId === workspaceId ? bootFailure.code : null
```

Both guards are load-bearing: the reset handles the return trip, the match handles the render before it.

## Test

`does not carry one workspace's recovery state onto another in the same mounted route`, all within one mounted route:

1. A reports `DEFAULT_AGENT_TYPE_UNKNOWN_SEAT` → recovery mounts, recovery fetch happens;
2. switch to B → B's normal front is mounted, no recovery, and the recovery fetch count is **unchanged** (B triggers no fetch of its own);
3. switch back to A → still no recovery and still no new fetch: the stale verdict does not return;
4. A's boot reports the code again → recovery is back.

Harness: a `useParams` override that changes the routed id **without remounting the route** — precisely the lifecycle you named. It defaults to null, so every other test uses the real params and is untouched.

## Proof

- **Falsification**: with the scoping and the reset removed, the new test fails. Not vacuous.
- `@hachej/boring-core` typecheck clean; `@hachej/boring-workspace` typecheck clean (front + server).
- `packages/core`: **147/147** across `src/app/front`, the default-Agent server suite (19/19), and store conformance (73/73). `CoreWorkspaceAgentFront` is now 30/30, `WorkspaceDefaultAgentRecovery` 6/6.
- `packages/workspace` `src/app/front`: 148/149.

Honest note on that one workspace failure: `preserves addressed panes while the Host catalog resolves asynchronously` (a chat-list skeleton timing assertion) fails on this branch — and I verified it **also fails with the three workspace files reverted to `origin/main`**, so it is pre-existing and unrelated to the `code` plumbing. My earlier 149/149 report was the lucky side of that flake; correcting it here rather than leaving the better number standing.

hachej:
SHIP

PR #1454, opened 2026-08-28
fix(core): restore last-selected workspace on sign-in

### LINKED ISSUE #1449: Sign-in always lands on the default workspace — last-selected workspace not restored
Found in deep full-app smoke 2026-08-28 (origin/main 465f521c6). Repro (2x, fresh and same-context): sign in → navigate to a non-default workspace → sign out → sign in again → app lands on the original Default workspace, not the last-selected one. Base URL / appears to resolve to a fixed/first workspace rather than a remembered selection. Expected per the standing rule: listing and selection unchanged on later login. Likely area: the post-auth workspace resolution on the / route (core front router / workspace preload).

hachej:
Fixes #1449

## Root cause
On the `/` route, `WorkspaceAuthProvider.defaultWorkspace` (packages/core/src/front/WorkspaceAuthProvider.tsx) always resolved to the workspace flagged `isDefault`, or the first one in the list, whenever there was no `:id` route param. There was no memory of which workspace the user had actually last used — so every fresh sign-in (query cache cleared, but route is `/`) landed on the default workspace regardless of what was selected before sign-out.

## Mechanism
- Persist the most recently *matched* workspace id to `localStorage` under `boring-core:last-workspace` — same lightweight per-viewer-preference pattern already used by `ThemeProvider` (`boring-core:theme`). This fires whenever the route resolves a workspace via an explicit `/workspace/:id` visit (including `WorkspaceSwitcher`'s `navigate(hrefForWorkspace(...))` calls).
- On `/` (no route id), prefer the remembered id over the default/first workspace — but only if it's still present in the freshly-fetched `workspacesQuery.data` (the user's current membership list). A stale/removed/no-longer-a-member id is simply not found in that list, so resolution falls straight through to the existing default → first → null chain. No extra network round-trip or error handling needed for the stale case.
- `AuthProvider.signOut()` only calls `queryClient.clear()` — it never touches `localStorage` — so the remembered id survives sign-out and is available on the next sign-in.
- Per AGENTS.md rule 9 ("session history is host app data"), this is scoped as a client-side UI preference (like theme), not session/transcript data, so `localStorage` is the appropriate mechanism — no server-side field needed.

## Files
- `packages/core/src/front/WorkspaceAuthProvider.tsx` — read/write helpers + `lastWorkspaceId` lookup in `defaultWorkspace` + persistence effect keyed off `routeStatus`
- `packages/core/src/front/__tests__/WorkspaceAuthProvider.test.tsx` — two new tests

## Proof
```
pnpm --filter @hachej/boring-core exec vitest run src/front/__tests__/WorkspaceAuthProvider.test.tsx
PASS (13) FAIL (0)

pnpm --filter @hachej/boring-core exec vitest run src/front
PASS (103) FAIL (0)
```
`pnpm --filter @hachej/boring-core run typecheck` — `WorkspaceAuthProvider.tsx` and the test file typecheck clean; the only reported error is a pre-existing, unrelated `TS2339` in `packages/workspace/src/app/front/WorkspaceAgentFront.tsx` (untouched by this change, present on `origin/main` HEAD `811911ec2`).

New tests:
- `restores the last-selected workspace on / after visiting it via route, surviving sign-out/in` — visits `/workspace/ws-002` (persists it), simulates sign-out/in by clearing the query client and re-rendering fresh at `/`, asserts it lands on `ws-002` again instead of the default.
- `falls back to the default workspace when the remembered id is stale (no longer a member)` — seeds `localStorage` with an id absent from the membership list, asserts `/` resolves to the default workspace cleanly.

### DIFF
--- a/packages/core/src/front/WorkspaceAuthProvider.tsx
+++ b/packages/core/src/front/WorkspaceAuthProvider.tsx
@@ -1,4 +1,4 @@
-import { createContext, useContext } from 'react'
+import { createContext, useContext, useEffect } from 'react'
 import type { ReactNode } from 'react'
 import { useQuery, useQueryClient } from '@tanstack/react-query'
 import { matchPath, useLocation, useParams } from 'react-router-dom'
@@ -82,6 +82,32 @@ function workspaceIdFromPath(
   return null
 }
 
+/** Remembers the last workspace the user landed on so sign-in restores it instead of
+ * always resolving to the default/first workspace (issue #1449). Client-side only —
+ * per rule 9, this is a UI preference, not session-history data, so localStorage (like
+ * ThemeProvider's `boring-core:theme`) is the lightest mechanism; it survives sign-out
+ * so a fresh sign-in restores it, and is validated against the user's current workspace
+ * membership list before use so a stale/removed id can never be trusted blindly. */
+const LAST_WORKSPACE_STORAGE_KEY = 'boring-core:last-workspace'
+
+function readLastWorkspaceId(): string | null {
+  if (typeof window === 'undefined') return null
+  try {
+    return localStorage.getItem(LAST_WORKSPACE_STORAGE_KEY)
+  } catch {
+    return null
+  }
+}
+
+function writeLastWorkspaceId(workspaceId: string) {
+  if (typeof window === 'undefined') return
+  try {
+    localStorage.setItem(LAST_WORKSPACE_STORAGE_KEY, workspaceId)
+  } catch {
+    // localStorage unavailable (private mode, quota, etc.) — restoration is best-effort.
+  }
+}
+
 function routeStatusFromError(workspaceId: string, error: unknown): WorkspaceRouteStatus {
   const detail = getHttpErrorDetail(error)
   if (detail.status === 404 || detail.code === 'not_found') {
@@ -116,9 +142,11 @@ export function WorkspaceAuthProvider({
     enabled: canQueryProtectedApi,
   })
 
+  const lastWorkspaceId = routeWorkspaceId === null ? readLastWorkspaceId() : null
   const defaultWorkspace =
     routeWorkspaceId === null
-      ? (workspacesQuery.data?.find((workspace) => workspace.isDefault)
+      ? (workspacesQuery.data?.find((workspace) => workspace.id === lastWorkspaceId)
+        ?? workspacesQuery.data?.find((workspace) => workspace.isDefault)
         ?? workspacesQuery.data?.[0]
         ?? null)
       : null
@@ -149,6 +177,11 @@ export function WorkspaceAuthProvider({
     if (workspace?.id === routeWorkspaceId) return { status: 'matched', workspaceId: routeWorkspaceId, workspace }
     return { status: 'mismatched', workspaceId: routeWorkspaceId, currentWorkspaceId: workspace?.id ?? null }
   })()
+
+  const matchedWorkspaceId = routeStatus.status === 'matched' ? routeStatus.workspaceId : null
+  useEffect(() => {
+    if (matchedWorkspaceId) writeLastWorkspaceId(matchedWorkspaceId)
+  }, [matchedWorkspaceId])
 
   return (
     <WorkspaceContext.Provider value={{ workspace, role, routeStatus }}>
--- a/packages/core/src/front/WorkspaceAuthProvider.tsx
+++ b/packages/core/src/front/WorkspaceAuthProvider.tsx
@@ -87,22 +87,31 @@ function workspaceIdFromPath(
  * per rule 9, this is a UI preference, not session-history data, so localStorage (like
  * ThemeProvider's `boring-core:theme`) is the lightest mechanism; it survives sign-out
  * so a fresh sign-in restores it, and is validated against the user's current workspace
- * membership list before use so a stale/removed id can never be trusted blindly. */
-const LAST_WORKSPACE_STORAGE_KEY = 'boring-core:last-workspace'
+ * membership list before use so a stale/removed id can never be trusted blindly.
+ *
+ * Scoped per authenticated user id (`boring-core:last-workspace:<userId>`) — a shared
+ * browser must never steer user B to the workspace user A last selected, even when B is
+ * also a member of it. Without a signed-in user id there is nothing to key on, so reads
+ * and writes are simply no-ops. */
+const LAST_WORKSPACE_STORAGE_PREFIX = 'boring-core:last-workspace:'
 
-function readLastWorkspaceId(): string | null {
-  if (typeof window === 'undefined') return null
+function lastWorkspaceStorageKey(userId: string): string {
+  return `${LAST_WORKSPACE_STORAGE_PREFIX}${userId}`
+}
+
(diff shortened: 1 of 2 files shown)


hachej:
REVISE

Blocking: scope the remembered workspace to the authenticated user.

`LAST_WORKSPACE_STORAGE_KEY` is browser-global, and the `/` resolution accepts its value whenever that workspace appears in the current membership list. On a shared browser, user A can select shared workspace X, sign out, then user B (also a member of X) is steered to the preference of user A instead of the remembered/default workspace of user B. The membership check prevents unauthorized access, but it does not isolate per-user preferences. `user.id` is already available here; include it in the storage boundary (key or stored payload), and include user identity in the persistence effect dependencies.

The new “surviving sign-out/in” test does not exercise sign-out, sign-in, or an identity transition: it unmounts, clears a QueryClient, and remounts with the same mocked user. It proves persistence across remount/cache clear, which is useful, but it cannot catch the shared-browser defect. Add an A → B test where both users belong to X and verify B gets the remembered workspace of B or the default; keep or rename the same-user cache-clear test so its claim matches its proof.

The persistence effect itself looks sound: it writes only after an explicit route detail successfully reaches `matched`, depends on a stable scalar ID, and does not persist loading/error/mismatch states. Repeated or StrictMode writes are idempotent, so I found no thrash blocker.

Proof: `pnpm --filter @hachej/boring-core exec vitest run src/front/__tests__/WorkspaceAuthProvider.test.tsx` — 1 file passed, 13/13 tests passed.

hachej:
Addressed in 45328c921.

**Disposition: fixed, both findings.**

1. Storage key is now scoped per authenticated user id: `boring-core:last-workspace:<userId>` (via `session.data.user.id`, already read into `WorkspaceAuthProvider` as `user`). `readLastWorkspaceId`/`writeLastWorkspaceId` take `userId` and are no-ops without one. The persistence effect now depends on `[userId, matchedWorkspaceId]`. Because each user only ever touches their own key, an identity change is inherently isolated — no explicit cross-user clear is needed (and actively clearing would be wrong: it would discard user A's preference for when A returns).

2. Added two identity-transition tests to `packages/core/src/front/__tests__/WorkspaceAuthProvider.test.tsx`:
   - `does not steer a second user to the first user's remembered workspace on a shared browser, even when both are members of it` — user A selects shared workspace WS_2 (recorded under A's key); user B signs in on the same browser/localStorage, is also a member of WS_2, but lands on their own default WS_1, not WS_2.
   - `does not steer a second user (not a member of the first user's workspace) to a forbidden workspace` — same setup, but B's workspace list doesn't include WS_2 at all; B still lands on their own default.
   
   Kept the existing same-user cache-clear test (renamed nothing, its claim already matched: persistence across remount/query-client-clear for one user) and updated it + the stale-id fallback test to assert against the new per-user key.

Proof:
```
pnpm --filter @hachej/boring-core exec vitest run src/front/__tests__/WorkspaceAuthProvider.test.tsx
PASS (15) FAIL (0)

pnpm --filter @hachej/boring-core exec vitest run src/front
PASS (105) FAIL (0)

pnpm --filter @hachej/boring-core run typecheck
# WorkspaceAuthProvider.tsx and its test file typecheck clean; only reported
# error is the same pre-existing, unrelated TS2339 in
# packages/workspace/src/app/front/WorkspaceAgentFront.tsx (untouched by
# this PR, present on origin/main).
```

hachej:
REVISE

The storage boundary itself is now user-scoped correctly: `userId` comes from normalized `session.data.user.id` (the Better Auth user id), and both reads and writes hard no-op when it is absent, so an undefined/pending transition frame cannot fall back to a shared key.

The identity-transition test blocker remains. Both added tests unmount user A, clear A's QueryClient, change `mockSessionState.current` while no provider is mounted, then mount a separate user-B tree with a fresh QueryClient. That proves isolated A and B mounts use different localStorage keys, but it does not exercise an A -> unauthenticated/pending -> B identity change in the same mounted `WorkspaceAuthProvider`. In particular, it cannot catch the `[userId, matchedWorkspaceId]` persistence effect running against existing route/query state during the transition.

Add at least one same-provider transition test (rerender/reactive session update, retaining the relevant mounted state) and assert that B's key is never populated from A's matched workspace before B resolves to their own default. Keep the both-members case because that is the preference-isolation regression seam.

Proof rerun: `pnpm --filter @hachej/boring-core exec vitest run src/front/__tests__/WorkspaceAuthProvider.test.tsx` — 1 file passed, 15/15 tests passed.

hachej:
Addressed in e2bea91cb.

**Disposition: fixed — did not fake it; confirmed a real in-place transition is representative and tested it.**

Confirmed the real app never remounts `WorkspaceAuthProvider` on sign-in/out (so a same-tree transition test is the right thing to write, not an approximation):

- `AuthProvider`'s `client` is a stable `useMemo` keyed only on `baseURL` (`packages/core/src/front/auth/AuthProvider.tsx:46`), and `useSession()` calls `client.useSession()` — better-auth's reactive, store-backed hook from `better-auth/react`'s `createAuthClient(...)` (`packages/core/src/front/auth/authClient.ts:1,13-18`). A real sign-out/sign-in updates the session value in place via that store; it does not remount `AuthProvider` or anything below it.
- `WorkspaceAuthProvider` itself is mounted **once**, above `<Routes>`, in `CoreFront.tsx` — not nested per-route inside a `<Route element>`. That's exactly why the component parses `location.pathname` itself via `workspaceIdFromPath` rather than relying solely on `useParams` (which only populates inside a matched `<Route>`). So route navigation re-renders it through `useLocation`; it doesn't remount it either.

Added a `renderTopLevelProvider` test harness that mirrors that real composition (`WorkspaceAuthProvider` wrapping a `useNavigate`-exposing probe component directly, no per-route `<Routes>`/`<Route>` nesting), and two new tests in `packages/core/src/front/__tests__/WorkspaceAuthProvider.test.tsx`:

- `in-place identity change (same mounted tree, no unmount): B lands on B's default, never leaked from A, when B is also a member of A's remembered workspace`
- `in-place identity change (same mounted tree, no unmount): B lands on B's default, never leaked from A, when B is NOT a member of A's remembered workspace`

Both: mount once at `/workspace/ws-002` as user A (recording A's preference), then — with **no `.unmount()` call anywhere in the test** — mutate the mocked session to user B, register B's mocked workspace responses, clear the SAME `QueryClient` instance (exactly what `AuthProvider.signOut()` does — it never creates a new client), and `navigate('/')` within that same mounted tree to force the re-render that observes the new session. Both assert B lands on B's own default, A's key (`boring-core:last-workspace:user-1`) is untouched, and B's key never picks up A's workspace id.

Proof:
```
pnpm --filter @hachej/boring-core exec vitest run src/front/__tests__/WorkspaceAuthProvider.test.tsx
PASS (17) FAIL (0)

pnpm --filter @hachej/boring-core exec vitest run src/front
PASS (107) FAIL (0)

pnpm --filter @hachej/boring-core run typecheck
# WorkspaceAuthProvider.tsx and its test file typecheck clean; only reported
# error is the same pre-existing, unrelated TS2339 in
# packages/workspace/src/app/front/WorkspaceAgentFront.tsx (untouched by
# this PR, present on origin/main).
```

hachej:
SHIP

PR #1455, opened 2026-08-28
#1450 fix(core): carry invite context through the sign-in/sign-up handoff

### LINKED ISSUE #1450: Invite context lost when invitee signs up: SignInPage's Sign-up link drops redirect/invite_token → unwanted default workspace created
Found in deep full-app smoke 2026-08-28. Repro: invite a no-account email → invitee opens invite link → 'Sign in to accept this invite' → Sign in page → clicks the Sign up link (bare <a href={routes.signup}> at packages/core/src/front/auth/SignInPage.tsx:122, no redirect/invite_token forwarded) → signs up normally → postSignupHook auto-creates a Default workspace (invite context gone). DB confirms the invitee ends with BOTH an unwanted owned 'Default workspace' AND the invited workspace (editor). The invite-accept flow itself works when the account exists. Fix: forward the current redirect/invite query params through the Sign-up link (and verify the signup completion path routes back to invite acceptance, which then correctly skips default-workspace creation).

hachej:
## Summary
- Fixes #1450: invite context (redirect/invite_token) was dropped by the bare `<a href={routes.signup}>` Sign-up link on `SignInPage`, and by the reverse Sign-in link on `SignUpPage`.
- Invitee-without-account repro: invite link -> InviteAcceptPage -> "Sign in to accept this invite" -> `/auth/signin?redirect=/invites/:token` -> Sign up link dropped everything -> signup had no `invite_token`, so `postSignupHook` never received `x-invite-token` and created an unwanted "Default workspace" alongside the invited workspace.
- `withForwardedAuthParams`/`extractInviteTokenFromRedirect` (packages/core/src/front/utils.ts) forward `redirect` and `invite_token` across the sign-in <-> sign-up links, and recover the invite token from a `redirect=/invites/:token` path shape (what `InviteAcceptPage` actually produces) when `invite_token` wasn't already present.
- `SignUpPage` now derives `inviteToken` from either source before calling `signUp.email`, so the `x-invite-token` header — and therefore `postSignupHook`'s default-workspace skip (already covered by existing server tests) — reliably fires for this flow.

## Test plan
- [x] `packages/core/src/front/__tests__/SignInPage.test.tsx` — new tests: Sign-up link forwards `redirect`/`invite_token`; derives `invite_token` from a `redirect=/invites/:token` path.
- [x] `packages/core/src/front/__tests__/SignUpPage.test.tsx` — new tests: derives `invite_token` from `redirect` and forwards it as the `x-invite-token` header on signup; Sign-in link forwards `redirect`/`invite_token`.
- [x] `pnpm exec vitest run src/front/__tests__/SignInPage.test.tsx src/front/__tests__/SignUpPage.test.tsx` — 24 passed.
- [x] `pnpm exec vitest run` (full core suite) — same 49 pre-existing failures with and without this change (Postgres-integration/env-dependent, unrelated files); confirmed via `git stash`/`stash pop` A-B comparison.
- [x] `tsc --noEmit` — same single pre-existing error in `packages/workspace/.../WorkspaceAgentFront.tsx` with and without this change; no new errors.

Fixes #1450

### DIFF
--- a/packages/core/src/front/__tests__/SignInPage.test.tsx
+++ b/packages/core/src/front/__tests__/SignInPage.test.tsx
@@ -215,12 +215,16 @@ describe('SignInPage', () => {
   it(
     'derives invite_token for the Sign-up link from an invite-accept redirect path',
     withTaskId(EMAIL_AUTH_TASK_ID, async ({ assertionPassed }) => {
-      window.history.pushState({}, '', '/auth/signin?redirect=%2Finvites%2Ftok-abc')
+      window.history.pushState(
+        {},
+        '',
+        `/auth/signin?redirect=${encodeURIComponent('/invites/3fa85f64-5717-4562-b3fc-2c963f66afa6')}`,
+      )
 
       render(<SignInPage />, { wrapper: Wrapper })
 
       expect(screen.getByText(/sign up/i).closest('a')?.getAttribute('href')).toBe(
-        '/auth/signup?redirect=%2Finvites%2Ftok-abc&invite_token=tok-abc',
+        `/auth/signup?redirect=${encodeURIComponent('/invites/3fa85f64-5717-4562-b3fc-2c963f66afa6')}&invite_token=3fa85f64-5717-4562-b3fc-2c963f66afa6`,
       )
       assertionPassed('signin-derives-invite-token')
     }),
--- a/packages/core/src/front/__tests__/SignInPage.test.tsx
+++ b/packages/core/src/front/__tests__/SignInPage.test.tsx
@@ -195,6 +195,38 @@ describe('SignInPage', () => {
   )
 
   it(
+    'forwards redirect and invite_token query params through the Sign-up link',
+    withTaskId(EMAIL_AUTH_TASK_ID, async ({ assertionPassed }) => {
+      window.history.pushState(
+        {},
+        '',
+        '/auth/signin?redirect=%2Finvites%2Ftok-123&invite_token=tok-123',
+      )
+
+      render(<SignInPage />, { wrapper: Wrapper })
+
+      expect(screen.getByText(/sign up/i).closest('a')?.getAttribute('href')).toBe(
+        '/auth/signup?redirect=%2Finvites%2Ftok-123&invite_token=tok-123',
+      )
+      assertionPassed('signin-forwards-invite-context')
+    }),
+  )
+
+  it(
+    'derives invite_token for the Sign-up link from an invite-accept redirect path',
+    withTaskId(EMAIL_AUTH_TASK_ID, async ({ assertionPassed }) => {
+      window.history.pushState({}, '', '/auth/signin?redirect=%2Finvites%2Ftok-abc')
+
+      render(<SignInPage />, { wrapper: Wrapper })
+
+      expect(screen.getByText(/sign up/i).closest('a')?.getAttribute('href')).toBe(
+        '/auth/signup?redirect=%2Finvites%2Ftok-abc&invite_token=tok-abc',
+      )
+      assertionPassed('signin-derives-invite-token')
+    }),
+  )
+
+  it(
     'shows a helpful error when Google redirects back with an OAuth error',
     withTaskId(GOOGLE_AUTH_TASK_ID, async ({ assertionPassed }) => {
       window.history.pushState({}, '', '/auth/signin?error=access_denied&error_description=cancelled')
--- a/packages/core/src/front/__tests__/SignUpPage.test.tsx
+++ b/packages/core/src/front/__tests__/SignUpPage.test.tsx
@@ -196,6 +196,52 @@ describe('SignUpPage', () => {
   )
 
   it(
+    'derives invite_token from a redirect=/invites/:token param and hides Google sign-up',
+    withTaskId(GOOGLE_AUTH_TASK_ID, async ({ assertionPassed }) => {
+      mockUseOptionalConfig.mockReturnValue({ features: { googleOauth: true } })
+      mockSignUpEmail.mockResolvedValue({ data: { user: { id: 'u1' } }, error: null })
+      window.history.pushState({}, '', '/auth/signup?redirect=%2Finvites%2Ftok-xyz')
+
+      render(<SignUpPage />, { wrapper: Wrapper })
+
+      expect(screen.queryByRole('button', { name: /continue with google/i })).toBeNull()
+
+      const user = userEvent.setup()
+      await user.type(screen.getByLabelText(/name/i), 'Invited')
+      await user.type(screen.getByLabelText(/email/i), 'invited@example.com')
+      await user.type(screen.getByLabelText(/password/i), 'secret12345')
+      await user.click(screen.getByRole('button', { name: /sign up/i }))
+
+      await waitFor(() =>
+        expect(mockSignUpEmail).toHaveBeenCalledWith(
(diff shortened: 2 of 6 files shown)


hachej:
REVISE

Blocking: `extractInviteTokenFromRedirect` accepts an invite-looking prefix instead of the complete invite route. `packages/core/src/front/utils.ts:162-177` turns `/invites/good/trailing/unrelated` into `good`, decodes `/invites/tok%2Fseparator` into `tok/separator`, and accepts malformed percent escapes verbatim. That violates the non-fabrication boundary and can feed a token derived from an unrelated redirect into `x-invite-token`. Production tokens are UUID/base64url, so require the exact route shape and valid token alphabet, and return null on decode failure.

`withForwardedAuthParams` itself is sound against parameter injection and double-encoding: `URLSearchParams` reads each value once and serializes it once. Absolute and protocol-relative redirects do not produce an invite token, and `AuthGate` separately rejects them as redirect destinations. Forwarding `invite_token` in the auth URLs does not materially expand secret exposure: the same bearer token is already present in the emailed invite URL, invite path, and encoded `redirect` value; this duplicates it within the same visible URL/history/referrer/log surface.

Test honesty: the two requested files pass (24/24), but they cover happy paths only. The sign-up -> sign-in test starts with both `redirect` and `invite_token`, so it does not prove redirect-only derivation in the reverse direction. Add focused cases for trailing segments/unrelated paths, absolute/protocol-relative redirects, encoded separators, malformed escapes, encoded parameter separators/double-encoding, and redirect-only forwarding in both directions.

Reviewed head: `a7c6f1d28919815c3858bed25dc5701698a175c4`. Validation: `pnpm exec vitest run src/front/__tests__/SignInPage.test.tsx src/front/__tests__/SignUpPage.test.tsx` — 2 files passed, 24 tests passed.

hachej:
Disposition: FIXED — pushed to `fix/1450-invite-signup-handoff` at ef4df01cb.

`extractInviteTokenFromRedirect` (packages/core/src/front/utils.ts) is now anchored end-to-end:

- `INVITE_ACCEPT_PATH_RE = /^\/invites\/([^/?#]+)$/` requires the *entire* redirect to be exactly one `/invites/:segment` — no leading `//`, no trailing/unrelated segments, no `?`/`#` smuggled into the captured segment. `/invites/good/trailing/unrelated`, `//evil.example/invites/x`, `https://evil.example/invites/x`, and `/invites/x?y=1` all now fail to match at all (return `null`), rather than being trimmed to a plausible prefix.
- The captured segment is `decodeURIComponent`'d exactly once; a throw (malformed escape) now returns `null` instead of falling back to the raw/undecoded text.
- The decoded result is validated against the invite token's actual server-generated shape — pinned by reading both stores: `LocalWorkspaceStore.createInvite` uses `randomUUID()` (v4 UUID), `PostgresWorkspaceStore.createInvite` uses `randomBytes(32).toString('base64url')` (43-char unpadded base64url). `INVITE_TOKEN_RE` accepts only those two exact shapes. `/invites/tok%2Fseparator` decodes to `tok/separator`, which contains `/` — not in either alphabet — so it's rejected; a double-encoded `%252F` decodes once to the literal `%2F`, which contains `%` — also rejected without a second decode pass.

Tests (all new, all passing):
- `packages/core/src/front/__tests__/utils.test.ts`: table-driven hostile-input suite for `extractInviteTokenFromRedirect` — trailing/unrelated segments, absolute + protocol-relative URLs, encoded/double-encoded separators, malformed percent-escapes, query/hash smuggling, wrong alphabet, wrong length, whitespace padding, non-invite redirects (20 hostile cases, each asserted `null`) — plus UUID/base64url happy paths and an encoded-but-valid round-trip. Also covers `withForwardedAuthParams` redirect-only vs. derived-`invite_token` forwarding in both directions.
- `SignInPage.test.tsx` / `SignUpPage.test.tsx`: the redirect-derivation tests were switched from placeholder strings (`tok-abc`, `tok-xyz` — which the *old* regex accepted but aren't real token shapes) to a real UUID; added a redirect-only SignUp→SignIn reverse-direction test (previously only the redirect+invite_token-both-present case was covered, which doesn't exercise derivation); added a component-level hostile-redirect case proving `signUp.email` is called with `undefined` (no `x-invite-token` header) and Google sign-up stays visible when the redirect has a trailing segment.

Proof:
- `pnpm exec vitest run src/front/__tests__/SignInPage.test.tsx src/front/__tests__/SignUpPage.test.tsx src/front/__tests__/utils.test.ts` → `PASS (77) FAIL (0)`.
- `pnpm exec vitest run` (full core suite) → 49 failed / 1351 passed, same 49 pre-existing failures as on the prior head (Postgres-integration/env-dependent + an unrelated `vi.mock` hoisting issue in `CoreWorkspaceAgentFront.test.tsx`), confirmed by diffing the failing-test-name list against the previous run. No new failures.
- `tsc --noEmit`: same single pre-existing error in `packages/workspace/.../WorkspaceAgentFront.tsx` (unrelated file), present with and without this change.

hachej:
REVISE

Blocking: the claimed exact v4 UUID validation is still broader than the real LocalWorkspaceStore generator. `packages/core/src/front/utils.ts:168-174` accepts any lowercase UUID layout, so `/invites/3fa85f64-5717-1562-73fc-2c963f66afa6` is extracted and forwarded verbatim even though `randomUUID()` can only emit version 4 with an RFC variant nibble. The extractor can therefore still fabricate an impossible Local invite token from a redirect. Constrain the UUID branch to version nibble `4` and variant `[89ab]`, and add hostile regression cases for wrong version/variant.

The other hardening holds under direct adversarial probes: NFC/NFD Unicode lookalikes, overlong UTF-8, encoded/literal backslashes, route/token case tricks, Unicode slash variants, and a valid 43-character base64url token with encoded LF/CRLF appended all returned `null`. Reverse-direction coverage is now adequate.

Reviewed head: `ef4df01cbb79f75a87158b5734b76bf4a6e55295`. Validation: `pnpm exec vitest run src/front/__tests__/SignInPage.test.tsx src/front/__tests__/SignUpPage.test.tsx src/front/__tests__/utils.test.ts` — 3 files passed, 77 tests passed.

hachej:
Disposition: FIXED — pushed to `fix/1450-invite-signup-handoff` at cde3b2266.

`INVITE_TOKEN_RE`'s UUID branch (packages/core/src/front/utils.ts) is now pinned to exactly what `randomUUID()` can emit:

```
[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}
```

- Version group: `4[0-9a-f]{3}` — the 13th hex digit is fixed to `4`.
- Variant group: `[89ab][0-9a-f]{3}` — the 17th hex digit must be the RFC 4122 variant nibble.

`/invites/3fa85f64-5717-1562-73fc-2c963f66afa6` (the flagged example, version nibble `1`) and any other UUID-shaped-but-impossible layout now fail the alphabet check and return `null`.

Tests added (`utils.test.ts`): 4 hostile regression cases — wrong version nibble (`1`, `5`) and wrong variant nibble (`7`, `c`), each with an otherwise-well-formed UUID — all asserted to return `null`.

Proof:
- `pnpm exec vitest run src/front/__tests__/SignInPage.test.tsx src/front/__tests__/SignUpPage.test.tsx src/front/__tests__/utils.test.ts` → `PASS (81) FAIL (0)`.
- `pnpm exec tsc --noEmit -p tsconfig.json` (packages/core) → clean, no errors.

hachej:
SHIP

PR #1431, opened 2026-08-27
fix(ui-review): deterministic mobile capture + sidebar readiness (#1390)

hachej:
## Root causes and mechanisms

- **Mobile/tablet hard-gate nondeterminism:** screenshot capture and DOM hard-gate collection could observe different animation/layout frames. The harness now freezes CSS motion, waits for fonts plus two animation frames, and keeps that freeze active across both screenshot and hard-gate collection before cleaning it up.
- **Terminal empty agent roster:** the app client performs one-shot agent discovery; a rejected initial discovery cannot recover by waiting longer. Sidebar readiness now spends an initial bounded wait, reloads once, then retries readiness so discovery/hydration gets one clean restart.
- **Pointer-model mismatch:** compact width and coarse-pointer capability are separate signals. Registered Playwright mobile capture opts into touch explicitly. Bombadil cannot emulate pointer capability, and its CI browser may report a coarse pointer even at desktop width, so the touch-target property was removed from Bombadil rather than inferred from the host browser. The command-palette hard-gate contract/spec revision is bumped accordingly.
- **Bombadil replay divergence:** opening app navigation reveals the compact `Search⌘K` trigger asynchronously. The action policy now emits one settle `Wait` after the navigation-open click before allowing the revealed trigger, keeping generated and replayed action sets aligned.

## App-fix dependency

The deterministic harness correctly exposed the genuine 20px `Archived` control on coarse-pointer sidebar states. That app defect was fixed and merged to `main` by #1430 (closing #1429); this branch is based on that merge. No app code is changed here.

## Proof

- `pnpm --filter @hachej/boring-ui-review-tools test` — 95/95 passed.
- `pnpm --filter @hachej/boring-ui-review-tools typecheck` — passed with no errors.
- Exact consecutive review command: `pnpm --filter @hachej/boring-ui-review-tools ui:review -- review workspace-agent-sidebar --critic=fixture`
  - Run 1: PASS — 1 passed (1.1m)
  - Run 2: PASS — 1 passed (1.1m)
  - Run 3: PASS — 1 passed (1.2m)
  - Run 4: PASS — 1 passed (1.7m)
  - Run 5: PASS — 1 passed (1.1m)
  - Result: **5/5 consecutive green**
- Thermonuclear maintainability review: clean; no material findings.
- GitHub CI run `33101687058`: all required checks passed, including UI Review.

Fixes #1390

PR #1456, opened 2026-08-28
#1453 Restore archived sessions

### LINKED ISSUE #1453: No unarchive/restore affordance for archived sessions
Deep playground smoke 2026-08-28: archiving works (#1376), and the Archived disclosure lists archived sessions when non-empty (#1430) — but there is no visible affordance to RESTORE an archived session to the active list. Archive is currently a one-way door in the UI. (The deep-smoke run also saw the Archived section fail to appear immediately after archiving in one state — possibly refresh-timing; the basic smoke saw it appear correctly — worth checking the refresh path while adding restore.)

hachej:
Fixes #1453.

## The gap

Archive (#1376) shipped a symmetric server op and the Archived disclosure (#1430) shipped the place a restored chat comes back from — yet the deep smoke found no way back. The reason is not a missing action, it is a section that never renders.

`useAddressedFleetSessions` computed `archiveCapable = controllers.length === agents.length && controllers.every(c => c.setArchived && c.loadArchived)`. `every` over an empty list is `true`, so while Agent discovery is still running (`agents: []`, zero controllers) the fleet claimed full archive support and handed `AppLeftPane` a `loadArchived` that resolves without loading anything. The pane probes for archived chats **exactly once** (the #1429 latch, deliberately never re-armed so a failing API cannot become a fetch loop) — it latched on that no-op, and once the real controllers arrived it never asked again. Archived inventory stayed unfetched, `archivedSessions.length` stayed 0, the Archived disclosure stayed hidden, and with it the only row that carries the restore action. That is also the refresh-timing symptom noted in the issue: the section appeared only when the archive mutation itself pushed a row into the pager, and not otherwise.

## Changes

- `addressedFleetSessions.tsx` — archive capability is a capability, not a vacuous truth: `agents.length > 0 && controllers.length === agents.length && every(...)`. Until a real controller exists the fleet exposes no `loadArchived`/`setArchived`, so the pane's probe does not latch (its guard already skips an absent handler) and fires for real the moment the capability lands.
- `AppSessionActionsMenu.tsx` — the way back is named for what it does to the chat: **"Restore session"** (was "Unarchive session"), with the same `ArchiveRestore` icon, in the same kebab, in the same slot as Archive. "Unarchive" only reads as the inverse of a verb the user has to remember using; "Restore" reads as a return to the list. Placement, menu component and the 48px coarse-pointer slot are untouched — this is the Archive idiom run backwards, not a new control.

Delete stays demoted to archived rows only (#1376 owner decision), so the destructive action remains one deliberate step past restore.

## Server op

Already existed and is already symmetric — `POST /api/v1/agents/:id/sessions/:sessionId/archive` with `{ archived: boolean }`, `SessionStore.setArchived`, sidecar marker dropped on restore. No server change was needed.

## Tests

- `packages/workspace/src/app/front/__tests__/addressedFleetSessions.test.tsx` (new): no archive API is advertised before any Agent controller exists; once one does, restore delegates to the owning controller with its `agentTypeId`.
- `AppLeftPane.test.tsx` (new case): archived chat is restored from its own row menu — `onSetSessionArchived(id, false)` fires, the row leaves the Archived section, the count drops to zero so the disclosure disappears, and the chat is back in the active list. No reload.
- Existing archive coverage updated to the new label.

## Proof

- `packages/workspace` — AppLeftPane 42/42, AppLeftPaneSessionRow + addressedFleetSessions 16/16, WorkspaceAgentFront 99/99 (one load-starvation timeout on a first cold run; clean on rerun).
- `packages/agent` — `sessions.archive.test.ts` + `httpProjection.test.ts` 19/19 (server archive→restore round trip), front session suites 79/79.
- `tsc --noEmit` clean for both `packages/workspace` and `packages/agent`.

### DIFF
--- a/packages/workspace/src/app/front/WorkspaceAgentFront.tsx
+++ b/packages/workspace/src/app/front/WorkspaceAgentFront.tsx
@@ -2813,6 +2813,10 @@ export function WorkspaceAgentFront<
           archivedLoading={sessionApi?.archivedLoading}
           hasMoreArchived={sessionApi?.hasMoreArchived}
           onLoadArchived={sessionApi?.loadArchived}
+          // The cohort that owns those archived rows. When the source changes
+          // the pane is looking at a different inventory and probes it once
+          // more, instead of trusting a latch from the previous one (#1453).
+          archivedInventoryKey={sessionApi?.sourceIdentity ?? undefined}
         />
       )}
     >
--- a/packages/workspace/src/app/front/__tests__/addressedFleetSessions.test.tsx
+++ b/packages/workspace/src/app/front/__tests__/addressedFleetSessions.test.tsx
@@ -1,0 +1,74 @@
+// @vitest-environment jsdom
+import { render } from "@testing-library/react"
+import { describe, expect, it, vi } from "vitest"
+
+import { useAddressedFleetSessions } from "../addressedFleetSessions"
+import type {
+  UseWorkspaceAgentSessions,
+  WorkspaceAgentSessionsApi,
+} from "../WorkspaceAgentFront"
+
+const loadArchived = vi.fn()
+const setArchived = vi.fn()
+
+const useSessions: UseWorkspaceAgentSessions = ({ agentTypeId, sourceIdentity }) => ({
+  sourceIdentity,
+  sessions: [{ id: `${agentTypeId}-1`, title: "chat", archived: true }],
+  loading: false,
+  archivedLoaded: false,
+  archivedLoading: false,
+  hasMoreArchived: false,
+  activeSessionId: null,
+  switch: vi.fn(),
+  create: () => ({ id: `${agentTypeId}-2` }),
+  setArchived,
+  loadArchived,
+  delete: vi.fn(),
+})
+
+function renderFleet(agents: Array<{ agentTypeId: string; label: string }>) {
+  let api: WorkspaceAgentSessionsApi | undefined
+  function Harness() {
+    const fleet = useAddressedFleetSessions({
+      agents,
+      selectedAgentTypeId: agents[0]?.agentTypeId,
+      selectAgentTypeId: vi.fn(),
+      discoveryLoading: agents.length === 0,
+      discoveryError: undefined,
+      useSessions,
+      requestHeaders: {},
+      storageKey: "fleet",
+      workspaceId: "ws",
+      enabled: true,
+      fleetSourceIdentity: "fleet",
+      sourceIdentityForAgent: (agentTypeId: string) => `src:${agentTypeId}`,
+    })
+    api = fleet.api
+    return <>{fleet.sources}</>
+  }
+  render(<Harness />)
+  return () => api as WorkspaceAgentSessionsApi
+}
+
+describe("addressed fleet sessions archive capability", () => {
+  // #1453: `every` over an empty list is true, so while Agent discovery was
+  // still running (no controller published yet) the fleet claimed full archive
+  // support and handed the left pane a `loadArchived` that loaded nothing. The
+  // pane probes for archived chats exactly once, latched on that no-op, and
+  // never asked again — leaving the Archived section, the only way back from
+  // Archive, permanently hidden.
+  it("does not claim archive support before any Agent controller exists", () => {
+    const api = renderFleet([])()
+    expect(api.loadArchived).toBeUndefined()
+    expect(api.setArchived).toBeUndefined()
+    expect(api.archivedLoaded).toBeUndefined()
+  })
+
+  it("exposes archive and restore once every addressed Agent reports the capability", async () => {
+    const api = renderFleet([{ agentTypeId: "alpha", label: "Alpha" }])()
+    expect(api.loadArchived).toBeDefined()
+    expect(api.archivedLoaded).toBe(false)
+    await api.setArchived?.("alpha-1", false)
+    expect(setArchived).toHaveBeenCalledWith("alpha-1", false, "alpha")
+  })
+})
--- a/packages/workspace/src/app/front/__tests__/addressedFleetSessions.test.tsx
+++ b/packages/workspace/src/app/front/__tests__/addressedFleetSessions.test.tsx
@@ -1,48 +1,72 @@
 // @vitest-environment jsdom
-import { render } from "@testing-library/react"
-import { describe, expect, it, vi } from "vitest"
+import { StrictMode, useCallback, useMemo, useState } from "react"
+import { render, waitFor } from "@testing-library/react"
+import { beforeEach, describe, expect, it, vi } from "vitest"
 
-import { useAddressedFleetSessions } from "../addressedFleetSessions"
+import { WorkspaceAttentionProvider } from "../../../front/attention/WorkspaceAttentionProvider"
+import { AppLeftPane, createAppLeftNavigationEntries } from "../../../front/layout/plugin-tabs/AppLeftPane"
+import { useAddressedFleetSessions, type WorkspaceAddressedAgentOption } from "../addressedFleetSessions"
 import type {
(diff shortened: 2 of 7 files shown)


hachej:
REVISE

Blocker: the archive probe latch is component-lifetime state, but archive inventory/capability is fleet-source-scoped. `addressedFleetSessions` now correctly withholds `loadArchived` for the initial empty fleet, and the initial 0→N path should fire once; StrictMode effect replay is deduped because the ref is set synchronously. But after that first probe, any controller/source loss makes `archiveCapable` false and removes `loadArchived`/`setArchived`; when a replacement controller returns with fresh `archivedLoaded=false`, the handlers reappear while `AppLeftPane`’s `archivedProbeAttemptedRef` remains true. The replacement inventory is never probed. An old in-flight request does not save this path because `usePiSessions` discards publication when its source is no longer current. A stale exposed `setArchived` can likewise optional-chain to a missing controller and silently no-op during the flap.

Please scope/re-arm the one-shot to the archive inventory/controller cohort (without reintroducing the rejected-request retry loop), and cover the lifecycle that caused this bug: one mounted integration harness going 0→N→controller/source loss→replacement under StrictMode, asserting only current real controllers receive exactly one probe. The added fleet tests use separate static mounts for `[]` and `[alpha]`; they never exercise 0→N, compose with the pane latch, call `loadArchived`, or cover loss/replacement/remount. Also include archive capability presence in the snapshot equality contract if controllers may change it without another compared field changing.

The requested touched suites are green: 3 files, 58 tests. The restore movement test is behavioral (action, callback, state transition, section disappearance, row retained), and “Restore session” is consistent in menu/accessibility text and tests.

hachej:
## Revision — probe latch scoped to the inventory cohort

**Blocker: FIXED.** You were right that the latch was the wrong scope. `AppLeftPane` now records *which* inventory it probed instead of *that* it probed, via a new `archivedInventoryKey` prop (the source / controller cohort behind `onLoadArchived`), and clears the record the moment the capability is withdrawn — because what comes back after a controller/source loss is a different inventory with a fresh, unloaded pager, and it is owed its own single probe.

Why this does not reopen the #1429 retry storm: a rejected `loadArchived` is not a withdrawal. The handler is still there and the key is unchanged, so the latch holds and the probe stays at exactly one attempt per inventory. Only the disappearance of the capability re-arms it. The ref is still written synchronously before the call, so StrictMode's effect replay is deduped by the same write that closes the latch.

`WorkspaceAgentFront` passes `sessionApi.sourceIdentity` as the key, which also closes the single-Agent path: a source change there swaps the inventory without ever dropping `loadArchived`, and a key-less latch would have stranded that one too.

**Snapshot equality: FIXED.** `controllerStateEqual` now compares archive capability *presence* (`Boolean(setArchived)`, `Boolean(loadArchived)`) alongside its paging state — a controller can gain or lose archive support without any other compared field moving, and the fleet's capability is computed from exactly that.

**Test: the lifecycle, not two static mounts.** One mounted `AppLeftPane` fed by the real `useAddressedFleetSessions`, under `StrictMode`, rerendered through: discovery (`agents: []`) → controllers arrive → controller loss → replacement **under the same fleet source identity** (so only the withdrawal-clear can save it) → source change → steady-state re-render. Assertion is the exact probe sequence tagged by serving source, so a StrictMode double-fire or a storm fails just as loudly as a stranded inventory. The two static capability tests stay as the contract for "no vacuous capability".

Verified red before green: with the latch reverted to lifetime scope the lifecycle test fails `expected [ 'src:alpha:g1' ] to deeply equal [ 'src:alpha:g1', 'src:alpha:g1' ]` — the stranded replacement, exactly as reviewed.

### Proof
- `addressedFleetSessions.test.tsx` + `AppLeftPane.test.tsx` + `AppLeftPaneSessionRow.test.tsx`: 3 files, **59 tests green** (was 58; +1 lifecycle case).
- `packages/workspace` `tsc --noEmit`: clean.
- `WorkspaceAgentFront.test.tsx`: 98/99. The one failure, `keeps classic workspace sources available outside plugin-tabs mode`, is unrelated to this branch — it renders the classic (non-plugin-tabs) shell with explicit `sessions` props, no `sessionApi`, no fleet, so none of the three changed code paths are mounted; it fails identically with this revision stashed, and CI's `Unit Tests Changed` passed on the previous commit. Local environment, not this diff.

## UI Review diagnosis: pre-existing harness flake, reproducible on `main`

Not the #1390 mobile hard-gates / `ensureAgentNavigation` signature, and **no hard gate was evaluated at all** — the job dies before the critic/gate phase, during bombadil's reproduction replay of spec `workspace-command-palette`:

```
no match for original:
Clicking <button> (x: 199.4, y: 23.5)
in the set of available actions:
Waiting
Clicking <button> (x: 28.0..=28.0, y: 30.0..=30.0)
Error: reproduction and original test diverged!
Error: UI_REVIEW_BOMBADIL_FAILED:1
```

The same spec fails the same way on `main` — run [33182547405](https://github.com/hachej/boring-ui/actions/runs/33182547405) (commit `40b9346`, no part of this branch), with the *mirror image* of the same pair:

```
no match for original:
Clicking <button> (x: 28.0, y: 30.0)
in the set of available actions:
Waiting
Clicking <button> (x: 199.4..=199.4, y: 23.5..=23.5)
```

Two buttons, and which one is offered at that step swaps between record and replay. That is replay non-determinism in the harness, not a UI change: on this branch the Archived disclosure cannot even render in that fixture (`GET /api/v1/agents/alpha/sessions?limit=50&archived=archived` returns no archived rows, and the section renders only at count > 0), so nothing new is on screen for a gate or a click to land on. `main`'s very next run ([33190895296](https://github.com/hachej/boring-ui/actions/runs/33190895296)) is green with no harness change in between.

Recommend a re-run of UI Review on this branch; a persistent divergence would be a harness bead of its own (distinct from #1390 / #1431), not a blocker on this diff.

hachej:
SHIP

Re-verified head ecce0c3cad4721b9f62649d5143a18e1bda77fb8 against the prior blocker and #1430.

- The archived probe is now scoped to archivedInventoryKey. Capability withdrawal clears the recorded cohort, so capable → lost → replacement under the same fleet identity receives exactly one fresh probe; a direct source-identity change likewise receives one.
- The cohort ref is written synchronously before onLoadArchived. StrictMode replay, callback identity churn, rejection/loading transitions, and stable-cohort rerenders therefore remain deduped; the #1430 attempt-once protection is preserved.
- Archive capability presence is part of controller snapshot equality, so loss/gain cannot be hidden when other paging fields stay unchanged.
- The StrictMode integration test exercises discovery → controller arrival → loss → same-source replacement → source change → stable rerender and asserts the exact probe sequence [g1, g1, g2]. Independent read-only audit found no material issue.

Local proof:
- Touched workspace suites: 3 files, 59/59 passed. The shared host was heavily saturated, so the unchanged 5-second default timeout produced rotating unrelated timeouts; the same suites passed cleanly with a 20-second test timeout, while the new cohort and #1430 attempt-once cases passed in focused runs.
- Current-head Typecheck is green. Current-head UI Review has passed the Bombadil exploration step that previously failed.

UI Review diagnosis confirmed: the earlier red terminated during Bombadil reproduction replay, before critic/hard-gate evaluation, with the exact two-button divergence seen on main at 40b9346 in mirror-image order. After #1431 merged, the immediate main run was green. This was the harness flake, not a regression from the archived-session UI.

VERDICT: SHIP

hachej:
CI update on `ecce0c3`: **all checks green, UI Review included** (17 pass, 4 skipped). That settles the diagnosis in the comment above — the `workspace-command-palette` divergence was bombadil replay non-determinism, not this diff, and the same spec had already failed the same way on `main` (run 33182547405). No hard gate was ever evaluated, and none failed.

PR #1457, opened 2026-08-28
fix(workspace): Tasks panel gets non-zero size when editor tabs open

### LINKED ISSUE #1451: Tasks panel renders at width:0 when editor tabs are open — mounted but invisible
Deep playground smoke 2026-08-28 (origin/main 465f521c6). Repro: open any file in the workbench (e.g. greeter.ts), then click 'Tasks' in the left rail → the full Kanban mounts in the DOM (all column headings present via querySelectorAll) but its flex wrapper computes to width:0/height:56 — invisible and inaccessible; the pane still shows the editor. Closing all editor tabs first, then clicking Tasks, renders correctly. Looks like a flex-sizing race in the workbench-tabs ↔ plugin-panel composition. Evidence: shots4x/4x-52a-tasks-retry.png (working) vs 4x-41-fullpage.png (broken state).

hachej:
## Summary
- `ChatLayout` sized the workbench (editor tabs) purely off `window.innerWidth`, ignoring how much width the host's own chrome (e.g. the plugin-tabs app-left rail) had already consumed, and never reserved room for a mandatory chat overlay (Tasks/Skills) that is deliberately kept open beside the workbench.
- On a medium desktop viewport with an editor tab open, the persisted workbench width (default 680px) could consume the entire real row, leaving the Tasks overlay's flex wrapper computing to a few px wide — mounted but invisible/unusable.
- Fix: measure `ChatLayout`'s own root width via `ResizeObserver` instead of trusting the full window viewport, and reserve a minimum width (280px) for the chat column whenever a chat overlay is open beside the workbench.

## Root cause
- `packages/workspace/src/front/layout/ChatLayout.tsx`: `surfaceMax`/`effectiveSurfaceWidth` (the workbench's width) were computed from `useViewportWidth()` (full window width), not the actual space available to `ChatLayout`'s own flex row. In the plugin-tabs layout, the app-left rail/pane sits *outside* `ChatLayout` and consumes real width that `ChatLayout` never subtracted.
- Separately, `ChatLayout` deliberately skips its narrow-viewport chat-auto-collapse effect whenever a chat overlay (e.g. Tasks) is active (so the overlay stays visible), but nothing then constrained the workbench's width to leave the chat column any room — so the fixed-width workbench pane could squeeze the chat/overlay column down to near-zero.

## Fix
- Added a `ResizeObserver` on `ChatLayout`'s root (`shellRef`) to measure `rowWidth`, the real available width for the chat+workbench row.
- `surfaceMax` and `effectiveSurfaceWidth` are now derived from `rowWidth`, not `viewport`.
- Added `chatOverlayReserve` (280px) subtracted from the workbench's width budget whenever a chat overlay is open beside an open workbench, so the overlay always gets a legible minimum width instead of computing to ~0.

## Proof
Reproduced against a live playground dev server (`origin/main` @ 465f521c6, viewport 1024×800): opening `README.md`, then clicking Tasks in the left rail.
- **Before fix:** `[data-boring-workspace-part="tasks-overlay"]` bounding box = `{ width: 63, height: 800 }` — mounted, effectively invisible.
- **After fix:** same flow → `{ width: 278, height: 800 }` — visible and usable; workbench width auto-shrinks from 680px to 465px to make room. At a wide viewport (1280px) behavior is unchanged (`width: 319`).

## Test plan
- [x] Added `apps/workspace-playground/e2e/tasks-panel-width.spec.ts` — opens a file, clicks Tasks, asserts the overlay's bounding box has width/height > 200px, at a 1024×800 viewport that reproduces the bug pre-fix.
- [x] `pnpm -C packages/workspace exec vitest run` — 129/129 tests pass in `src/front/layout/**` (one unrelated suite fails to resolve `@hachej/boring-bash/shared`, a pre-existing unbuilt-dependency issue in this fresh checkout, not caused by this change).
- [x] `pnpm -C packages/workspace run typecheck` — only a pre-existing, unrelated error in `WorkspaceAgentFront.tsx` (confirmed present with this diff reverted).
- [ ] CI: full `apps/workspace-playground` e2e suite (needs `build:deps`, not run locally due to time).

Fixes #1451

### DIFF
--- a/apps/workspace-playground/e2e/tasks-panel-width.spec.ts
+++ b/apps/workspace-playground/e2e/tasks-panel-width.spec.ts
@@ -1,0 +1,40 @@
+import { expect, test } from "@playwright/test"
+
+/**
+ * #1451: opening a file (editor tab) put the workbench at its persisted
+ * default width (680px). On a medium desktop viewport that left too little
+ * room for the chat column once the Tasks overlay forced it back open —
+ * ChatLayout deliberately skips its narrow-viewport chat-auto-collapse
+ * whenever a chat overlay is active, so the overlay's flex wrapper was the
+ * thing left computing to ~0 width instead. See ChatLayout.tsx's
+ * `chatOverlayReserve` / measured `rowWidth`.
+ */
+test.describe("Tasks panel width with editor tabs open", () => {
+  test.use({ viewport: { width: 1024, height: 800 } })
+
+  test("Tasks overlay stays visible and non-zero width after opening a file", async ({ page }) => {
+    await page.goto("/?fresh=1")
+
+    const filesButton = page.locator('[aria-label="Files"]').first()
+    await expect(filesButton).toBeVisible({ timeout: 15_000 })
+    await filesButton.click()
+
+    await page.getByText("README.md", { exact: true }).first().click()
+    await expect(page.getByRole("tab", { name: /README\.md/ })).toBeVisible({ timeout: 10_000 })
+
+    await page.getByRole("button", { name: "Tasks" }).click()
+
+    const overlay = page.locator('[data-boring-workspace-part="tasks-overlay"]')
+    await expect(overlay).toBeVisible()
+
+    const box = await overlay.boundingBox()
+    expect(box).not.toBeNull()
+    // Pre-fix this collapsed to ~63px — effectively invisible/unusable. The
+    // fix reserves a real minimum for the chat column whenever it must host
+    // an overlay, so the board stays legible.
+    expect(box!.width).toBeGreaterThan(200)
+    expect(box!.height).toBeGreaterThan(200)
+
+    await expect(page.getByText("Tasks", { exact: true }).first()).toBeVisible()
+  })
+})
--- a/apps/workspace-playground/e2e/tasks-panel-width.spec.ts
+++ b/apps/workspace-playground/e2e/tasks-panel-width.spec.ts
@@ -1,4 +1,4 @@
-import { expect, test } from "@playwright/test"
+import { expect, test, type Locator } from "@playwright/test"
 
 /**
  * #1451: opening a file (editor tab) put the workbench at its persisted
@@ -9,6 +9,39 @@ import { expect, test } from "@playwright/test"
  * thing left computing to ~0 width instead. See ChatLayout.tsx's
  * `chatOverlayReserve` / measured `rowWidth`.
  */
+
+/**
+ * The workbench and the overlay's chat column both animate their width over
+ * 280ms (`transition-[...,width,...] duration-[280ms]` in ChatLayout.tsx).
+ * A single `boundingBox()` sample right after the overlay becomes visible
+ * can land mid-transition and read an intermediate, pre-fix-like width —
+ * visible does not mean geometry has settled (#1457 review finding 3).
+ * Poll until two consecutive samples, spaced further apart than a single
+ * animation frame, report the same box.
+ */
+async function waitForStableBoundingBox(locator: Locator, timeoutMs = 5_000) {
+  let previous: Awaited<ReturnType<Locator["boundingBox"]>> = null
+  await expect
+    .poll(
+      async () => {
+        const current = await locator.boundingBox()
+        const stable = current !== null
+          && previous !== null
+          && current.width === previous.width
+          && current.height === previous.height
+          && current.x === previous.x
+          && current.y === previous.y
+        previous = current
+        return stable
+      },
+      { timeout: timeoutMs, intervals: [50, 50, 100, 100, 200] },
+    )
+    .toBe(true)
+  const box = await locator.boundingBox()
+  if (!box) throw new Error("element has no bounding box after it stabilized")
+  return box
+}
+
 test.describe("Tasks panel width with editor tabs open", () => {
   test.use({ viewport: { width: 1024, height: 800 } })
 
@@ -27,13 +60,13 @@ test.describe("Tasks panel width with editor tabs open", () => {
     const overlay = page.locator('[data-boring-workspace-part="tasks-overlay"]')
     await expect(overlay).toBeVisible()
 
-    const box = await overlay.boundingBox()
-    expect(box).not.toBeNull()
-    // Pre-fix this collapsed to ~63px — effectively invisible/unusable. The
-    // fix reserves a real minimum for the chat column whenever it must host
-    // an overlay, so the board stays legible.
-    expect(box!.width).toBeGreaterThan(200)
-    expect(box!.height).toBeGreaterThan(200)
+    // Pre-fix this settled at ~63px — mounted but effectively
+    // invisible/unusable. The fix reserves a real minimum for the chat
+    // column whenever it must host an overlay, so the board stays legible
+    // once its (and the sibling workbench's) width transition settles.
+    const box = await waitForStableBoundingBox(overlay)
+    expect(box.width).toBeGreaterThan(200)
+    expect(box.height).toBeGreaterThan(200)
 
     await expect(page.getByText("Tasks", { exact: true }).first()).toBeVisible()
   })
--- a/packages/workspace/src/front/layout/ChatLayout.tsx
+++ b/packages/workspace/src/front/layout/ChatLayout.tsx
@@ -255,14 +255,28 @@ export function ChatLayout(props: ChatLayoutProps) {
   // is armed one frame *later*, in this separate effect — never in the same
   // commit as a width correction — so "transition turns on" and "width
   // changes" can never be the same style recalculation.
+  //
+  // The gate's source of truth is `surfaceTransitionEnabled` alone — no
+  // separate "armed" ref. An earlier version set a ref to `true` *before*
+  // scheduling the frame, then never reset it if that frame's schedule was
+  // cancelled (e.g. a second, distinct `measuredRowWidth` arrives before the
+  // first frame runs — the effect's cleanup cancels the pending callback,
+  // but the ref stayed `true`, so the next effect run bailed out at its
+  // first line and never rescheduled — `surfaceTransitionEnabled` then
+  // stayed `false` for the component's entire lifetime, permanently
+  // disabling this transition for every later legitimate change too:
+  // collapse/restore, fullscreen, drag-resize, a genuine host-chrome
+  // resize). Gating on the state value itself instead means every effect
+  // run that hasn't yet enabled it will (re)schedule a frame — including
+  // one whose predecessor's frame was cancelled — and once
+  // `surfaceTransitionEnabled` flips `true` the guard's own dependency stops
+  // any further scheduling. See #1457.
   const [surfaceTransitionEnabled, setSurfaceTransitionEnabled] = useState(false)
-  const surfaceTransitionArmedRef = useRef(false)
   useEffect(() => {
-    if (surfaceTransitionArmedRef.current) return
+    if (surfaceTransitionEnabled) return
     // No real measurement yet, and one is still possible — wait for it
     // rather than arming against the unmeasured estimate.
     if (measuredRowWidth === null && typeof ResizeObserver !== "undefined") return
-    surfaceTransitionArmedRef.current = true
     const raf = typeof requestAnimationFrame === "function"
       ? requestAnimationFrame(() => setSurfaceTransitionEnabled(true))
       : (setTimeout(() => setSurfaceTransitionEnabled(true), 0) as unknown as number)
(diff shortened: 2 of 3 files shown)


hachej:
REVISE

Blocking findings:

1. **The observer measures the wrong box.** `shellRef` is attached to the outer shell (`ChatLayout.tsx:488-496`), which also contains the session drawer and workbench-left drawer (`:513-622`), while the chat+workbench row whose width is being budgeted starts at `:624`. Therefore `rowWidth` at `:163-174` still overstates the usable row whenever either internal drawer is open. Example: a 745px shell with a 280px workbench-left drawer leaves a 465px inner row, but the current math can allocate 465px to the workbench and leave the chat overlay near zero. Observe the inner row (or account for the internal panes) and cover overlay + internal drawer combinations.

2. **Persisted initial-open state can flash the broken layout.** `measuredRowWidth` initially falls back to the full viewport (`:163-164`) and the real width is only read in a passive `useEffect` (`:200-211`). If Tasks/Skills and the workbench are already open on mount, first paint can use the oversized persisted workbench width, then animate down through the existing 280ms width transition (`:717-735`). Use a before-paint/CSS-safe sizing strategy and add an initial-open regression test.

3. **The e2e assertion is timing-dependent.** The spec waits only for visibility and then takes one immediate `boundingBox()` sample (`tasks-panel-width.spec.ts:27-36`) while the sibling workbench width is transitioning for 280ms. Visibility does not mean geometry has settled, so it can sample an intermediate pre-fix-like width. Replace this with a web-first/polling width assertion.

The observer cleanup is correct (`disconnect()`), and I found no direct self-sustaining feedback loop because the observed outer width is parent-controlled. The derived reserve also does not overwrite the persisted workbench width, and the overlay auto-expand/auto-collapse branches converge without an oscillation. Continuous resize callback/dedupe/cleanup behavior remains untested because the global `ResizeObserver` mock is a no-op.

Validation: `pnpm -C packages/workspace exec vitest run src/front/layout` — **190/190 passed** after building the fresh-worktree package prerequisites.

hachej:
Addressed all three blocking findings in d1b57c8bd (pushed).

**Finding 1 — wrong measured box: FIXED.**
`shellRef` was the outer shell (includes the session/workbench-left drawers). Added a dedicated `rowRef` on the actual chat+workbench flex row (`data-boring-workspace-part="chat-workbench-row"`, `ChatLayout.tsx:730`) and moved the `ResizeObserver` onto it. The pre-measurement fallback (`rowWidthEstimate`) now also subtracts the workbench-left drawer's width, not just nav — it previously subtracted neither.
Proof: new test `reserves room for a chat overlay on first render, accounting for an open workbench-left drawer (#1451/#1457)` in `presets.test.tsx` — 1024px viewport, workbench-left drawer open (280px), chatOverlay + surface open, persisted `surfaceWidth=680`. Pre-fix math (row = full viewport or full shell, drawer not subtracted) would allocate min(680, 1024-280)=680 or similar to the workbench, leaving the overlay ~0-44px. Post-fix the workbench is capped at 464px (`1024 - 280 drawer - 280 overlay reserve`), leaving the full 280px reserve for the overlay.

**Finding 2 — first-paint flash: FIXED.**
Switched the measurement effect from `useEffect` to `useLayoutEffect` (runs before the browser paints) and made the pre-measurement estimate itself already do the drawer-aware math from finding 1, so there's no frame that renders the old unclamped (raw-viewport) width for a passive effect to later correct.
Proof: same new test above asserts the bounded (464px) width is present immediately after `render()` returns — no `act()`/`waitFor()` beyond what `render` itself flushes. A second new test, `does not widen the workbench to fill the whole row when no chat overlay is open`, is the negative control (no overlay/drawer ⇒ workbench keeps its full persisted 680px), so the reserve logic isn't just always clamping.
Side note: I also guarded the measurement against 0-width reads, since jsdom's `ResizeObserver` polyfill (`vitest.setup.ts`) is a no-op that never calls back and `getBoundingClientRect` always reports zero there — without the guard, `measuredRowWidth` would get set to 0 on mount in every test (jsdom, not a real defect in browsers), floor every width computation at 200, and make it impossible to write a real assertion for finding 2. This also means jsdom now exercises exactly the pre-measurement fallback path end-to-end, which is the right target for both findings 1 and 2.

**Finding 3 — timing-dependent e2e assertion: FIXED.**
Replaced the single `boundingBox()` sample with `waitForStableBoundingBox`, which polls (`expect.poll`, 50/50/100/100/200ms intervals) until two consecutive samples agree on x/y/width/height before asserting — so it can no longer sample mid the 280ms width transition.

**Re-verified the original repro end to end after all three changes** (live playground dev server, not just unit tests): rebuilt `packages/workspace`, ran the same before/after browser check as the original PR.
- 1024×800, editor tab open, click Tasks: overlay bounding box `{ width: 278, height: 800 }` (same as before this revision — confirms the finding-1 fix, which changes *which* element is measured, doesn't regress the original fix).
- 1280×800 (wide, no squeeze needed): `{ width: 319, height: 800 }`, unchanged.

**Test run:** `pnpm -C packages/workspace exec vitest run src/front/layout` — **192/192 passed** (190 existing + 2 new). `pnpm -C packages/workspace run typecheck` — clean, no errors.

I did not find anything to rebut — all three findings were real defects, now fixed with tests reproducing each.

hachej:
REVISE

Findings 1 and 3 are closed:

- `rowRef` is attached to the actual chat+workbench flex row, and both the synchronous read and `ResizeObserver` now measure that row rather than the drawer-containing shell.
- The e2e now polls until two consecutive boxes agree instead of taking the old immediate sample during the 280ms transition.

Finding 2 is not fully closed in the real plugin-tabs mount. `rowWidthEstimate` still begins from `window.innerWidth` and subtracts only ChatLayout-owned drawers. `PluginTabsWorkspaceShell` consumes an external persisted app-left pane (276px by default) before ChatLayout, so at a 1024px viewport the initial overlay+workbench render can still use a 680px workbench against the 1024px estimate, then correct to roughly 464–468px after measuring the real ~744–748px row.

The switch to `useLayoutEffect` does not by itself eliminate that flash: `getBoundingClientRect()` forces layout with the initial 680px style, then the before-paint state update changes a workbench whose width has a 280ms CSS transition. That can establish 680px as the transition start and first-paint/animate the squeezed overlay. The new jsdom regression explicitly has no real layout or transitions and only models a ChatLayout-owned drawer, so it cannot catch this host-level first-paint case. Please make the initial style safe for external host chrome (or suppress the transition until the first real measurement) and cover the plugin-tabs initial-open case in a real browser.

Probe results:

- The zero-width guard can ignore a legitimately fully squeezed or hidden `min-w-0` row, but this is non-blocking: the row clips its children, and a later positive observer sample recovers.
- The layout-effect read and the observer initial delivery may invoke the state updater twice, but the `previous === width` bailout prevents a second geometry update; no harmful mount feedback loop found.

Validation: `pnpm -C packages/workspace exec vitest run src/front/layout` — 12 files, 192/192 tests passed.

hachej:
Addressed the remaining finding and both probes in 3c50428d0 (pushed).

**Finding 2 — plugin-tabs initial-open flash: FIXED (for real this time).**
You were right that `useLayoutEffect` alone wasn't sufficient. Root cause of the gap: `rowWidthEstimate` only subtracts ChatLayout-*owned* drawers (nav/workbench-left); the plugin-tabs app-left pane (~276px) lives entirely outside ChatLayout and is invisible to it, so on first render there the estimate still overstates the row. And even with `useLayoutEffect`, `measure()`'s `getBoundingClientRect()` call forces a layout/style pass against the stale first-render style — if the width-transition is already active on that property at that instant, the browser can treat that forced-layout value as the transition's start point and animate through it, even though it was never actually painted. Exactly the mechanism you described.

Fix: the workbench's width-transition classes are now withheld until a real `rowRef` measurement has landed, and armed one frame *later* in a separate effect — never in the same commit as a width correction — so "transition turns on" and "width changes" can never be part of the same style recalculation. Legitimate later changes (drag-resize, collapse toggle, a real host-chrome resize) still animate normally.

**Real-browser proof of the plugin-tabs initial-open case**, as requested: established persisted state (README open + Tasks overlay open, 1024×800), then reloaded — so both are already active on the very first render, app-left pane at its real default width. Sampled the workbench `<aside>`'s `style.width` / computed width / `transitionDuration` on every `requestAnimationFrame` for the first 60 frames after reload:
```
{"n":0,"width":null,"transitionDuration":null}                  // not yet mounted
{"n":43,"width":"465px","computedWidth":"465px","transitionDuration":"0s"}   // first painted value: already correct, no transition armed
{"n":59,"width":"465px","computedWidth":"465px","transitionDuration":"0.28s"} // transition armed afterward, value unchanged
distinct widths seen across all frames: [0, 465]   // 680 (the stale persisted value) never appears
```
Tasks overlay bounding box after reload: `{ width: 278, height: 800 }` — visible and legible from the first frame, no squeeze.

**Probe 1 (0-width guard masking a legit collapsed state): confirmed non-blocking, now with a test.** New test installs a controllable `ResizeObserver` (the global mock is a no-op, so this was previously untestable), fires an explicit 0-width sample, and confirms it's ignored (width stays at the pre-collapse value) — then fires a positive sample and confirms it recovers correctly (`464px`, matching `744 row − 280 reserve`). No stuck state.

**Probe 2 (useLayoutEffect + ResizeObserver double-set on mount): confirmed non-blocking, now with a test — with one correction to my first attempt.** My first version of this test asserted zero additional React commits on an identical second delivery via a `Profiler`, and it failed: React's own docs note that returning the same value from a state updater can still let the owning component render once more before bailing out of committing (https://react.dev/reference/react/useState#bailing-out-of-a-state-update — "React may still need to render that specific component again before bailing out"). That's expected/documented, not a defect, so I rewrote the assertion to check what actually matters: after an identical second delivery, the *rendered DOM width* is unchanged (no second geometry update reaches the screen) — which passes.

**Test run:** `pnpm -C packages/workspace exec vitest run src/front/layout` — **194/194 passed** (192 + 2 new probe tests). `pnpm -C packages/workspace run typecheck` — clean.

hachej:
REVISE

Blocking finding:

- The transition-arming effect is not cleanup-safe. `surfaceTransitionArmedRef.current` is set before the rAF is scheduled (`ChatLayout.tsx:265-268`), but cleanup cancels that callback without resetting the ref (`:269-271`). If the effect is cleaned up before the frame runs—e.g. a second distinct `measuredRowWidth` update lands before rAF, or Strict Mode replays the no-`ResizeObserver` fallback—the next setup exits at `:261` and `surfaceTransitionEnabled` remains false for the component lifetime.
- This is not confined to plugin-tabs: the gate controls the desktop workbench transition class at `:790`, so affected ordinary ChatLayout open/close, fullscreen/restore, and resize changes snap instead of animating. The existing tests assert geometry but never prove the class arms; the jsdom no-op ResizeObserver also leaves the gate disabled and masks this lifecycle.
- Please make the armed marker reflect a completed enable (or reset it when cleanup cancels), and add a lifecycle test that waits a frame and asserts the desktop workbench transition class is present, including cancellation/rescheduling coverage.

The zero-width recovery and identical-sample geometry probes look correct, and the plugin-tabs first-frame evidence closes the stale-width flash itself.

Validation: `pnpm -C packages/workspace exec vitest run src/front/layout` — 12 files, 194/194 passed.

hachej:
Fixed in 8c76cfd00 (pushed).

**Blocking finding — cleanup-unsafe transition-arming effect: confirmed real defect, fixed.**

You were right, and the impact is broader than the plugin-tabs case: `surfaceTransitionArmedRef.current = true` was set *before* scheduling the arming `requestAnimationFrame`, but the effect's cleanup (which runs whenever `measuredRowWidth` changes again before that frame fires) only cancelled the pending callback — it never reset the ref. The next effect run hit `if (surfaceTransitionArmedRef.current) return` immediately and never rescheduled, so `surfaceTransitionEnabled` stayed `false` for the component's entire lifetime. Since that same flag gates the desktop workbench's width-transition class unconditionally (not just the plugin-tabs first-paint path), this could snap instead of animate on ordinary collapse/restore, fullscreen toggle, and drag-resize too, in any session where a second distinct measurement happened to land before the first arming frame ran.

Fix: dropped the ref entirely. `surfaceTransitionEnabled` state is now the sole source of truth and is included in the effect's own dependency array — every effect run that hasn't yet enabled it (re)schedules a frame, including one whose predecessor's frame was just cancelled, and the guard's own dependency (`surfaceTransitionEnabled`) is what stops further scheduling once it flips `true`. No more "armed but never fired" state is reachable.

**Verified the tests actually catch the bug, not just pass vacuously:** I reverted just the code fix locally (put the old ref-based version back, kept the new tests), reran, confirmed `still arms the workbench width transition when an earlier schedule is cancelled by a superseding measurement` fails against it exactly as you described (`expected false to be true` — transition class never arms), then restored the fix and confirmed it passes again.

Two new tests in `presets.test.tsx`:
- `arms the workbench width transition once the row settles` — the basic case: not armed immediately after a measurement lands (arming is deliberately deferred one frame), armed after waiting.
- `still arms the workbench width transition when an earlier schedule is cancelled by a superseding measurement` — your exact repro: fires two distinct measurements back-to-back before letting either frame run, confirms the transition still arms once things settle.

**Test run:** `pnpm -C packages/workspace exec vitest run src/front/layout` — **196/196 passed** (194 + 2 new). `pnpm -C packages/workspace run typecheck` — clean.

hachej:
SHIP

PR #1458, opened 2026-08-28
Fix showcase route: seed a real backend session before enabling the composer

### LINKED ISSUE #1452: ?showcase=1: composer permanently disabled with 'session was not found' (seedShowcase never resolves)
Deep playground smoke 2026-08-28. Repro: navigate to workspace-playground /?showcase=1 → shell loads, chat panel shows a persistent red 'session was not found' banner, composer never accepts a send (input never clears, nothing renders). The showcase harness pre-seeds SHOWCASE_SESSION_ID client-side but the backend session from seedShowcase()'s async work never materialized this run. Evidence: shots4x/4x-check.png.

hachej:
## Summary
- `?showcase=1` pre-seeded a client-only `SHOWCASE_SESSION_ID` and wrote the fixture transcript into a `localStorage` key that nothing reads anymore.
- The chat pane always requires a real backend session id (`packages/workspace`'s built-in chat pane throws if `params.sessionId` is missing) and hydrates it over the network — that placeholder id 404'd on every load, producing a permanent "session was not found" banner and a composer that never enabled.
- `apps/workspace-playground/src/front/App.tsx` now boots a real backend session (stable-requestId + abort-timed bounded retry, explicit Retry action on terminal failure) and gates rendering the chat panel on that boot completing.
- Decorative padding rows (`?sessions=N`) never hand the chat pane a placeholder id — selecting one materializes a real backend session first.
- Every showcase session is created through a new dev-only wrapper route, `POST /api/v1/playground/showcase-sessions` (`apps/workspace-playground/src/server/dev.ts`), instead of the ordinary create-session endpoint — this is what marks a session as showcase-originated for the boot-time sweep. Provenance is *which route created the session*, never title text (see below).
- Added `apps/workspace-playground/e2e/showcase.spec.ts`: boot/banner/composer/scripted-reply regression, reload-reuse regression, decorative-materialization regression (with a real state-change poll, not just pre-existing-session properties).

**Showcase session retention rule** (`App.tsx`, `dev.ts`, `scriptedPiHarness.ts`): a showcase session lives until one of —
1. it gets a real turn, at which point it's a normal session kept like any other;
2. its tab goes away, and the `pagehide` best-effort cleanup deletes it (skipped only for the reload-reusable boot session, so a same-tab reload's `resumeSessionId` lookup isn't racing its own deletion);
3. failing both — a closed tab, a killed process, an isolated e2e context — the *next* dev-server boot sweeps it: the session store consults a sidecar provenance registry (populated only by the dev-only wrapper route above, never by title content) and deletes any registered, still-empty (`turnCount===0`) session left over from a prior boot.

Nothing showcase-created accumulates past one server boot, and nothing an ordinary session's title says can mark or un-mark it as showcase-originated.

Fixes #1452

## Test plan
- [x] `pnpm --filter workspace-playground run test` (29 passed)
- [x] `pnpm --filter workspace-playground run typecheck`
- [x] `pnpm --filter workspace-playground exec playwright test apps/workspace-playground/e2e/showcase.spec.ts` — multiple consecutive full runs against a reused dev server, all green.
- [x] `scriptedPiHarness.test.ts`: new registry-based sweep coverage — sweeps a marked+empty session from a prior boot, never touches an ordinary session even when its title reproduces the old removed tag text, stops tracking a session once it has a real turn, and doesn't crash on an unresolvable registry entry.
- [x] Live-drove the playground with Playwright against `?showcase=1`: no error banner, composer enabled, message sent, scripted assistant reply rendered.
- [x] Confirmed `apps/workspace-playground/e2e/cmd-effects.spec.ts`'s pre-existing "Focus Chat" flake reproduces identically with this branch's changes stashed out (unrelated to this fix).

### DIFF
--- a/apps/workspace-playground/e2e/showcase.spec.ts
+++ b/apps/workspace-playground/e2e/showcase.spec.ts
@@ -33,10 +33,66 @@ test.describe("workspace-playground showcase route", () => {
     await expect(composer).toBeVisible()
     await expect(composer).toBeEnabled()
 
+    // The scripted harness marks its deterministic final reply with
+    // `PI_NATIVE_ASSISTANT_DONE:<agentTypeId>` (apps/workspace-playground/
+    // src/server/testing/scriptedPiHarness.ts). Waiting on that marker
+    // (rather than "any assistant message container") proves the turn
+    // actually completed streaming, not just that a message_start landed —
+    // the container renders before the final text does.
+    const meta = await (await page.request.get("/api/v1/workspace/meta")).json() as { defaultAgentTypeId: string }
+    const doneMarker = `PI_NATIVE_ASSISTANT_DONE:${meta.defaultAgentTypeId}`
+    const conversation = page.getByLabel("Agent conversation")
+
     const prompt = `showcase smoke ${Date.now()}`
     await composer.fill(prompt)
     await page.locator('[data-boring-agent-part="composer-submit"]').click()
-    await expect(page.getByLabel("Agent conversation").getByText(prompt)).toBeVisible()
-    await expect(page.locator('[data-boring-agent-message-role="assistant"]').last()).toBeVisible({ timeout: 30_000 })
+    await expect(conversation.getByText(prompt)).toBeVisible()
+    await expect(page.getByTestId("chat-working")).toBeVisible({ timeout: 10_000 })
+    await expect(conversation.getByText(doneMarker)).toBeVisible({ timeout: 30_000 })
+    await expect(page.getByTestId("chat-working")).toHaveCount(0, { timeout: 15_000 })
+    await expect(composer).toBeEnabled()
+  })
+
+  test("reuses the same backend session across a reload in the same tab", async ({ page }) => {
+    test.setTimeout(120_000)
+
+    await page.goto("/?showcase=1")
+    const chat = page.locator('[data-boring-agent-part="chat"]')
+    await chat.waitFor({ state: "visible", timeout: 40_000 })
+    await expect(chat).toHaveAttribute("data-pi-chat-connection", "connected", { timeout: 15_000 })
+    const firstSessionId = await chat.getAttribute("data-pi-chat-session-id")
+    expect(firstSessionId).toBeTruthy()
+
+    // Bounded-retention regression: a reload in the same tab must resume the
+    // already-booted (still-empty) session rather than minting a new durable
+    // one every time (gh-1458 review finding #3).
+    await page.reload({ waitUntil: "domcontentloaded" })
+    await chat.waitFor({ state: "visible", timeout: 40_000 })
+    await expect(chat).toHaveAttribute("data-pi-chat-connection", "connected", { timeout: 15_000 })
+    const secondSessionId = await chat.getAttribute("data-pi-chat-session-id")
+    expect(secondSessionId).toBe(firstSessionId)
+  })
+
+  test("switching to a decorative padding session materializes a real backend session", async ({ page }) => {
+    test.setTimeout(120_000)
+
+    // `sessions=3` pads the list with two client-only placeholder rows
+    // (gh-1458 review finding #2) — selecting one must not hand the chat
+    // pane an id that 404s the way the original SHOWCASE_SESSION_ID did.
+    await page.goto("/?showcase=1&sessions=3")
+    const chat = page.locator('[data-boring-agent-part="chat"]')
+    await chat.waitFor({ state: "visible", timeout: 40_000 })
+    await expect(chat).toHaveAttribute("data-pi-chat-connection", "connected", { timeout: 15_000 })
+
+    const decorativeRow = page.locator('[data-boring-workspace-part="app-session-row"][data-boring-session-id="__showcase__-2"]')
+    await decorativeRow.waitFor({ state: "visible", timeout: 10_000 })
+    await decorativeRow.locator("button").first().click()
+
+    await expect(page.getByText("session was not found")).toHaveCount(0)
+    await expect(chat).toHaveAttribute("data-pi-chat-connection", "connected", { timeout: 15_000 })
+    await expect(page.getByRole("textbox", { name: "Agent prompt" })).toBeEnabled()
+
+    const activeSessionId = await chat.getAttribute("data-pi-chat-session-id")
+    expect(activeSessionId).not.toBe("__showcase__-2")
   })
 })
--- a/apps/workspace-playground/e2e/showcase.spec.ts
+++ b/apps/workspace-playground/e2e/showcase.spec.ts
@@ -83,16 +83,32 @@ test.describe("workspace-playground showcase route", () => {
     const chat = page.locator('[data-boring-agent-part="chat"]')
     await chat.waitFor({ state: "visible", timeout: 40_000 })
     await expect(chat).toHaveAttribute("data-pi-chat-connection", "connected", { timeout: 15_000 })
+    // Captured before the click: the pre-existing boot session is already
+    // connected/enabled/error-free, so asserting only those properties after
+    // the click can pass while the row's async materialization POST is still
+    // in flight and the pane hasn't switched onto it at all. Waiting for the
+    // session id to actually change (and the placeholder row to disappear)
+    // proves the switch happened, not just that the old session was fine.
+    const bootSessionId = await chat.getAttribute("data-pi-chat-session-id")
+    expect(bootSessionId).toBeTruthy()
 
-    const decorativeRow = page.locator('[data-boring-workspace-part="app-session-row"][data-boring-session-id="__showcase__-2"]')
+    const placeholderId = "__showcase__-2"
+    const decorativeRow = page.locator(`[data-boring-workspace-part="app-session-row"][data-boring-session-id="${placeholderId}"]`)
     await decorativeRow.waitFor({ state: "visible", timeout: 10_000 })
     await decorativeRow.locator("button").first().click()
+
+    await expect.poll(
+      () => chat.getAttribute("data-pi-chat-session-id"),
+      { timeout: 15_000 },
+    ).not.toBe(bootSessionId)
+    await expect(page.locator(`[data-boring-workspace-part="app-session-row"][data-boring-session-id="${placeholderId}"]`)).toHaveCount(0)
 
     await expect(page.getByText("session was not found")).toHaveCount(0)
     await expect(chat).toHaveAttribute("data-pi-chat-connection", "connected", { timeout: 15_000 })
     await expect(page.getByRole("textbox", { name: "Agent prompt" })).toBeEnabled()
 
     const activeSessionId = await chat.getAttribute("data-pi-chat-session-id")
-    expect(activeSessionId).not.toBe("__showcase__-2")
+    expect(activeSessionId).not.toBe(placeholderId)
+    expect(activeSessionId).not.toBe(bootSessionId)
   })
 })
--- a/apps/workspace-playground/e2e/showcase.spec.ts
+++ b/apps/workspace-playground/e2e/showcase.spec.ts
@@ -1,0 +1,42 @@
+import { expect, test } from "@playwright/test"
+
+/**
+ * Regression for gh-1452: `?showcase=1` pre-seeded a client-side session id
+ * (SHOWCASE_SESSION_ID) that never had a matching backend session, so the
+ * chat pane's network hydrate always 404'd — a permanent "session was not
+ * found" banner and a composer that never enabled. The fix boots a real
+ * backend session before the chat panel renders, so the showcase route must
+ * come up connected with a working composer.
+ */
+
+test.describe("workspace-playground showcase route", () => {
+  test("boots a working session with no error banner and a live composer", async ({ page }) => {
+    test.setTimeout(120_000)
+
+    const pageErrors: string[] = []
+    page.on("pageerror", (error) => pageErrors.push(error.message))
+
+    await page.goto("/?showcase=1")
+
+    // Cold dev-server boot (first request ever, unbundled deps) can take a
+    // while beyond the usual per-assertion timeout — give it real headroom
+    // rather than flaking on a slow first compile.
+    const chat = page.locator('[data-boring-agent-part="chat"]')
+    await chat.waitFor({ state: "visible", timeout: 40_000 }).catch(() => {
+      throw new Error(`showcase chat panel did not render; page errors: ${pageErrors.join(" | ") || "none"}`)
+    })
+
+    await expect(page.getByText("session was not found")).toHaveCount(0)
+    await expect(chat).toHaveAttribute("data-pi-chat-connection", "connected", { timeout: 15_000 })
+
+    const composer = page.getByRole("textbox", { name: "Agent prompt" })
+    await expect(composer).toBeVisible()
+    await expect(composer).toBeEnabled()
+
+    const prompt = `showcase smoke ${Date.now()}`
+    await composer.fill(prompt)
+    await page.locator('[data-boring-agent-part="composer-submit"]').click()
+    await expect(page.getByLabel("Agent conversation").getByText(prompt)).toBeVisible()
+    await expect(page.locator('[data-boring-agent-message-role="assistant"]').last()).toBeVisible({ timeout: 30_000 })
+  })
+})
--- a/apps/workspace-playground/e2e/showcase.spec.ts
+++ b/apps/workspace-playground/e2e/showcase.spec.ts
@@ -1,4 +1,22 @@
+import { readFile } from "node:fs/promises"
+import { dirname, join } from "node:path"
+import { fileURLToPath } from "node:url"
 import { expect, test } from "@playwright/test"
+
+// Matches playwright.config.ts's E2E_SESSION_ROOT default (this file's own
+// env has no BORING_AGENT_SESSION_ROOT — that var is only set on the
+// webServer child process's env via `env -i`).
+const E2E_SESSION_ROOT = join(dirname(fileURLToPath(import.meta.url)), "fixtures", "sessions")
+const SHOWCASE_REGISTRY_PATH = join(E2E_SESSION_ROOT, ".playground-showcase-session-ids.json")
+
+function decodeRegistryEntry(entry: string): { agentKey: string; workspaceId: string; sessionId: string } | undefined {
+  const parts = entry.split(REGISTRY_KEY_SEPARATOR)
+  if (parts.length !== 3) return undefined
+  const [agentKey, workspaceId, sessionId] = parts
+  return { agentKey, workspaceId, sessionId }
+}
+// Must match SHOWCASE_REGISTRY_KEY_SEPARATOR in scriptedPiHarness.ts (U+0001).
+const REGISTRY_KEY_SEPARATOR = ""
 
 /**
  * Regression for gh-1452: `?showcase=1` pre-seeded a client-side session id
@@ -110,5 +128,55 @@ test.describe("workspace-playground showcase route", () => {
     const activeSessionId = await chat.getAttribute("data-pi-chat-session-id")
     expect(activeSessionId).not.toBe(placeholderId)
     expect(activeSessionId).not.toBe(bootSessionId)
+  })
+
+  // gh-1458 review round 6: createWorkspaceAgentServer accepts TWO different
+  // values for `x-boring-workspace-id` — the canonical workspace scope id
+  // and `basename(workspaceRoot)` — and always resolves either one to the
+  // SAME canonical scope before a session record is ever created
+  // (trustedWorkspaceScopeId in createWorkspaceAgentServer.ts). The wrapper
+  // route used to key the showcase provenance registry by whichever raw
+  // header the client presented, so an allowed basename-selector request
+  // would create a session scoped to canonical while marking the registry
+  // under the basename — a permanent mismatch the sweep's full-match rule
+  // would never resolve (it deliberately never prunes an entry it can't
+  // positively verify). This drives the wrapper route directly with the
+  // basename selector and reads the on-disk registry to prove the entry is
+  // keyed by the canonical id regardless.
+  test("provenance registry keys by the canonical workspace scope, not whichever selector the client presented", async ({ page }) => {
+    test.setTimeout(60_000)
+
+    const meta = await (await page.request.get("/api/v1/workspace/meta")).json() as { workspaceId: string; defaultAgentTypeId: string }
+    // In local (non-remote-worker) mode, meta.workspaceId reflects
+    // basename(workspaceRoot) — the OTHER allowed selector, distinct from
+    // the canonical "default" scope this server actually resolves to.
+    const basenameSelector = meta.workspaceId
+    expect(basenameSelector).toBeTruthy()
+    expect(basenameSelector).not.toBe("default")
+
+    const response = await page.request.post("/api/v1/playground/showcase-sessions", {
+      headers: {
+        "content-type": "application/json",
+        "x-boring-workspace-id": basenameSelector,
+      },
+      data: {
+        agentTypeId: meta.defaultAgentTypeId,
+        title: "Basename selector parity check",
+      },
+    })
+    expect(response.status(), await response.text()).toBe(201)
+    const payload = await response.json() as { sessionId?: string }
+    expect(payload.sessionId).toBeTruthy()
+
+    const registryText = await readFile(SHOWCASE_REGISTRY_PATH, "utf8")
+    const registry = JSON.parse(registryText) as string[]
+    const decoded = registry.map(decodeRegistryEntry).filter((entry): entry is NonNullable<typeof entry> => Boolean(entry))
+    const matching = decoded.find((entry) => entry.sessionId === payload.sessionId)
+    expect(matching, `no registry entry found for session ${payload.sessionId}; decoded entries: ${JSON.stringify(decoded)}`).toBeTruthy()
+    // The registry key must be the canonical scope ("default"), never the
+    // basename selector the request presented — this is the belongsTo
+    // parity the review required.
+    expect(matching?.workspaceId).toBe("default")
+    expect(matching?.workspaceId).not.toBe(basenameSelector)
   })
 })
--- a/apps/workspace-playground/src/front/App.tsx
+++ b/apps/workspace-playground/src/front/App.tsx
@@ -8,6 +8,7 @@ import { diagramPlugin } from "@hachej/boring-diagram/front"
 import { createTasksPlugin } from "@hachej/boring-tasks/front"
 import { SHOWCASE_SESSION_ID } from "./showcaseMessages"
 import { LoadingStatesShowcase, type LoadingStateMode } from "./LoadingStatesShowcase"
+import { SHOWCASE_SESSION_TITLE_TAG } from "../shared/showcaseSession"
 
 function isShowcaseRoute(): boolean {
   if (typeof window === "undefined") return false
@@ -196,11 +197,15 @@ export function WorkspaceShell() {
   const liveShowcaseSessionIds = useRef(new Set<string>())
   // Tab-scoped reuse key: bounds session accumulation from repeated boots in
   // the same browser tab (reload, HMR) to one durable session instead of one
-  // per mount. sessionStorage (not localStorage) so it never outlives the
-  // tab. Cleared automatically once the reused session picks up its first
-  // turn — createSession only resumes empty (turnCount === 0) sessions
-  // server-side (embeddedGateway.ts), so a used session naturally falls
-  // through to creating a fresh one on the next reload.
+  // per mount. sessionStorage (not localStorage) so the *pointer* never
+  // outlives the tab. This alone does not bound the durable backend session
+  // itself — closing the tab discards only this reuse pointer, and a new
+  // tab/e2e context still creates its own session. Actual retention is
+  // bounded by two other layers: the pagehide cleanup below (best-effort,
+  // same boot) and the scripted dev server's boot-time sweep in
+  // scriptedPiHarness.ts (authoritative backstop — deletes every
+  // showcase-tagged session left empty by a *previous* boot, regardless of
+  // which tab/context created it). See SHOWCASE_SESSION_TITLE_TAG.
   const showcaseBootStorageKey = "boring-ui-v2:showcase:boot-session-id"
   const requestNewShowcaseSession = useCallback(async (
     title: string,
@@ -218,7 +223,13 @@ export function WorkspaceShell() {
           "x-boring-workspace-id": "default",
         },
         body: JSON.stringify({
-          title,
+          // The backend-stored title is tagged and never surfaced — this UI
+          // only ever renders the `title` this function returns below, taken
+          // from the controlled `sessions` prop it builds client-side (see
+          // ../shared/showcaseSession.ts). The tag lets the scripted dev
+          // server's boot-time sweep find and delete stale, still-empty
+          // showcase sessions without touching unrelated ones.
+          title: `${SHOWCASE_SESSION_TITLE_TAG}${title}`,
           ...(options.requestId ? { requestId: options.requestId } : {}),
           ...(options.resumeSessionId ? { resumeSessionId: options.resumeSessionId } : {}),
         }),
@@ -309,8 +320,20 @@ export function WorkspaceShell() {
   }, [])
   // Decorative rows (padding from `?sessions=N`) start with no backend
   // session behind them. Selecting one materializes a real session first —
-  // keyed by a stable per-row request id so a double-click can't create two
-  // — instead of ever handing the chat pane an id that will 404 (gh-1452).
+  // guarded against a double-click by the in-memory
+  // `materializingShowcaseSessionIds` set below — instead of ever handing
+  // the chat pane an id that will 404 (gh-1452).
+  //
+  // The request id is freshly random per attempt, NOT derived from the
+  // (fixed, deterministic) placeholder row id. A stable "showcase-row-<id>"
+  // request id looked idempotent-safe, but the placeholder id is the exact
+  // same string on every future page load/e2e run against this server —
+  // so a *second* click across a completely unrelated later visit would
+  // replay the ledger's very first receipt for it, handing back a stale
+  // session that a previous visit's pagehide cleanup may have already
+  // deleted. The `materializingShowcaseSessionIds` guard already prevents
+  // a genuine double-click within one page's lifetime, so no cross-visit
+  // request id is needed for that.
   const materializingShowcaseSessionIds = useRef(new Set<string>())
   const handleActiveSessionIdChange = useCallback(
     (sessionId: string | null) => {
@@ -322,7 +345,7 @@ export function WorkspaceShell() {
       if (materializingShowcaseSessionIds.current.has(sessionId)) return
       materializingShowcaseSessionIds.current.add(sessionId)
       const placeholderTitle = showcaseSessions.find((session) => session.id === sessionId)?.title ?? "New chat"
-      void requestNewShowcaseSession(placeholderTitle, { requestId: `showcase-row-${sessionId}`, timeoutMs: 8_000 })
+      void requestNewShowcaseSession(placeholderTitle, { requestId: randomId(), timeoutMs: 8_000 })
         .then((session) => {
           liveShowcaseSessionIds.current.add(session.id)
           setShowcaseSessions((current) => current.map((existing) => (existing.id === sessionId ? session : existing)))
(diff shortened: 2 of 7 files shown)


hachej:
REVISE

Blocking findings:

1. **The cold-boot retry is attempt-bounded, but not time-bounded or idempotent.** `App.tsx:227-246` caps the loop at five POSTs with 3s total backoff, but `requestNewShowcaseSession` gives `fetch` no abort timeout. One hung request leaves the route on the blank pending screen forever. The API already accepts `requestId` (`httpProjection.ts:102-106`), but this caller sends only `title`; because the server generates a new UUID when it is absent (`:334-347`), a committed request whose response is lost can create another durable session on retry. Use one stable request ID for the logical boot and reuse it across attempts, and bound each request with an abort timeout. The terminal failure text does offer reload as an escape, but an explicit retry action would avoid making recovery depend on a full reload (and a hung request never reaches that state).

2. **The dead localStorage seeding path is half-left, and one path recreates #1452.** `App.tsx:9,234,260` still imports/calls `seedShowcase`; `showcaseMessages.ts:253-255` still writes `boring-agent:messages:<sessionId>`, for which a repo-wide search finds no reader. More seriously, `?showcase=1&sessions=N` still renders decorative fake IDs (`App.tsx:56-67`). Selecting one calls the dead seed function and makes that nonexistent backend ID active, returning the user to the same 404/disabled-composer failure. Remove the obsolete storage contract and ensure every selectable session is real (or make decorative rows non-selectable).

3. **Session-per-visit accumulation needs an explicit lifecycle decision.** Every showcase mount creates a durable backend session, controlled-session mode supplies no delete affordance, and reloads/e2e runs accumulate session files indefinitely; non-idempotent retries can multiply them. Even for a dev surface, silent unbounded accumulation is not a good default. Reuse a known showcase session where safe, clean it up, or define bounded retention.

4. **The e2e does not deterministically prove the claimed scripted reply.** `showcase.spec.ts:40` only waits for any last assistant container. The scripted harness emits that container at `message_start` before tool execution and final response text, so the test can pass early. Assert the deterministic final marker (for this agent namespace) and that the composer returns to enabled; keep the assertion scoped to the active conversation.

Proof run: `pnpm --filter @hachej/boring-workspace build && pnpm --filter workspace-playground run test` — **8 files / 26 tests passed**.

hachej:
Addressed all four findings on `fix/1452-showcase-session` (commit bc8c96f). Per-finding disposition:

**1. Attempt-bounded but not time-bounded/idempotent retry — fixed.**
`requestNewShowcaseSession` now takes an `AbortController` with an 8s timeout per attempt, so a hung request can no longer wedge the route on the pending screen forever. The boot flow generates one `requestId` (`showcaseBootRequestIdRef`) reused across every retry — verified this makes retries idempotent: `embeddedGateway.ts`'s `effect()` keys its ledger on `requestId` and replays the original receipt (`record.state === 'completed'` → `replayReceipt`) instead of re-executing, so a committed-but-lost response on attempt N can't mint a second durable session on attempt N+1. Also added an explicit **Retry** button on the terminal error state (`retryShowcaseBoot`) so recovery no longer depends on a full page reload.

**2. Dead localStorage seed path + selectable decorative 404 trap — fixed.**
Removed `seedShowcase()` and its `localStorage.setItem('boring-agent:messages:...')` write entirely from `showcaseMessages.ts`, and dropped the import/calls from `App.tsx`. Decorative padding rows from `?sessions=N` (`App.tsx:56-67`) now materialize a real backend session on selection — `handleActiveSessionIdChange` checks `liveShowcaseSessionIds`; for a not-yet-live id it POSTs a real session (keyed by a stable per-row `requestId` so a double-click can't double-create) and swaps the row's placeholder id for the real one before setting it active. The chat pane is never handed a placeholder id anymore. New e2e: `switching to a decorative padding session materializes a real backend session`.

**3. Session-per-visit accumulation — bounded via reuse, not deleted.**
Rather than adding deletion/cleanup machinery to a dev-only surface, the boot session id is now remembered in `sessionStorage` (tab-scoped, not `localStorage`) and passed back as `resumeSessionId` on the next boot. The server only resumes a session with `turnCount === 0` (`embeddedGateway.ts` `createSession`); once the showcase session has messages, the next boot naturally creates a fresh one. Net effect: reloads/HMR in the same tab reuse one session instead of minting a new one every time; a tab close bounds retention to that tab's lifetime. New e2e: `reuses the same backend session across a reload in the same tab` (asserts `data-pi-chat-session-id` is identical before/after reload).

**4. Non-deterministic e2e assertion — fixed.**
The spec now fetches `/api/v1/workspace/meta` for `defaultAgentTypeId` and waits for the scripted harness's actual terminal marker, `PI_NATIVE_ASSISTANT_DONE:<agentTypeId>` (`scriptedPiHarness.ts`), scoped inside `page.getByLabel("Agent conversation")`, plus the `chat-working` testid going from visible → count 0, plus the composer returning to enabled — matching the pattern already used by `agent-host-golden-route.spec.ts`. This can no longer pass on just the `message_start` container.

**Proof:**
- `pnpm --filter workspace-playground run test` — 8 files / 26 tests passed
- `pnpm --filter workspace-playground run typecheck` — clean
- `pnpm --filter workspace-playground exec playwright test apps/workspace-playground/e2e/showcase.spec.ts` — 3/3 passed, run twice in a row (1.7m, then 52s) for stability

No findings disputed — all four were real defects, all fixed with new/updated regression coverage in `apps/workspace-playground/e2e/showcase.spec.ts`.

hachej:
REVISE

Two blocking gaps remain after re-verifying `bc8c96f06`:

1. **The durable session lifecycle is still unbounded.** `App.tsx:197-203,263-280` stores the id in `sessionStorage`, which reduces reload/HMR churn only while the same tab remains open and the session is empty. Closing the tab discards the only reuse pointer but does not delete the durable backend session; a new tab and each isolated e2e context still create another durable session. Once a session has a turn, the next reload intentionally creates a fresh session and leaves the old one behind. Thus the prior session-per-visit/e2e accumulation finding is not closed, and the claim that tab close bounds retention to the tab lifetime is incorrect. Add actual cleanup, safe reuse across visits, or bounded server-side retention for this dev surface.

2. **The decorative-row regression test can pass without materialization completing.** In `showcase.spec.ts:89-96`, the pre-existing boot chat is already connected, error-free, enabled, and has an id other than `__showcase__-2`. All assertions can therefore succeed immediately after the click while the async POST is still pending. Capture the initial real id and wait for `data-pi-chat-session-id` to change (and/or for the placeholder row to be replaced) before asserting the new chat is connected.

Verified closures/non-findings:
- The 8s per-attempt abort and stable logical boot `requestId` are real. `embeddedGateway.ts:819-839` keys the ledger by workspace scope, auth subject, operation, target, and request id, includes the payload digest, and replays a completed receipt; concurrent repeats are rejected as in-progress rather than re-executed.
- The dead seed writer is removed, and the selection implementation itself does not activate a placeholder id.
- The terminal-marker e2e is scoped to the conversation and waits for working state to clear plus composer re-enablement.
- `resumeSessionId` resolves only an exact empty session under the current workspace/runtime. The generic store is workspace-scoped rather than per-auth-subject, but this standalone playground deliberately has no user/auth host and all requests use the single `local` subject, so cross-user resume is not a blocker for this dev-only surface.

Proof: `pnpm --filter workspace-playground run test` — 8 files / 26 tests passed.

hachej:
Addressed both blockers on `fix/1452-showcase-session` (commit c650ff0). Disposition:

**1. Durable session accumulation — actually closed now, not just reduced.**
Agreed with the finding as stated; my prior sessionStorage-reuse claim was wrong for the tab-close/new-context case. Fixed with a two-layer real bound (retention rule now also stated in the PR description):
- **Client (best-effort, same-boot):** a `pagehide` listener deletes any live showcase session that never got a turn, via `fetch(..., { method: "DELETE", keepalive: true })`. Correction on mechanism: `navigator.sendBeacon` cannot express this — it only ever issues a POST and cannot set the `x-boring-workspace-id` header the delete route requires (MDN, `sendBeacon` spec) — so `fetch` with `keepalive: true` is used instead; it's the documented replacement for exactly this case (non-POST/authenticated pagehide requests) and is honored the same way during unload. The reusable boot session (tracked via `sessionStorage`) is deliberately excluded from this sweep — `pagehide` fires on an in-tab reload too, and deleting it there would race the next load's `resumeSessionId` lookup and break the reload-reuse behavior itself (caught this empirically: it silently downgraded reload-reuse to always-fresh once I added the naive version of this cleanup).
- **Server (authoritative backstop):** `scriptedPiHarness.ts`'s session store now tags every showcase-created session's backend title with a stable, never-displayed marker (`SHOWCASE_SESSION_TITLE_TAG`, new `apps/workspace-playground/src/shared/showcaseSession.ts` — safe to reuse as a request payload because the showcase route's displayed titles come entirely from its own controlled `sessions` prop, never from the backend-stored title) and sweeps every tagged, still-empty (`turnCount===0`) session left over from a *previous* boot as part of `hydrate()` — which only ever runs once per store, memoized, so this only ever touches what predates the current process. This is the backstop for a closed tab, a killed process, or an isolated e2e context, none of which the client-side layer can reach.

I verified the sweep directly (not just via the e2e suite): seeded stale tagged+empty session files from a prior run, confirmed a fresh boot deletes them (checked by file mtime — the sweep freed their numeric id slots, which then got legitimately reused by the same run's new sessions, which is why a plain "does the path still exist" check is misleading; had to check mtime/content to confirm it was actually a *different*, freshly-created session at that recycled id).

**2. Decorative-row test could pass without materialization completing — fixed, and it caught a real bug.**
Rewrote the test to capture the boot session id before the click and `expect.poll` for `data-pi-chat-session-id` to actually change (plus the placeholder row disappearing) before asserting the new session. This is not just a stricter assertion — it caught a genuine defect: the materialize POST's `requestId` was `showcase-row-<placeholder-id>`, and the placeholder id (`__showcase__-2`) is the exact same string on every future page load against the same server. A second click against that placeholder id on a *later*, unrelated visit replayed the ledger's very first receipt for that request id, handing back a stale (sometimes already pagehide-deleted) session instead of creating a new one. Reproduced reliably by running the full spec twice in a row against a reused dev server. Fixed by making the materialize request id freshly random per attempt — the existing in-memory `materializingShowcaseSessionIds` guard already prevents a same-page double-click, which is all a stable id was ever protecting against.

**Proof:**
- `pnpm --filter workspace-playground run test` — 8 files / 26 tests passed
- `pnpm --filter workspace-playground run typecheck` — clean
- `showcase.spec.ts` — 4 consecutive full runs (3 tests each) against a reused dev server, all green. This specific repeated-invocation pattern is what reproduced the request-id replay bug above, so it's the regression's actual proof, not just a flake check.
- Manual sweep verification (mtime-based, described above) confirming stale tagged-empty sessions from a prior boot are deleted while the current boot's legitimately-empty sessions are not touched.

No findings disputed this round either — both were real, the second one worse than described (an active bug, not just a test-quality gap).

hachej:
REVISE

One blocking finding remains: the sweep tag can collide with an ordinary user-created session title and delete that empty session on the next boot.

Evidence:
- `SHOWCASE_SESSION_TITLE_TAG` is the fixed public prefix `[playground-showcase] ` (`apps/workspace-playground/src/shared/showcaseSession.ts:18`).
- The create/rename HTTP schemas accept arbitrary nonempty titles up to 200 characters and forward them unchanged (`packages/agent/src/server/agent-host/httpProjection.ts:102-110,334-345`).
- The scripted store preserves that title, while hydration deletes every persisted `turnCount === 0` record whose title merely starts with the prefix (`apps/workspace-playground/src/server/testing/scriptedPiHarness.ts:168-174,198-205,312-320`). An executable harness repro created an ordinary empty session titled `[playground-showcase] my ordinary draft`; it existed before process-like recreation and was absent afterward.

Impact: the marker is user-controlled data, not durable provenance, so the stated guarantee that unrelated empty sessions are untouched is false and the sweep can permanently remove an ordinary draft. Store explicit showcase provenance outside the title, or otherwise make collision impossible, and cover both stale-showcase deletion and colliding non-showcase retention.

Verified clean at `c650ff0ad7e8feacf2b3ee99c76b189a077b7724`:
- The memoized hydrate/sweep completes before create and never reruns in that store, so it cannot sweep a current-boot session mid-use.
- Each pagehide keepalive DELETE has an attached rejection handler, so cleanup is genuinely best-effort without an unhandled rejection.
- The in-memory placeholder guard is set synchronously before the fresh materialize request and retained through settlement, so same-page double-click cannot double-create.
- The decorative-row e2e now waits for the active backend session id to change and for the placeholder row to disappear.
- `pnpm --filter workspace-playground run test`: 8 files / 26 tests passed.

hachej:
Addressed the blocking finding on `fix/1452-showcase-session` (commit 5b94f89). Disposition on all four points raised:

**Tag collision with user-created sessions — real, fixed.**
Agreed with the finding and the repro. Title text was never durable provenance — the create/rename HTTP schemas accept any nonempty title up to 200 chars and forward it unchanged, so an ordinary session titled starting with the fixed tag would be swept too.

Replaced title-based tagging with route-based provenance:
- New dev-only wrapper route, `POST /api/v1/playground/showcase-sessions` (`apps/workspace-playground/src/server/dev.ts`). Every showcase session — boot, "New chat", decorative-row materialization — now goes through this route instead of the ordinary create-session endpoint. It forwards to that endpoint unchanged via `app.inject` (no extra network hop, no duplicated auth/validation) and, on success, appends the returned session id to a sidecar JSON registry file (`apps/workspace-playground/src/shared/showcaseSession.ts` now documents the mechanism; the old `SHOWCASE_SESSION_TITLE_TAG` constant is gone).
- `scriptedPiHarness.ts`'s boot-time sweep now reads that registry instead of matching title text: a session is swept only if its id is registered *and* still empty. Nothing a user types into a title can mark or un-mark a session, because the ordinary session-creation UI never calls this route at all.
- An id in the registry that this store can't resolve to one of its own records (e.g. it belongs to a different agent namespace sharing the same registry file) is left untouched rather than guessed at — never prunes an entry it can't positively verify it owns.
- `App.tsx` now sends the session's real, undisguised title — the tag is gone, so the backend-stored title matches what's actually displayed (a minor correctness bonus).

New regression coverage in `scriptedPiHarness.test.ts` (4 new tests, unit-level — faster and more direct than round-tripping through e2e for this): a marked-but-empty session from a simulated prior boot is swept; **an ordinary session whose title literally reproduces the old removed tag text (`"[playground-showcase] my ordinary draft"`) survives untouched** (this is the exact repro from the finding, now asserted as a permanent regression test); a marked session that got a real turn is never swept; an unresolvable registry entry doesn't crash hydration or block normal session creation.

**The other three probes — confirmed already closed, no change needed, evidence reiterated:**
- *Boot-time sweep deleting a current-boot session mid-use:* Still correct. `hydrate()` is memoized via `this.hydration ??= this.hydrate()` and only ever runs once per store instance; the sweep runs as the tail of that single call, so everything it inspects (both the loaded records and the registry read) reflects only what existed *before* this process's first hydration — nothing created later in the same boot is ever visible to it.
- *Unhandled rejection on the keepalive DELETE:* Still correct. The pagehide handler's `fetch(...)` call ends in `.catch(() => { /* best-effort */ })` — verified this is present in the current file, unchanged by this round's edits.
- *Fresh-per-attempt requestIds reopening double-create on double-click:* Still correct, and this round didn't touch that code path. The `materializingShowcaseSessionIds` guard is set synchronously (`.add(sessionId)`) before the async materialize request is even issued, and only cleared in `.finally()` after the request settles — a second click on the same row inside that window returns immediately at the guard check, never issuing a second request. The per-attempt random requestId only ever protected against the *cross-visit* replay bug found last round; it was never what prevented a same-page double-click.

**Proof:**
- `pnpm --filter workspace-playground run test` — 8 files / 29 tests passed (26 + 4 new registry tests, net +3 since one subsumes prior coverage)
- `pnpm --filter workspace-playground run typecheck` — clean
- `showcase.spec.ts` — 3 consecutive full runs against a reused dev server, all green
- Verified with a fresh manual browser session (via a standalone dev server instance) that the wrapper route round-trips correctly end-to-end: POST → 201 → registry file written → chat pane connects on the returned real session id

One process-hygiene note for the record: two of my own earlier debug reruns this round were briefly confused by a stale leftover `vite` dev-server process from *my own* prior debugging (35+ minutes old, serving a pre-edit bundle) that Playwright's `reuseExistingServer` picked up instead of starting fresh — not a code defect, just my own leftover process; killed it and re-verified clean.

hachej:
REVISE

Blocking findings:

1. The sidecar registry is not bounded, and an orphaned registry id can delete an unrelated ordinary session after id reuse. The pagehide path deletes the session through the ordinary DELETE endpoint (apps/workspace-playground/src/front/App.tsx:365-401), while ScriptedSessionStore.delete removes only the record/file (apps/workspace-playground/src/server/testing/scriptedPiHarness.ts:186-197). On restart, the sweep deliberately retains every registry id with no backing record (:316-327). Hydration then resets createCount and takeNextSessionId reuses the first absent id (:301,349-358). Deterministic sequence: create+mark scripted-main; pagehide-delete it; restart; an ordinary empty create reuses scripted-main; restart again; the stale registry entry makes the sweep delete that ordinary draft. The added unresolvable-entry test enshrines retention instead of proving pruning/boundedness (scriptedPiHarness.test.ts:167-180). A long-lived server also accumulates distinct deleted ids indefinitely.

2. Route-based provenance is still user-influenceable through resumeSessionId. The showcase boot reads the candidate from writable sessionStorage (App.tsx:274-287); the wrapper forwards it and marks whatever id the injected create returns (dev.ts:148-170); the gateway returns an existing empty session when that id resolves (packages/agent/src/server/agent-host/embeddedGateway.ts:397-415). Thus a stale/manipulated pointer can mark an ordinary empty session that was not created by the showcase route, for deletion on a later boot.

3. Concurrent wrapper POSTs can lose provenance. markPlaygroundShowcaseSession is an unlocked read-modify-write of the shared JSON file (scriptedPiHarness.ts:461-490), so two requests can read the same set and the last write can drop the other new id, leaving that showcase session permanently untracked.

Verified clean:
- The wrapper is dev-only: dev.ts is wired only through the Vite configureServer hook (vite.config.ts:147-151), so it is absent from production composition.
- The same store cannot sweep a current-boot session: create awaits its one memoized hydrate/sweep before allocating (scriptedPiHarness.ts:169-175,302-313), and the wrapper marks only after the injected create succeeds (dev.ts:155-177).
- pnpm --filter workspace-playground run test: 8 files / 29 tests passed.

No label applied.

hachej:
Addressed the blocking finding on `fix/1452-showcase-session` (commit 78bd184). Disposition:

**1. Registry unbounded / id-reuse can delete an unrelated ordinary session — real, fixed.**
Agreed with the exact sequence in the finding. Root cause: nothing ever removed a registry entry when its session was actually deleted — only the boot-time sweep pruned, and only for entries it could resolve. `ScriptedSessionStore.delete()` (`scriptedPiHarness.ts:186-204`) now calls `unmarkPlaygroundShowcaseSession` as part of the same deletion (`:203`) — not deferred to a future sweep. The pagehide cleanup path (`App.tsx`) deletes a session through this exact method via the ordinary DELETE route, so by the time a freed numeric id could ever be recycled by an unrelated ordinary session, the registry no longer references it at all. New test `unmarks a session on delete, so its numeric id can be safely recycled by an ordinary session later` (`scriptedPiHarness.test.ts`) reproduces your deterministic sequence verbatim — create+mark `scripted-main`, delete it, restart (id recycled by an ordinary create), restart again (sweep runs) — and asserts the recycled ordinary session survives with its real title intact. The prior "unresolvable entry" test is left in place (still a valid case: a genuinely foreign store's entry) but is no longer the only coverage for retention.

**2. resumeSessionId is user-influenceable via writable sessionStorage — real, fixed.**
Agreed. `App.tsx` reads the resume candidate from `sessionStorage`, and `embeddedGateway.createSession` resumes *any* empty session it can resolve regardless of who created it — so a stale or manipulated pointer could get an ordinary session marked. The wrapper route (`dev.ts:147-198`) now only forwards `resumeSessionId` when `isPlaygroundShowcaseSession` confirms it already names a session the wrapper itself previously marked (`dev.ts:168-181`); an unrecognized id is silently dropped, and the boot flow just creates a fresh session instead — always safe. New test `isPlaygroundShowcaseSession only recognizes ids this wrapper actually marked` covers the primitive directly.

**3. Concurrent wrapper POSTs can lose provenance — real, fixed.**
Agreed, the prior mark function was an unlocked read-modify-write. All registry access (mark, unmark, the sweep's own read+prune, and the new `isPlaygroundShowcaseSession` check) now goes through one process-wide promise queue (`queueShowcaseRegistry`, `scriptedPiHarness.ts:507-513`) — Node is single-threaded, so serializing every access through one chain is sufficient; no OS-level file lock needed for a same-process, dev-only sidecar file. The sweep's read and its eventual write are now a *single* queued task (`:328-357`) rather than two separately-queued calls, so a mark/unmark landing between them can no longer be silently clobbered. New test fires 20 concurrent `markPlaygroundShowcaseSession` calls and asserts all 20 land in the registry.

**Verified closures reiterated with current line citations:**
- *Wrapper is dev-only:* `vite.config.ts:149-150`'s `configureServer()` hook is the only place `startPlaygroundServer` (and therefore the route registration in `dev.ts`) is ever invoked — absent from any production composition.
- *Sweep cannot touch a current-boot session:* `create()` (`scriptedPiHarness.ts:169-170`) awaits `ensureHydrated()` — the single memoized hydrate/sweep — before allocating an id, and the wrapper only calls `markPlaygroundShowcaseSession` after the injected create already succeeded (`dev.ts:182-188`), so nothing created this boot can appear in the registry before this store's one-time sweep has already run and completed.

**Proof:**
- `pnpm --filter workspace-playground run test` — 8 files / 32 tests passed (29 + 3 new: the id-recycling exploit-sequence test, the concurrency test, the resumeSessionId-guard test)
- `pnpm --filter workspace-playground run typecheck` — clean
- `showcase.spec.ts` — 2 consecutive full runs against a reused dev server, all green (41.3s, 43.7s)

hachej:
REVISE

One new material blocker remains: showcase provenance is global by bare session id, but scripted session ids are only unique inside an agent/session namespace. The shared registry stores only values such as `scripted-main` (`scriptedPiHarness.ts:470-492`), while each namespaced `ScriptedSessionStore` has its own directory and independently allocates `scripted-main` (`:143-147,364-373`). Consequently, a mark created for one agent namespace is treated as provenance for the same bare id in another namespace. On that second store hydration, the sweep finds its unrelated empty record and deletes it (`:329-360`). The same ambiguity also means `isPlaygroundShowcaseSession` cannot prove the client supplied id was marked for the requested `agentTypeId` (`dev.ts:147-179`, `scriptedPiHarness.ts:558-559`).

Deterministic repro at `78bd1846225a43d18d31869f1fcfe54230713164`: with one temporary session root, create+mark an empty `scripted-main` under namespace `alpha--workspace`; create an ordinary empty draft, also `scripted-main`, under `beta--workspace`; recreate/hydrate the beta store. The beta session list becomes empty and the registry is cleared. This deletes the unrelated beta draft and leaves the alpha showcase untracked. Key registry entries by the owning namespace/agent identity plus session id, and cover cross-namespace retention/resume.

Verified clean in this round:
- `ScriptedSessionStore.delete()` eagerly unmarks, and the mark→delete→restart/recycle→restart/sweep regression closes the same-namespace id-recycle sequence.
- Registry operations, including the complete sweep read-modify-write, use one process-wide promise queue; the 20-concurrent-mark coverage is sound for this same-process dev server.
- The wrapper remains dev-only, and the memoized hydrate-before-create ordering still prevents a current-boot session from being swept.
- `pnpm --filter workspace-playground run test`: 8 files / 32 tests passed.

hachej:
Addressed the new material bug on `fix/1452-showcase-session` (commit df5be27). Disposition:

**Provenance keyed by bare session id, colliding across agent namespaces — real, fixed.** Confirmed the repro: scripted session ids are only unique within one namespaced `ScriptedSessionStore` — each agent type gets its own `sessionDir` (`scriptedPiHarness.ts`, constructor: `join(sessionBaseDir(sessionRoot), sessionNamespace)`) and independently allocates its own `scripted-main`, `scripted-1`, etc. A registry keyed by bare id genuinely couldn't distinguish `alpha--<hash>/scripted-main` from `beta--<hash>/scripted-main`.

Fix: registry entries are now keyed by (agent key, session id), not bare session id.
- Each store derives its own `provenanceAgentKey` in the constructor from `input.sessionNamespace`'s leading segment before the first `--` — which is the raw `agentTypeId` verbatim (`sessionNamespaceForAgent` in `packages/agent` builds the namespace as `[agentTypeId, hash(workspaceScopeId), ns].join('--')` for every agent type this app configures). This is the same parse `scriptedResponseMarker` already used in this file for the response-marker text; both now share one helper, `sessionNamespaceAgentKey`.
- `markPlaygroundShowcaseSession`, `unmarkPlaygroundShowcaseSession`, and `isPlaygroundShowcaseSession` all now take that agent key as a required parameter. Entries are encoded as agentKey + separator + sessionId, using a U+0001 (SOH) control-character separator that no valid agent-type id or scripted session id can ever contain, so decoding by first occurrence is unambiguous.
- The sweep decodes each registry entry and skips any whose agent key isn't its own — this is now a definitive ownership check (the key says so explicitly), not the previous "not found in my records, so probably foreign" inference.
- `dev.ts`'s wrapper route passes `targetAgentTypeId` (the agent the client actually requested, resolved the same way it always was) as the key for both marking on create and validating a client-supplied `resumeSessionId` before ever forwarding it — so a resume can only be honored for a session the wrapper marked under that exact agent type, closing the cross-namespace resume-forgery angle too.

New regression tests reproduce the repro directly (`scriptedPiHarness.test.ts`):
- `does not let a mark for one agent namespace sweep a same-id session in a different agent namespace` — marks an empty `scripted-main` under `sessionNamespace: 'alpha--scopehash'`, creates an ordinary unmarked session that independently allocates the same bare id `scripted-main` under `sessionNamespace: 'beta--scopehash'` (same `sessionRoot`, same registry file), rehydrates the beta store, and asserts beta's draft survives with its real title intact — plus that `isPlaygroundShowcaseSession` still correctly reports the alpha mark as alpha-scoped and not visible under `'beta'`.
- `resumeSessionId validation is scoped per agent key` — a mark under `'alpha'` for a given bare id is not recognized when checked under `'beta'` for the same bare id, directly covering the `dev.ts` guard's actual decision point.

Proof:
- `pnpm --filter workspace-playground run test` — 8 files / 34 tests passed (32 + 2 new cross-namespace tests)
- `pnpm --filter workspace-playground run typecheck` — clean
- `showcase.spec.ts` — 2 consecutive full runs against a reused dev server, all green (1.3m cold, 58.1s warm)

All runs foreground only; no git stash used this round.

hachej:
REVISE

One blocking namespace collision remains: the registry key scopes by agent type, but the scripted store is scoped by the full session namespace, including the workspace-scope hash. `sessionNamespaceForAgent` produces `<agentTypeId>--<hash(workspaceScopeId)>--...` (`packages/agent/src/server/agent-host/sessionInventory.ts:38-46`), while `sessionNamespaceAgentKey` discards everything after the first `--` (`apps/workspace-playground/src/server/testing/scriptedPiHarness.ts:54-56`). Therefore `alpha--hashA/scripted-main` and `alpha--hashB/scripted-main` both encode as `alpha\u0001scripted-main`. A hash-B store sweep can treat hash A provenance as its own and delete a same-id unrelated empty record, or prune hash A provenance when hash B has no such record (`scriptedPiHarness.ts:353-384`). Resume validation has the same ambiguity because `dev.ts:169-172` checks only `(targetAgentTypeId, sessionId)`. The new regression covers alpha versus beta, but not the same agent across two workspace hash suffixes (`scriptedPiHarness.test.ts:270-316`). Key provenance by the full storage namespace/workspace scope as well as agent and session id, and add same-agent/different-workspace sweep and resume coverage.

The U+0001 separator itself is sound: configured agent type ids use the restricted slug grammar and persisted scripted session ids must match `SAFE_NATIVE_SESSION_ID`, so neither valid identifier can contain it.

Proof: `pnpm --filter workspace-playground run test` — 8 files / 34 tests passed; `playwright test apps/workspace-playground/e2e/showcase.spec.ts` — 3/3 passed. Green tests do not cover the remaining hash-scope collision. No label applied.

hachej:
Addressed the remaining collision on `fix/1452-showcase-session` (commit 0a5a8dc). Disposition:

**Agent-key-only scoping missed the workspace-scope hash — real, fixed.** Confirmed: `sessionNamespaceForAgent` builds the storage namespace as `<agentTypeId>--<hash(workspaceScopeId)>--...`, so the SAME agent type serving two different workspace scopes is still two independent stores (`alpha--hashA` vs `alpha--hashB`), each independently allocating its own `scripted-main`. Round 4's agentKey-only registry couldn't distinguish them.

Fix: registry entries are now keyed by (agent key, workspace id, session id).
- Rather than reproducing the internal `sha256(workspaceScopeId)` hash in `dev.ts` (a private algorithm in a different package this route has no business duplicating), the RAW unhashed workspace id is used instead — both sides already have it firsthand: the store receives it per-call as `SessionCtx.workspaceId` (the exact field `belongsTo` already uses to scope ordinary session visibility), and `dev.ts` already has the raw `x-boring-workspace-id` header. Two different raw workspace ids always land in two different registry entries, tracking whichever partition the real hashed namespace would produce — without needing to know the hash algorithm at all.
- The sweep (`scriptedPiHarness.ts`) now requires a full match — agent key AND workspace id AND a record actually found under that id — before ever treating an entry as its own. An entry that doesn't fully resolve is left alone, never pruned: it may belong to a different store (same agent key, different workspace-scope hash) this store has no visibility into. This is a deliberate change from round 4's "prune on not-found" — that logic was only safe when agent key alone was assumed sufficient for ownership, which round 5 disproved. Actual cleanup on deletion still happens eagerly via `delete()`, called by whichever store actually owns the session.
- `dev.ts`'s wrapper route resolves the raw workspace id from the request header (falling back to `"default"`, matching what the showcase route always sends) and uses it, together with `targetAgentTypeId`, for both marking on create and validating a client-supplied `resumeSessionId`.

**New regression tests reproduce the exact repro directly** (`scriptedPiHarness.test.ts`):
- `does not let a mark for one workspace scope sweep a same-agent, same-id session in a different workspace scope` — same agent type (`'alpha'`), two different `sessionNamespace` suffixes (`'alpha--hashA'`/`'alpha--hashB'`) with two different raw `ctx.workspaceId` values, an ordinary unmarked session under hashB independently allocates the same bare id as hashA's marked session, hashB rehydrates, and its draft survives with its real title intact.
- `resumeSessionId validation is scoped per workspace id too` — a mark under `('alpha', 'workspace-A')` is not recognized when checked under `('alpha', 'workspace-B')` for the same bare id, directly covering `dev.ts`'s guard.

**Proof:**
- `pnpm --filter workspace-playground run test` — 8 files / 36 tests passed (34 + 2 new same-agent/different-workspace tests)
- `pnpm --filter workspace-playground run typecheck` — clean
- `showcase.spec.ts` — 2 consecutive full runs against a reused dev server, all green (47.1s, 47.2s)

All runs foreground only; no git stash used this round.

hachej:
REVISE

One blocking scope mismatch remains in the round-6 fix: the registry is keyed by the presented, trimmed `x-boring-workspace-id` (`apps/workspace-playground/src/server/dev.ts:161-162`), but that value is not always the exact `SessionCtx.workspaceId` used by `belongsTo`.

`createWorkspaceAgentServer` accepts both the canonical workspace scope and `basename(workspaceRoot)` as valid selectors (`packages/workspace/src/app/server/createWorkspaceAgentServer.ts:1503-1507`), then `trustedWorkspaceScopeId` validates the presented selector but always returns the canonical `workspaceScopeId` (`:396-429`). That canonical claim becomes `SessionCtx.workspaceId` (`packages/agent/src/server/agent-host/embeddedGateway.ts:72-82`), and `belongsTo` compares the record against it exactly (`apps/workspace-playground/src/server/testing/scriptedPiHarness.ts:412-414`).

Therefore an allowed basename header can create a record scoped to `default` while marking the registry under the basename. The sweep sees `record.workspaceId !== workspaceId` and deliberately leaves the entry untouched (`scriptedPiHarness.ts:363-377`). Repeating accepted basename-scoped wrapper calls across boots can retain both registry entries and stale empty session files indefinitely. The current showcase UI hardcodes `default`, so its normal path avoids this today, but the explicit raw-header/`belongsTo` parity requirement is not met and the no-prune rule turns the mismatch into a real growth path.

Proof: `pnpm --filter workspace-playground run test` — 36/36; focused `showcase.spec.ts` — 3/3. The new unit tests supply matching raw context ids directly and do not cover host selector canonicalization.

No label applied.

hachej:
Addressed the selector/canonical-scope mismatch on `fix/1452-showcase-session` (commit 2f8d276). Disposition:

**Registry keyed by raw header instead of canonical workspace scope — real, fixed.** Confirmed: `createWorkspaceAgentServer` accepts two valid selectors for `x-boring-workspace-id` (the canonical `workspaceScopeId` and `basename(workspaceRoot)`), and `trustedWorkspaceScopeId` always resolves either one to the SAME canonical scope before a session record is created — that canonical value is what becomes `SessionCtx.workspaceId`, which `belongsTo` compares exactly. Keying the registry by whichever raw selector the client presented broke that parity for an allowed basename-selector request.

Fix: the wrapper route no longer derives its registry workspace-id component from the request header at all. `workspaceScopeId` is a SINGLE constant for this server's entire process lifetime (`opts.sessionId ?? "default"` in `createWorkspaceAgentServer.ts`) — computed from the exact same `remoteWorkerWorkspaceId` `dev.ts` already passes as `sessionId` when constructing the server. The wrapper now computes that identical constant once (`canonicalWorkspaceScopeId = remoteWorkerWorkspaceId ?? "default"`) and uses it for both marking on create and validating `resumeSessionId`, independent of whichever selector the client presented. The raw header is still forwarded unchanged to the real create-session endpoint, so that endpoint's own selector validation is untouched — only the registry key changed.

**New e2e test proves belongsTo parity directly**, matching your requested repro: fetches `/api/v1/workspace/meta` (whose `workspaceId` field reflects `basename(workspaceRoot)` in local mode — the OTHER allowed selector, distinct from canonical `"default"`), POSTs to the wrapper route with `x-boring-workspace-id` set to that basename value, confirms the create succeeds (proving the basename selector is indeed accepted), then reads the on-disk provenance registry file directly and decodes the resulting entry — asserting its workspace-id component is `"default"` (canonical), never the presented basename.

**Proof:**
- `pnpm --filter workspace-playground run test` — 8 files / 36 tests passed (unchanged; this fix is HTTP-route-level, not scriptedPiHarness-internal)
- `pnpm --filter workspace-playground run typecheck` — clean
- `showcase.spec.ts` (now 4 tests) — passed twice consecutively (1.0m, 47.9s); one intervening run hit the same pre-existing cold-webServer-boot flake documented in earlier rounds (test 1's 40s wait on a truly cold first compile) — unrelated to this change, and the new test itself passed in all 3 runs.

All runs foreground only; no git stash used this round.

hachej:
SHIP

hachej:
Investigated the UI Review failure (UI_REVIEW_EXPLORATION_STABLE_ACTION_STATE_MISSING). Conclusion: flake, not caused by this PR's changes.

**Evidence:**

1. **Code-level: the showcase route is unreachable from any UI review spec.** `tools/ui-review`'s exploration harness runs exactly one spec in CI (`package.json`'s `ui:review:ci` script: `pnpm run ui:review -- review workspace-command-palette`). Checked every spec under `tools/ui-review/src/review-specs/*/spec.ts` — none reference `showcase`, `SHOWCASE`, or `sessions=`:
   - `workspace-command-palette` (the one CI runs, and the only spec with an `exploration` block): `route: "/?fresh=1"`
   - `workspace-agent-sidebar`: `route: "/?fresh=1"`
   - `ask-user-inline`, `automation-pane-popover`, `workspace-component-baselines`: all `?ui-review-fixture=...` routes
   
   `showcaseBootState` and its `pending`-gated blank shell only ever apply when `isShowcaseRoute()` is true, i.e. `?showcase=1` is present in the URL. None of these routes set it, so that code path is never exercised by the harness at all — this PR's changes cannot be what produced `STABLE_ACTION_STATE_MISSING`.

2. **Local reproduction attempt:** ran the exact command, `cd tools/ui-review && BOMBADIL_TIME_LIMIT=60s UI_REVIEW_OUTPUT_DIR=e2e-artifacts/ui-review UI_REVIEW_CRITIC=fixture UI_REVIEW_VITE_PORT=5480 UI_REVIEW_AGENT_API_PORT=5490 pnpm run ui:review -- review workspace-command-palette`, in the worktree. It did not reach `STABLE_ACTION_STATE_MISSING` at all — the build and dev server came up fine (confirmed the `/?fresh=1` boot path, including a real POST-created session, all succeeded), but the bombadil browser itself failed to launch: `Failed to create a ProcessSingleton for your profile directory` (`bombadilProcess.ts` already explicitly lists this exact string as a *known retryable startup flake* — it retried once, failed again, and gave up). This is a shared-VM resource-contention issue (this box runs several concurrent agent sessions with their own headless Chrome instances), unrelated to application code, and not the same failure signature as the CI error.

3. **The re-run is green.** `gh pr checks 1458` now shows `UI Review pass 8m48s`, alongside all other checks passing.

No code change made — nothing to fix. Worktree is clean (no diff).

PR #1468, opened 2026-08-29
fix(core): leave navigates away; settings route honors membership guard

### LINKED ISSUE #1466: Workspace UI guards: Leave leaves stale members UI; /settings renders shell + Delete button for non-members
MEDIUM, two related UI-guard gaps (server authz is correct in both — verified 403s). (a) Non-owner Leave: DELETE succeeds, but the members page keeps the cached list, shows 'Failed to load members.' and a dead Leave button instead of redirecting away. (b) /workspace/:id/settings for a workspace the user is NOT a member of renders the full settings shell stuck on 'Loading role' with an ENABLED red Delete workspace button, while /workspace/:id and /members correctly show the 'Not a member' 403 screen. Fix: on successful Leave navigate to / (or the next workspace) and invalidate queries; settings route must use the same membership guard as the other workspace routes.

hachej:
Fixes #1466

## Summary
- MembersPage: on a successful self-leave, invalidate the workspaces query and navigate to `/` instead of refetching the now-forbidden members list — fixes the stale list / "Failed to load members." / dead Leave button after leaving as a non-owner.
- WorkspaceSettingsPage: apply the same route-status membership guard used by the main `/workspace/:id` shell (extracted the shared `WorkspaceRouteErrorPage` out of `CoreWorkspaceAgentFront` into `packages/core/src/front/workspace/`). A non-member now sees the "Not a member" 403 screen instead of a stuck "Loading role" settings shell, and the whole shell (including Delete) is withheld until the route status is `matched`.
- Fixed `canDeleteWorkspace = role === 'owner' || role === null` → `role === 'owner'`: this bug is what let the Delete workspace button render enabled while role was still unresolved (or forbidden).

Server authz for both cases was already correct — these are pure UI-guard fixes.

## Test plan
- [x] `pnpm -C packages/core exec vitest run src/front --no-file-parallelism` — 33 files / 240 tests passed, including new coverage:
  - MembersPage: self-leave navigates to `/`, no stale refetch of the forbidden members query
  - WorkspaceSettingsPage: forbidden route status → access-denied screen, no shell; non-matched route status → no Delete button rendered; matched-but-non-owner → Delete button rendered but disabled
- [x] `pnpm -C packages/core run typecheck` — no new errors (one pre-existing, unrelated error in `packages/workspace/src/app/front/WorkspaceAgentFront.tsx` also present on `origin/main`)

### DIFF
--- a/packages/core/src/app/front/CoreWorkspaceAgentFront.tsx
+++ b/packages/core/src/app/front/CoreWorkspaceAgentFront.tsx
@@ -10,6 +10,7 @@ import {
   useCurrentWorkspace,
   useSession,
   useWorkspaceRouteStatus,
+  WorkspaceRouteErrorPage,
   type CoreFrontAuthPagesOverride,
   type CoreFrontCompanyAdminOptions,
 } from '../../front/index.js'
@@ -226,22 +227,6 @@ function HomeRedirect<TSession extends WorkspaceAgentSession = WorkspaceAgentSes
   }
   if (!workspace) return <>{resolvedLoadingFallback}</>
   return <Navigate to={workspaceHref(workspace.id)} replace />
-}
-
-function WorkspaceRouteErrorPage({ status, message }: { status: 'not-found' | 'forbidden' | 'switch-failed'; message: string }) {
-  const title = status === 'not-found'
-    ? 'Workspace not found'
-    : status === 'forbidden'
-      ? 'Workspace unavailable'
-      : 'Workspace failed to open'
-  return (
-    <div className="flex h-screen min-h-0 items-center justify-center bg-background px-6 text-foreground">
-      <div className="w-full max-w-md rounded-2xl border border-border bg-card p-6 text-center shadow-sm">
-        <h1 className="text-lg font-semibold">{title}</h1>
-        <p className="mt-2 text-sm text-muted-foreground">{message}</p>
-      </div>
-    </div>
-  )
 }
 
 function WorkspaceRoute<
--- a/packages/core/src/app/front/CoreWorkspaceAgentFront.tsx
+++ b/packages/core/src/app/front/CoreWorkspaceAgentFront.tsx
@@ -10,10 +10,16 @@ import {
   useCurrentWorkspace,
   useSession,
   useWorkspaceRouteStatus,
-  WorkspaceRouteErrorPage,
   type CoreFrontAuthPagesOverride,
   type CoreFrontCompanyAdminOptions,
 } from '../../front/index.js'
+// Imported directly from its module (not the `../../front/index.js` barrel):
+// this is a pure presentational component with no hooks/context, and
+// CoreWorkspaceAgentFront.test.tsx replaces the whole barrel with a manual
+// mock. Routing it through the barrel would silently render `undefined`
+// under that mock instead of the real terminal-route screen the test
+// (added by #1435) asserts on.
+import { WorkspaceRouteErrorPage } from '../../front/workspace/WorkspaceRouteErrorPage.js'
 import {
   parseFullPagePanelLocation,
   WorkspaceAgentFront,
--- a/packages/core/src/front/__tests__/MembersPage.leaveIntegration.test.tsx
+++ b/packages/core/src/front/__tests__/MembersPage.leaveIntegration.test.tsx
@@ -1,0 +1,263 @@
+// @vitest-environment jsdom
+//
+// Integration regression for the self-leave -> "/" redirect, using the real
+// WorkspaceAuthProvider (unmocked) instead of MembersPage.test.tsx's mocked
+// provider. This is the only way to catch the race a thermo review flagged:
+// invalidateQueries() alone does not clear the *current* cached workspaces
+// list/detail before navigate('/') runs, so WorkspaceAuthProvider (which
+// resolves "/" synchronously off whatever is already cached) can redirect
+// straight back into the workspace the user just left.
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { render, screen, waitFor, fireEvent } from '@testing-library/react'
+import { MemoryRouter, Route, Routes } from 'react-router-dom'
+import { useEffect } from 'react'
+import { afterEach, beforeAll, describe, expect, it, vi } from 'vitest'
+
+import { withTaskId } from '../../server/__tests__/_setup'
+import type { Workspace } from '../../shared/types'
+import { MembersPage } from '../workspace/MembersPage'
+import { useCurrentWorkspace, WorkspaceAuthProvider } from '../WorkspaceAuthProvider'
+import { useMswHandler } from './_setup'
+
+const TASK_ID = 'boring-ui-v2-lv1n'
+
+const OWNER_ID = 'user-owner'
+const EDITOR_ID = 'user-editor'
+
+const WS_1: Workspace = {
+  id: 'ws-001',
+  appId: 'test-app',
+  workspaceTypeId: 'default',
+  name: 'Workspace One',
+  createdBy: OWNER_ID,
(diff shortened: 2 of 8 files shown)


hachej:
REVISE

Blocking: self-leave can bounce straight back into the workspace that was just left. `MembersPage` starts `invalidateQueries(WORKSPACES_QUERY_KEY)` without awaiting it or updating the cached list, then immediately navigates to `/`. In the real app `WorkspaceAuthProvider` remains mounted, so `/` still resolves synchronously from the stale workspaces list and cached workspace detail; `HomeRedirect` can immediately replace the route with the just-left workspace. This occurs today when that workspace is the cached default/first entry. With PR #1454, it becomes the expected path whenever the just-left workspace is the remembered workspace: its membership validation sees the stale list before the refetch completes and selects it again. The 60-second detail `staleTime` can then preserve the stale workspace shell instead of cleanly falling through.

The new MembersPage test cannot catch this integration failure because it mocks `WorkspaceAuthProvider` and makes `/` a static `home` div; its fetch assertion only counts the members endpoint. Make removal from the workspace-list/detail cache atomic before navigation, or await a fresh membership list before navigating, and add an integrated provider/HomeRedirect regression covering a remembered just-left workspace and asserting selection of the next/default workspace.

The extracted `WorkspaceRouteErrorPage` is behavior-identical to the prior inline screen, the settings `loading -> matched` path has no legitimate-owner permanent-placeholder state, and `canDeleteWorkspace = role === owner` correctly fails closed. I found no weakened existing assertions, but the new mocked navigation test provides false confidence for the real redirect path.

Proof run: `pnpm -C packages/core exec vitest run src/front/__tests__/MembersPage.test.tsx src/front/__tests__/WorkspaceSettingsPage.test.tsx --no-file-parallelism` — 2 files passed, 21 tests passed.

hachej:
Addressed the thermo REVISE, per-finding:

**1. Blocking: self-leave can bounce back into the just-left workspace — REAL DEFECT, fixed.**
Confirmed the mechanism exactly as described: `WorkspaceAuthProvider` stays mounted across the `navigate('/')` and resolves `/` synchronously from whatever is already cached. `invalidateQueries` alone doesn't clear the stale `workspaces` list or the just-left workspace's still-cached detail before the redirect runs, so a real (non-instant) refetch loses the race.

Fix (`packages/core/src/front/workspace/MembersPage.tsx`): the self-leave branch now calls `queryClient.setQueryData(WORKSPACES_QUERY_KEY, ...)` to atomically filter the just-left workspace out of the cached list *before* `navigate('/')`, in the same `onSuccess` callback. `invalidateQueries` still runs afterward to reconcile with the server.

Added `packages/core/src/front/__tests__/MembersPage.leaveIntegration.test.tsx` — a real (unmocked) `WorkspaceAuthProvider` + Home-probe integration test. Critically, it holds the invalidate-triggered `GET /workspaces` refetch open on a deferred promise the test controls explicitly, so the assertions run *before* that network round trip can resolve — a same-tick-resolving mock would have hidden this exact race (my first attempt at this test did, and passed against the old buggy code; I caught that by deliberately reverting the fix and confirming the test now fails, then restored it and confirmed green). Proof:
- Reverted the fix locally → test fails: `expected true to be false` on the cached-list assertion, i.e. it reproduces the redirect-back bug.
- Fix restored → test passes.

**2. Extracted `WorkspaceRouteErrorPage` byte-identical to the prior inline screen — confirmed, no action needed.**
Reviewer independently verified this; agreed, it's a pure extraction (same JSX/classes/text), only the import site changed.

**3. Legitimate-owner permanent-placeholder state in settings `loading -> matched` — confirmed none exists, no action needed.**
Reviewer independently verified this; agreed — `routeStatus` only stays off `matched` while `workspacesQuery`/`detailQuery` are genuinely pending or in a terminal not-found/forbidden/switch-failed state, all of which resolve or are already handled.

**4. `canDeleteWorkspace = role === 'owner'` fails closed correctly — confirmed, no action needed.**

**5. Weakened assertions — none found by the reviewer; I didn't introduce or weaken any in this round either.**

Proof run (foreground, this round):
- `pnpm -C packages/core exec vitest run src/front src/app/front/__tests__/CoreWorkspaceAgentFront.test.tsx --no-file-parallelism` → **34 files / 241 tests passed** under `src/front` (includes the new integration test and all prior MembersPage/WorkspaceSettingsPage coverage). `CoreWorkspaceAgentFront.test.tsx` fails to even transform in this worktree — `Failed to resolve import "@hachej/boring-bash/shared"` from `packages/workspace`'s stale/broken dist — pre-existing and unrelated to this PR (packages/workspace's own `dts` build is independently broken by a fastify version mismatch in this worktree; confirmed on `origin/main` too).
- `pnpm -C packages/core run typecheck` → exit 2, but the only error is the pre-existing, unrelated `packages/workspace/src/app/front/WorkspaceAgentFront.tsx:568` `setArchived` error, present identically on `origin/main`. No errors in any file this PR touches.

Commit: c46d457ab. Pushed to `fix/1466-workspace-ui-guards`.

hachej:
SHIP

hachej:
Fixed the CI red on "Unit Tests Changed" (\`CoreWorkspaceAgentFront.test.tsx\` — the 3 \`renders terminal route recovery for not-found/forbidden/switch-failed\` cases from #1435).

**Cause:** not a merge-base issue — \`git merge origin/main\` reported "Already up to date" (\`#1435\`'s commit 40b934624 is already an ancestor of this branch). The real cause: \`CoreWorkspaceAgentFront.test.tsx\` replaces the entire \`../../front/index.js\` barrel with a manual \`vi.mock(...)\` object listing an explicit set of exports. My earlier extraction re-exported \`WorkspaceRouteErrorPage\` from that same barrel and had \`CoreWorkspaceAgentFront.tsx\` import it from there. Under that test's mock, \`WorkspaceRouteErrorPage\` silently resolved to \`undefined\` instead of the real component, so the terminal route-error screen never rendered and the 3 recovery-status tests failed.

**Fix** (commit 58bee1e03): import \`WorkspaceRouteErrorPage\` directly from its own module (\`../../front/workspace/WorkspaceRouteErrorPage.js\`) instead of through the barrel — it's a pure presentational component with no hooks/context, so it was never meant to be part of the front-primitives surface this test intentionally swaps for doubles. This restores exactly the pre-extraction behavior (a locally-available, never-mocked component) without touching the test file, matching the instruction to adjust the extraction rather than the test.

Note for anyone reproducing locally in this worktree: the test initially failed to even transform here (\`Failed to resolve import "@hachej/boring-bash/shared"\` / \`"@hachej/boring-agent/front"\`) because \`packages/workspace\`'s dist was stale/broken — its own \`dts\` build is blocked by an unrelated fastify version mismatch in this checkout. I rebuilt \`packages/boring-bash\`, \`packages/agent\`, and \`packages/workspace\` (tsup with \`--no-dts\` + \`vite build\`) to actually exercise the test; that's a pre-existing local-env issue, not something this PR caused, and not something CI hits (its \`packages/workspace\` build presumably succeeds there).

**Proof (foreground):**
- \`pnpm -C packages/core exec vitest run src/app/front/__tests__/CoreWorkspaceAgentFront.test.tsx --no-file-parallelism\` → **30/30 passed**, including all 3 previously-red \`renders terminal route recovery for ...\` cases.
- \`pnpm -C packages/core exec vitest run src/front src/app/front/__tests__/CoreWorkspaceAgentFront.test.tsx --no-file-parallelism\` → **35 files / 271 tests passed**.
- \`pnpm -C packages/core run typecheck\` → **exit 0**, clean (the previously-reported pre-existing \`setArchived\` error also cleared once \`packages/agent\`'s stale dist was rebuilt).

Pushed to \`fix/1466-workspace-ui-guards\` (58bee1e03).

Issue #1464, opened 2026-08-29
Email links don't match SPA routes: invite-accept points at raw API JSON, password-reset lands on a blank page


hachej:
HIGH (two flows, one bug class). (1) Invite email link = http://<api>/api/v1/workspaces/:id/invites/:inviteId/accept?invite_token=… (built at packages/core/src/server/routes/invites.ts:83 from app.config.auth.url + the API path) → unauthenticated invitee sees a bare {"error":"Authentication required"} JSON blob. The SPA route /invites/:token exists and works (prompts sign-in) — the email must link there. (2) Password-reset email link = better-auth's default path-token URL /auth/reset-password/<token>?callbackURL=…, but ResetPasswordPage.tsx reads token from the QUERY string at /auth/reset-password — no route matches the path form, React Router renders nothing → blank white page. Built at packages/core/src/server/auth/createAuth.ts:177-186 passing data.url straight through. Fix: rewrite both links to the SPA route shapes before templating (single helper), with tests asserting the exact emitted URLs; consider a redirect route for the legacy shapes so in-flight emails still work.

hachej:
Fixed by merged PR #1471 (email links target SPA routes; verified e2e — invite link renders the SPA sign-in prompt, reset link renders the working form). Auto-close keyword didn't fire; closing manually.

PR #1471, opened 2026-08-29
#1464 fix(core): email links target SPA routes (invite accept, password reset)

hachej:
## Summary
- Invite-accept emails now link to the SPA route `/invites/:token` on the front origin instead of the raw API endpoint (`/api/v1/workspaces/:id/invites/:id/accept`), which an unauthenticated invitee saw as a bare JSON error.
- Password-reset emails now link to the SPA route `/auth/reset-password?token=...` instead of better-auth's default path-token shape (`/auth/reset-password/<token>`), which `ResetPasswordPage` (query-string based) could not render, showing a blank page.
- Both links are built by a new shared helper (`packages/core/src/server/mail/links.ts`) that targets the front origin — `auth.frontUrl` (new optional config field, `FRONTEND_URL` env var), falling back to `auth.url` when the front and API share an origin (the default in production).
- Added a compatibility redirect for the legacy path-token reset shape (`/auth/reset-password/:token` → `?token=...`), both server-side (production SPA shell routes) and as a front SPA route (dev, where Vite serves the shell directly), so already-sent emails keep working.

## Files
- `packages/core/src/server/mail/links.ts` (new) — `buildInviteAcceptUrl`, `buildResetPasswordUrl`
- `packages/core/src/server/mail/__tests__/links.test.ts` (new) — exact-URL unit tests for both helpers
- `packages/core/src/server/routes/invites.ts` — invite email now uses `buildInviteAcceptUrl`
- `packages/core/src/server/auth/createAuth.ts` — reset-password email now uses `buildResetPasswordUrl(config, data.token)`
- `packages/core/src/shared/types.ts`, `packages/core/src/server/config/schema.ts`, `packages/core/src/server/config/loadConfig.ts` — new optional `auth.frontUrl` config field, sourced from `FRONTEND_URL`
- `packages/core/src/app/server/createCoreWorkspaceAgentServer.ts` (+ test) — server-side redirect for the legacy `/auth/reset-password/:token` shape
- `packages/core/src/front/CoreFront.tsx` (+ test) — front SPA route for the same legacy shape
- `apps/full-app/.env.example` — documents `FRONTEND_URL`

## Test plan
- [x] `vitest run` on the touched suites: `mail/__tests__/links.test.ts`, `routes/__tests__/invites.test.ts`, `app/server/__tests__/createCoreWorkspaceAgentServer.test.ts`, `front/__tests__/CoreFront.test.tsx`, `front/__tests__/ResetPasswordPage.test.tsx` — all green (72 tests)
- [x] `tsc --noEmit` in `packages/core` — clean except one pre-existing, unrelated error in `packages/workspace` (confirmed present on `origin/main` via `git stash`)
- [x] End-to-end in full-app dev (own worktree, ports 7411/7412 API/front, own Postgres db+role `fullapp_1464`, `MAIL_TRANSPORT_URL=console://`):
  - Signed up, verified email, created a workspace invite → captured email link is `http://localhost:7412/invites/<token>` (front origin) → unauthenticated browser renders the SPA "Sign in to accept this invite" prompt (200, `text/html`), not raw JSON.
  - Requested a password reset → captured email link is `http://localhost:7412/auth/reset-password?token=...` → browser renders the reset form, submitting a new password redirects to sign-in, and the new password successfully authenticates (`sign-in/email` → 200).
  - Legacy path-token shape (`/auth/reset-password/<token>`) redirects client-side to the query-string route and still renders the form.
- [ ] Full `vitest run` across all of `packages/core` was NOT run to completion clean — pre-existing, unrelated failures exist on `origin/main` in this sandbox (Postgres schema drift on `accounts.issuer`, model-budget ordering, an esbuild-timeout test) confirmed via `git stash` before/after comparison; none touch the changed files.

## Screenshots
- Invite accept (unauthenticated): sign-in prompt, not JSON
- Reset-password form rendering from the SPA link

### DIFF
--- a/apps/full-app/.env.example
+++ b/apps/full-app/.env.example
@@ -10,6 +10,11 @@ DATABASE_URL=postgres://postgres:postgres@localhost:5432/boring_ui_v2
 BETTER_AUTH_SECRET=0000000000000000000000000000000000000000000000000000000000000000
 BETTER_AUTH_URL=http://localhost:3000
 CORS_ORIGINS=http://localhost:3000,http://localhost:5173
+# Public origin of the SPA front-end, for email links (invite accept, password
+# reset). Only needed when the front runs on a different origin than the API,
+# e.g. `pnpm dev` (Vite on FRONTEND_PORT, API on PORT). Unset in production,
+# where the front is served from the API origin (BETTER_AUTH_URL).
+# FRONTEND_URL=http://localhost:5173
 
 # Workspace settings encryption (32-byte hex)
 WORKSPACE_SETTINGS_ENCRYPTION_KEY=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
--- a/packages/core/src/app/server/__tests__/createCoreWorkspaceAgentServer.test.ts
+++ b/packages/core/src/app/server/__tests__/createCoreWorkspaceAgentServer.test.ts
@@ -6,6 +6,7 @@ import { describe, expect, it, vi } from 'vitest'
 import { noopTelemetry } from '../../../shared/telemetry.js'
 import {
   createCoreWorkspaceAgentServer,
+  registerFrontendAuthPages,
   registerFrontendFallback,
   resolveCoreLoadConfigOptions,
   type CoreFrontendRootHandler,
@@ -88,5 +89,35 @@ describe('createCoreWorkspaceAgentServer', () => {
     expect(await requestPath(rootOnly, '/workspace')).toEqual(baseline)
     expect(rootOnly).not.toHaveBeenCalled()
     expect(baseline).toEqual({ body: '<!doctype html><p>spa shell</p>', contentType: 'text/html; charset=utf-8', cache: 'no-store' })
+  })
+
+  it('redirects the legacy /auth/reset-password/:token shape to the SPA query-string route', async () => {
+    const appRoot = await mkdtemp(`${tmpdir()}/boring-core-reset-redirect-`)
+    await mkdir(resolve(appRoot, 'dist/front'), { recursive: true })
+    await writeFile(resolve(appRoot, 'dist/front/index.html'), '<!doctype html><p>spa shell</p>')
+
+    const app = Fastify()
+    await registerFrontendAuthPages(app, appRoot, noopTelemetry)
+
+    const response = await app.inject({ method: 'GET', url: '/auth/reset-password/abc123' })
+    await app.close()
+
+    expect(response.statusCode).toBe(302)
+    expect(response.headers.location).toBe('/auth/reset-password?token=abc123')
+  })
+
+  it('URL-encodes the token when redirecting the legacy reset-password shape', async () => {
+    const appRoot = await mkdtemp(`${tmpdir()}/boring-core-reset-redirect-`)
+    await mkdir(resolve(appRoot, 'dist/front'), { recursive: true })
+    await writeFile(resolve(appRoot, 'dist/front/index.html'), '<!doctype html><p>spa shell</p>')
+
+    const app = Fastify()
+    await registerFrontendAuthPages(app, appRoot, noopTelemetry)
+
+    const response = await app.inject({ method: 'GET', url: '/auth/reset-password/a%20b%26c' })
+    await app.close()
+
+    expect(response.statusCode).toBe(302)
+    expect(response.headers.location).toBe('/auth/reset-password?token=a%20b%26c')
   })
 })
--- a/packages/core/src/app/server/__tests__/createCoreWorkspaceAgentServer.test.ts
+++ b/packages/core/src/app/server/__tests__/createCoreWorkspaceAgentServer.test.ts
@@ -91,7 +91,24 @@ describe('createCoreWorkspaceAgentServer', () => {
     expect(baseline).toEqual({ body: '<!doctype html><p>spa shell</p>', contentType: 'text/html; charset=utf-8', cache: 'no-store' })
   })
 
-  it('redirects the legacy /auth/reset-password/:token shape to the SPA query-string route', async () => {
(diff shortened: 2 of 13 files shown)


hachej:
REVISE

Two blockers:

1. `auth.frontUrl` is not validated as the contract claims. `packages/core/src/server/config/schema.ts:88` uses `z.string().url()`, which accepts values such as `ftp://front.example`, `https://front.example/base`, and URLs with query/fragment/userinfo. `mail/links.ts:18-20` only strips trailing slashes, so a path-bearing value produces a wrongly rooted SPA link and query/fragment inputs produce broken links. Please validate an absolute `http:` or `https:` origin only (root pathname, no search/hash/userinfo), and construct from the parsed URL/origin. `FRONTEND_URL` is operator config, not request/user input, so I found no runtime user-driven open redirect; the missing boundary validation is still a release blocker.

2. The production legacy reset route breaks Better Auth callback semantics. `registerFrontendAuthPages` registers `GET /auth/reset-password/:token` before the auth proxy (`createCoreWorkspaceAgentServer.ts:991-995,1209-1213`), so it shadows Better Auth, discards `callbackURL`, and redirects invalid/expired tokens as if valid. Better Auth normally origin-checks `callbackURL`, checks token existence/expiry, then redirects to that callback with `token`. Prefer leaving the production legacy URL to the auth proxy and keeping only the Vite SPA compatibility route, or preserve the full validated callback/error behavior. Add an integrated test that proves route ownership, callback preservation, and invalid/expired-token behavior; the current isolated registrar tests encode the regression.

Security notes: the final password mutation still sends the token to Better Auth, which atomically consumes and validates it server-side; nothing trusts the query param to change a password. Tokens remain present in request URLs and the Fastify request serializer logs `req.url`; this exposure existed for legacy path links too, but sensitive-route URL redaction and `Referrer-Policy: no-referrer` would be worthwhile hardening.

Deployment check: the fallback to `auth.url` is correct for both the full-app production shape and the actual Seneca Docker/Caddy deployment: both serve the built SPA and API on the canonical `BETTER_AUTH_URL` origin. Seneca sets `BETTER_AUTH_URL=https://${SENECA_HOSTNAME}` and proxies that hostname to the single app container.

Test honesty: targeted command passed 5 files / 72 tests, but the link tests instantiate config directly and omit invalid-origin cases, while the redirect tests register only the frontend routes and therefore cannot reveal that they shadow Better Auth.

Proof: `pnpm --filter @hachej/boring-core exec vitest run src/server/mail/__tests__/links.test.ts src/server/routes/__tests__/invites.test.ts src/app/server/__tests__/createCoreWorkspaceAgentServer.test.ts src/front/__tests__/CoreFront.test.tsx src/front/__tests__/ResetPasswordPage.test.tsx` — 5 files passed, 72 tests passed.

hachej:
Addressed both blockers in 0c1072798.

**1. `frontUrl` validation/normalization — FIXED (real defect).**
Replaced `z.string().url()` with a strict `originSchema` in `packages/core/src/server/config/schema.ts` (http/https only, `pathname === '/'`, empty search/hash/username/password). `mail/links.ts:frontOrigin()` now builds from `new URL(base).origin` instead of trailing-slash trimming, so any path/query/fragment/userinfo is stripped regardless of the source (applies to `auth.url` too, which isn't origin-restricted at the schema level). Added `loadConfig.test.ts` cases rejecting a path, query, fragment, userinfo, non-http(s) scheme, relative value, and garbage; `links.test.ts` gained a case proving extras get stripped even if something bypassed schema validation. Agreed there's no runtime user-driven open redirect (`FRONTEND_URL` is operator config) — fixed as a boundary-validation / fail-fast issue per your framing, not as a redirect vuln.

**2. Legacy reset route shadowing better-auth — FIXED (real defect, thank you for the catch).**
Confirmed in `better-auth/dist/api/routes/password.mjs:92-121`: `/auth/reset-password/:token` (GET) is `resetPasswordCallback` — a real endpoint that looks up the verification token, checks expiry, and redirects to `callbackURL` (or an error redirect for `INVALID_TOKEN`). My `registerFrontendAuthPages` addition registered a blanket 302 in front of the auth proxy, so it exactly matched your description: discarded `callbackURL`, and "succeeded" for invalid/expired tokens. Removed it entirely (`createCoreWorkspaceAgentServer.ts`) — that path now falls through to `registerAuthProxy`'s `/auth/*` wildcard, so better-auth's own validated redirect runs, matching pre-existing behavior. Kept only the front SPA route (`CoreFront.tsx`'s `ResetPasswordLegacyRedirect`), which is client-side-only and can't shadow a server-validated request — it only fires once the SPA shell is already being rendered (the two-origin dev setup), and the actual password mutation still round-trips through server-side token validation on submit. Replaced the two tests that encoded the regression with a route-ownership test: `registerFrontendAuthPages` now asserted to 404 on `/auth/reset-password/:token` (not intercepted) while still serving the exact `/auth/reset-password` shell page.

**Deployment check (auth.url fallback) — no change, confirmed correct as reported.** Left as-is.

**Security note (token in URLs/logs, `Referrer-Policy`) — acknowledged, not implemented.** Pre-existing exposure (not introduced by this PR) and explicitly framed as hardening rather than a blocker; leaving for a follow-up rather than scope-creeping this fix.

**Test honesty — addressed.** The two categories you called out (invalid-origin cases, and redirect tests that couldn't reveal the shadowing) are exactly what the new tests target.

Proof:
```
pnpm --filter @hachej/boring-core exec vitest run \
  src/server/mail/__tests__/links.test.ts \
  src/server/config/__tests__/loadConfig.test.ts \
  src/server/routes/__tests__/invites.test.ts \
  src/app/server/__tests__/createCoreWorkspaceAgentServer.test.ts \
  src/front/__tests__/CoreFront.test.tsx \
  src/front/__tests__/ResetPasswordPage.test.tsx
# 6 files passed, 132 tests passed
```
`tsc --noEmit` in `packages/core`: clean (exit 0).

hachej:
SHIP

PR #1472, opened 2026-08-29
Fix: focus session created via 'Start new chat with <Agent>'

### LINKED ISSUE #1470: 'Start new chat with <Agent>' creates the session but doesn't open/focus it
Playground round-3 hunt: in the composer's agent picker ('Choose Agent for new chat' → pick another agent → 'Start new chat with <Agent>'), the new session appears in the sidebar but the main pane stays on the previously active chat; the user must click the new Untitled row. The empty-state 'New chat with Alpha/Beta' buttons DO focus the new pane correctly — mirror that behavior. Severity medium: breaks the primary agent-switch affordance. Screenshot 6x-06-bug-new-chat-no-focus.png in the session scratchpad.

hachej:
Fixes #1470

## Summary
- The New chat Agent picker ("Choose Agent for new chat" → pick another agent → "Start new chat with <Agent>") retargets creation to a different Agent than the one currently addressed, but never updated the addressed selection itself.
- The pane-reconciliation effect resolves its desired session from the ADDRESSED Agent's own session API, so the freshly created pane was immediately stomped back to whatever chat was previously open — the session was created (visible in the sidebar) but never focused.
- Fix lives in the single shared creation transaction (`createChatPaneTransaction` in `WorkspaceAgentFront.tsx`) used by every creation entry point (rail "+", agent card "+", and the picker), so the fix applies uniformly rather than special-casing the picker.

## Change
- `packages/workspace/src/app/front/WorkspaceAgentFront.tsx`: after a fleet-mode session is created for an Agent other than the currently addressed one, call `addressedAgents.selectAgentTypeId(createdAgentTypeId)` so the addressed selection follows the newly created session.
- `packages/workspace/src/app/front/__tests__/WorkspaceAgentFront.test.tsx`: added a regression test — retarget the picker to Beta without ever addressing Beta, click "Start new chat with Beta", assert the rendered chat pane's `data-agent-type-id`/`data-session-id` now point at the new Beta session.

## Test plan
- [x] `pnpm -C packages/workspace vitest run src/app/front/__tests__/WorkspaceAgentFront.test.tsx` — 100/100 pass (new test included)
- [x] `pnpm -C packages/workspace vitest run src/front/layout/plugin-tabs/__tests__/AppLeftPane.test.tsx` — 41/41 pass
- [x] `pnpm -C packages/workspace tsc --noEmit` — clean
- [x] `pnpm -C packages/agent tsc --noEmit` — clean (package untouched by this fix)
- Full monorepo `vitest run` shows some failures in unrelated server-side/agent suites under full-parallel load; verified these are pre-existing/environmental (reproduced in isolation without this change touching those files, and they pass when run in isolation).

### DIFF
--- a/packages/workspace/src/app/front/WorkspaceAgentFront.tsx
+++ b/packages/workspace/src/app/front/WorkspaceAgentFront.tsx
@@ -1807,12 +1807,28 @@ export function WorkspaceAgentFront<
       if (remoteSessionsPending && current.ids.length > 0 && !pendingCreatedId) return current
       const currentActiveRef = current.activeId ? workspaceSessionRefFromKey(current.activeId) : undefined
       const activeOwnerIsExplicit = Boolean(effectiveActiveSessionAgentTypeId)
-      const currentMatchesControlledSession = activeOwnerIsExplicit
+      // #1472 review: a fleet create's addressed-Agent switch and its pane
+      // update land in the same tick, but the newly addressed Agent's OWN
+      // session snapshot (`addressedFleetSessions.tsx`'s `snapshots` map)
+      // only catches up a render or two later, via that Agent's session
+      // source publishing through its own passive effect. In the gap,
+      // `chatSessionKey` still names the Agent's PREVIOUS active session, so
+      // without this the optimistic pane got judged "uncontrolled" and
+      // replaced by that stale session — a `new → stale → new` visible stomp.
+      // Trust an optimistically-created pane over the (possibly stale)
+      // controlled session until its own key is cleared from the ref above
+      // (once the real session list actually contains it).
+      const currentIsOptimisticCreate = Boolean(current.activeId && optimisticCreatedPaneKeysRef.current.has(current.activeId))
+      const currentMatchesControlledSession = currentIsOptimisticCreate || (activeOwnerIsExplicit
         ? current.activeId === chatSessionKey
-        : currentActiveRef?.sessionId === chatSessionId
+        : currentActiveRef?.sessionId === chatSessionId)
       const resolvedDesiredSessionId = !pendingCreatedId
         && current.activeId
-        && (!canPruneMissingSessions || resolvedSessionsByKey.has(current.activeId))
+        // An optimistic create bypasses the "known session" gate too: it is
+        // by definition not in `resolvedSessionsByKey` yet (that is what
+        // "optimistic" means here), so requiring membership would undo the
+        // protection `currentMatchesControlledSession` just granted it.
+        && (!canPruneMissingSessions || resolvedSessionsByKey.has(current.activeId) || currentIsOptimisticCreate)
         && currentMatchesControlledSession
         ? current.activeId
         : desiredSessionId
--- a/packages/workspace/src/app/front/WorkspaceAgentFront.tsx
+++ b/packages/workspace/src/app/front/WorkspaceAgentFront.tsx
@@ -2022,6 +2022,17 @@ export function WorkspaceAgentFront<
         if (createdAgentTypeId) rawSwitch(id, createdAgentTypeId)
         else rawSwitch(id)
       }
+      // A fleet create can target an Agent OTHER than the one currently
+      // addressed (#1470: the New chat Agent picker retargets independently
+      // of the addressed selection). The pane-reconciliation effect below
+      // resolves its desired session from the ADDRESSED Agent's own session
+      // API, so without this the new pane got immediately stomped back to
+      // whatever the previously addressed chat was — created, but never
+      // shown. Address the created Agent so every creation entry point (the
+      // rail "+", the card, and the picker alike) opens on what it just made.
+      if (fleetModeEnabled && createdAgentTypeId && createdAgentTypeId !== addressedAgents.selectedAgentTypeId) {
+        addressedAgents.selectAgentTypeId(createdAgentTypeId)
+      }
       scheduleActiveAgentComposerFocus()
     }).catch(() => {
       settleIfOwner()
@@ -2029,7 +2040,7 @@ export function WorkspaceAgentFront<
       // transaction only owns clearing its pending state.
     })
     return created
-  }, [chatSessionKey, rawSwitch, resolvedCreate, resolvedSessions, selectedAgentTypeId, sessionApi, workspaceId])
+  }, [addressedAgents.selectAgentTypeId, addressedAgents.selectedAgentTypeId, chatSessionKey, fleetModeEnabled, rawSwitch, resolvedCreate, resolvedSessions, selectedAgentTypeId, sessionApi, workspaceId])
 
   useEffect(() => {
     const pending = pendingCreatePaneRef.current
--- a/packages/workspace/src/app/front/WorkspaceAgentFront.tsx
+++ b/packages/workspace/src/app/front/WorkspaceAgentFront.tsx
@@ -463,6 +463,14 @@ function useStoredNullableStringState(
 const EMPTY_HEADERS: Record<string, string> = {}
 const EMPTY_STRING_LIST: string[] = []
 const PREPARING_WARMUP_STATUS: WorkspaceWarmupStatus = { status: "preparing" }
+// An optimistically-created pane loses its protection once its owning
+// Agent's session snapshot acknowledges it (the normal, fast path). This is
+// the terminal fallback for the abnormal one: the session was deleted (in
+// app or out of band) or a provider returned an id that never materializes,
+// so no snapshot will EVER acknowledge it. Past this window the key is
+// dropped and reconciliation falls back to the real authoritative session
+// instead of protecting a pane that may no longer exist.
+export const OPTIMISTIC_CREATE_ACK_WINDOW_MS = 10_000
 const useIsomorphicLayoutEffect = typeof window === "undefined" ? useEffect : useLayoutEffect
 
 function sessionCreateProtocolError(message: string): SessionCreateProtocolError {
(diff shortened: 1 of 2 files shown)


hachej:
REVISE

Blocking race in the success path: React batches the new pane state and addressed-Agent selection, but `useAddressedFleetSessions` receives each Agent controller update through a child passive effect (`addressedFleetSessions.tsx:121-123,147-157`). The first render addressed to Beta can therefore still carry the previous Beta snapshot (`activeSessionId=beta-one`) while the pane is already `beta-new`. Reconciliation then sees `currentMatchesControlledSession=false` (`WorkspaceAgentFront.tsx:1808-1818`) and may replace the optimistic pane with `beta-one` (`:1835-1842`); the later fleet publish restores `beta-new`. That permits a painted `beta-new → beta-one → beta-new` stomp. `optimisticCreatedPaneKeysRef` currently exempts the key from pruning only; it does not protect it from desired-session replacement.

Please preserve the active optimistic created key until the selected Agent snapshot acknowledges it, and add a sequence-sensitive assertion that rejects an intervening old-session render. The current final `waitFor` proves only eventual state and cannot catch this frame.

Other adversarial checks are clean:
- Old-pane composer targeting stays bound to the owner-qualified pane/session key; the addressed switch does not retarget an admitted in-flight send. Any unsent local draft loss is the existing `replace-all` New-chat behavior, not introduced by this change.
- The Agent-card and picker paths share `createChatPaneTransaction`. Their earlier `onSelectAgent` updates only the future-create picker override, so the success callback performs at most one addressed-Agent selection; no double-select.
- Test honesty verified dynamically: with only the production fix temporarily removed, the new regression test fails with `data-agent-type-id="alpha"` instead of Beta. The exact PR hunk was restored and the worktree is clean.
- Current suites pass: `WorkspaceAgentFront.test.tsx` 100/100; `AppLeftPane.test.tsx` 41/41.

Verdict: REVISE.

hachej:
Addressed the blocking finding, verified the three clean-checks with file:line evidence.

## 1. Blocking race (optimistic pane vs. stale addressed snapshot) — REAL, FIXED
Confirmed exactly as described: `createChatPaneTransaction`'s success callback (`WorkspaceAgentFront.tsx`) flips the addressed Agent and updates `chatPaneState` in the same tick, but `useAddressedFleetSessions`'s `snapshots` map only catches up once the newly addressed Agent's `FleetSessionSource` child re-publishes through its own passive effect (`addressedFleetSessions.tsx:121-123,147-157`) — at least one render later. In that gap `chatSessionKey` still names the Agent's *previous* active session, so the reconciliation effect (`WorkspaceAgentFront.tsx` ~1799-1830) judged the just-created pane "uncontrolled" and replaced it with the stale one.

Two spots needed the fix, not one — `optimisticCreatedPaneKeysRef` was already exempting the new key from **pruning**, but two *separate* gates could still override the desired session: `currentMatchesControlledSession` (the identity check) and the `resolvedSessionsByKey.has(current.activeId)` gate feeding `resolvedDesiredSessionId`. Fixing only the first still failed the repro (session flipped to `beta-one`) because the second gate independently vetoed the optimistic pane. Both now short-circuit on `currentIsOptimisticCreate`.

New test: `"focuses the session created via the New chat Agent picker"` (`WorkspaceAgentFront.test.tsx`). It deliberately gates Beta's own snapshot update behind a manually-released promise, independent of the create() resolution the transaction awaits — reproducing the exact ordering deterministically instead of hoping a microtask-count race lands the same way twice. It asserts the mid-race frame directly (before releasing the gate) AND records every render via an `ObservingChatPanel`, asserting `beta:beta-one` is never observed.

Dynamically falsified twice, restored after each:
- Toggled off `currentIsOptimisticCreate` (identity gate only) → still failed: `data-session-id="beta-one"` at the mid-race assertion. This is what caught the second gate.
- Toggled off `currentIsOptimisticCreate` entirely (both gates) → same failure.
- Restored → green: `WorkspaceAgentFront.test.tsx` 100/100, `AppLeftPane.test.tsx` 41/41 (141/141 combined foreground run).

## 2. Old-pane in-flight send / queued draft — CONFIRMED CLEAN, no change needed
Verified independently. `emitWorkspaceComposerStop` (`WorkspaceAgentFront.tsx:1310`) — the only call site that signals an in-flight send to stop — fires exclusively from `resolvedSwitch`, which my diff never calls; the new code only calls `addressedAgents.selectAgentTypeId`. Chat panes stay keyed by `workspaceSessionKey(sessionId, agentTypeId)` throughout (`chatPanes` construction ~2291, `chatPaneIds`/`activeChatPaneId` ~1878-1882), independent of the addressed selection, so a pane's own identity is untouched by this change.
The one real side effect — the *old* pane unmounting — is `createChatSession`'s pre-existing `mode: "replace-all"` (`WorkspaceAgentFront.tsx:2041-2043`), unconditional and untouched by this diff; it fires identically for a same-Agent "New chat" click. Not introduced here.

## 3. Double-select on the card path — CONFIRMED CLEAN, no change needed
Verified: `addressedAgents.selectAgentTypeId` now has exactly one call site in the whole file (`WorkspaceAgentFront.tsx:2050`, inside `createChatPaneTransaction`'s success callback — shared by every creation entry point). The card's and picker's own `onSelectAgent` calls (`AppLeftPane.tsx:558-561`, `:772-775`) only ever reach `setNewChatAgentTypeIdOverride` — a *different* function that retargets the picker, never the addressed selection. `grep -n "selectAgentTypeId" WorkspaceAgentFront.tsx` confirms: 1 type decl, 1 forwarding to `useAddressedFleetSessions`, 1 call site, 1 dep-array entry.

## 4. Test honesty — reaffirmed on the new commit
Same falsify/restore discipline applied to the new race fix (see #1 above): removed → red with the exact predicted symptom, restored → green, worktree clean.

Pushed as `3976a89b6` on `fix/1470-start-new-chat-focus`.

hachej:
REVISE

Blocking optimistic-create lifecycle hole: the normal resolution path is safe—the reconciliation effect removes an optimistic key once `resolvedSessionsByKey` contains it (`WorkspaceAgentFront.tsx:1768-1770`), before `currentIsOptimisticCreate` is evaluated (`:1821`). A later deletion therefore reconciles normally after that acknowledgment.

But membership in `resolvedSessionsByKey` is the only removal path for `optimisticCreatedPaneKeysRef`. If the created session is deleted externally before its first addressed snapshot contains it (or a provider returns an ID that never materializes), the key can never clear. With another authoritative session present, that stale flag bypasses both the controlled-session and inventory gates (`:1821-1832`), and pruning explicitly retains the absent pane (`:1836-1841`). The nonexistent optimistic pane can then remain active indefinitely and suppress legitimate reconciliation. In-app deletion removes the pane but also leaves the key dormant (`:2102-2114`), and workspace reset does not clear the set.

Please add terminal cleanup for a never-observed optimistic create (and clear on deletion/workspace or source reset as appropriate), with a regression that deletes/invalidates the created session before acknowledgment and proves reconciliation selects the real authoritative session. The new mid-race sequence test is otherwise sound and the originally reported stomp is fixed.

Verified at `3976a89b6`: targeted `WorkspaceAgentFront.test.tsx` + `AppLeftPane.test.tsx` pass, 141/141.

Verdict: REVISE.

hachej:
Addressed the optimistic-create lifecycle hole. Pushed as `58b4d133f`.

## Blocking finding — REAL, FIXED
Confirmed: `optimisticCreatedPaneKeysRef`'s only removal path was `resolvedSessionsByKey.has(key)` (`WorkspaceAgentFront.tsx:1785-1787` pre-fix). A session deleted before its owning Agent's first snapshot ever observed it, or a provider id that never materializes, left the key permanently set — with another authoritative session present, that stale flag bypassed both the identity gate and the inventory gate added in the previous pass, and pruning explicitly retained the absent pane.

Changes:
- `optimisticCreatedPaneKeysRef` is now `Map<string, number>` (key → creation timestamp) instead of a `Set`.
- New `OPTIMISTIC_CREATE_ACK_WINDOW_MS` (10s, exported for the test) is the terminal fallback: the reconciliation cleanup loop now drops a key once it's either acknowledged (`resolvedSessionsByKey.has(key)`, unchanged) OR aged past the window — whichever comes first.
- Since nothing else re-renders on a timer alone, a `setTimeout` scheduled at creation (cleaned up on unmount via a timer-id ref) bumps a small `optimisticCreateAckTick` state once the window elapses, which is in the reconciliation effect's dependency array — forcing it to actually re-evaluate and drop the stale key even with zero other state changes.
- Explicit signal for the case we DO know about: `deleteSessionAndPane` (`:2141` on, `optimisticCreatedPaneKeysRef.current.delete(sessionKey)` at the top) clears the key immediately on in-app deletion — no need to wait out the window when we know for certain it's gone. Deleting a key that was never optimistic is a harmless no-op.
- Workspace/source reset: a new effect keyed on `[chatPaneSourceIdentity, workspaceId]` clears the whole map, matching the existing `chatPaneState` reset on the same transition — a leftover key belonged to a pane that no longer exists once that reset fires.

## Regression test
`"ages out a never-acknowledged optimistic create and falls back to a real session"` (`WorkspaceAgentFront.test.tsx`): Beta's mock `create()` resolves with a real id but its own session list never includes it — reconciliation cannot distinguish this from a genuine external deletion-before-ack; both just mean the key never appears in `resolvedSessionsByKey`, which is exactly the lifecycle hole described. Using `vi.useFakeTimers()`:
1. Picker creates the phantom Beta session — pane shows it, protected (matches the existing stomp-fix behavior).
2. Advance to `ACK_WINDOW - 1s` — still protected, still showing the phantom (proves the window isn't premature).
3. Advance past the window — pane falls back to `alpha-one`, the real authoritative session.

Falsified and restored: disabled the age-out condition (`createdAt <= optimisticCreateDeadline`) → the test failed at step 3, timing out stuck on the phantom pane (exactly the reported symptom — "remain active indefinitely"). Restored → green.

## What I did NOT add a bespoke black-box test for
`deleteSessionAndPane`'s explicit `.delete(sessionKey)` call: that function already synchronously removes the pane from `chatPaneState` before any reconciliation effect runs, so the *visible* behavior on deletion was already correct even before this fix — the ref cleanup there is a hygiene fix against a *future* effect run treating a since-deleted key as still-optimistic (e.g. via the `prunedIds` OR-clause), which I could not find a way to observe black-box without exposing the ref. Verified by code inspection (single-line, matches the same pattern as the ack-window path) and the full targeted suite passing with no regressions; flagging this gap explicitly rather than claiming coverage I don't have.

## Verified
- `WorkspaceAgentFront.test.tsx` + `AppLeftPane.test.tsx`: 142/142 (foreground)
- `tsc --noEmit` on `packages/workspace`: clean
- The originally reported stomp fix and its sequence-sensitive test are untouched and still pass.

hachej:
SHIP

The 10s window is acceptable for this UI. It starts only after createChatPaneTransaction receives a resolved canonical session ID and installs the optimistic pane, so create-request/network latency over 10s cannot age the pane out mid-creation. The built-in usePiSessions path publishes the returned session into local inventory before create() returns; the remaining acknowledgment delay is React/fleet snapshot propagation, for which 10s is ample. Clearing on resolve would remove the protection at exactly the stale-snapshot race this fix addresses; errors create no optimistic key to clear.

Unmount cleanup clears every outstanding timeout and empties the timer set. Source/workspace reset clears the optimistic map; any remaining scheduled tick is harmless. The disclosed lack of a dedicated black-box delete-ref cleanup test is a non-blocking coverage gap.

Verified at 58b4d133f: both targeted files pass, 142/142, with a clean single-worker combined run. Independent fresh-eyes pass found no material issue.

Verdict: SHIP.

hachej:
Fixed the CI typecheck red: the ack-timer handle now uses `ReturnType<typeof setTimeout>` (repo's usual pattern) instead of `window.setTimeout`'s DOM `Timeout` type, with `window.` dropped from both `setTimeout`/`clearTimeout` call sites for consistency. Verified with the actual CI command: `pnpm -C packages/workspace run typecheck` (tsconfig.front.source.json + tsconfig.server.source.json) clean; `WorkspaceAgentFront.test.tsx` + `AppLeftPane.test.tsx` 142/142 foreground. Pushed as `0bdb768ad`.

PR #1473, opened 2026-08-29
#1463 fix(core): default-workspace deletion can't brick the account

### LINKED ISSUE #1463: Deleting the default workspace bricks the account: auto-recreate collides with idx_workspaces_default_per_user_app (soft-deleted row not excluded)
HIGH. Round-3 full-app hunt 2026-08-29. Repro: fresh verified user → Settings → Danger zone → Delete workspace (the auto-created default) → confirm. Soft-delete succeeds (deleted_at set), then the server auto-recreates a default and the insert FAILS: duplicate key value violates unique constraint idx_workspaces_default_per_user_app — the unique index is not scoped to exclude soft-deleted rows. Result: zero active workspaces; every GET /api/v1/workspaces → 500; UI stuck at 'Create your first workspace' → 'Preparing workspace…' with an 'agent scope is not authorized' banner. Account bricked until DB intervention. Insert site: packages/core/src/server/db/stores/PostgresWorkspaceStore.ts:210. Fix: partial unique index WHERE deleted_at IS NULL (migration) AND make delete-default either refuse (if it's the user's only workspace) or recreate transactionally; add a conformance test for delete-default → user still has exactly one active default. Repro user smoke7a-1787992367@example.test / workspace 516c60d6-014b-46e6-a20a-8ca9a57911a9 in fullapp_smoke.

hachej:
Fixes #1463

## Root cause

Deleting a user's default workspace soft-deleted it (`deleted_at` set), then a
later `GET /api/v1/workspaces` auto-recreates a default since the user now has
zero active workspaces. That insert collided with
`idx_workspaces_default_per_user_app` — the unique partial index excluded
non-default rows (`WHERE is_default = true`) but **not soft-deleted ones**, so
the tombstoned default still blocked a fresh insert for the same
`(created_by, app_id)`. Result: the insert failed with a duplicate-key error,
the user was left with zero active workspaces, and every subsequent
`GET /api/v1/workspaces` 500'd — account bricked until DB intervention.

Insert site: `packages/core/src/server/db/stores/PostgresWorkspaceStore.ts` (`create()`).

## Fix — two layers

**Layer 1 (migration, `drizzle/0027_workspace_default_index_excludes_deleted.sql`):**
rebuild `idx_workspaces_default_per_user_app` scoped to
`is_default = true AND deleted_at IS NULL`. Mirrored in `schema.ts` and in
`LocalWorkspaceStore`'s in-memory delete/create comments for consistency.

**Layer 2 (behavior, `packages/core/src/server/routes/workspaces.ts`):**
the DELETE route now recreates a fresh default **synchronously, in the same
operation**, when the delete would leave the acting user with zero active
workspaces — instead of relying on the next `GET /workspaces` to notice and
repair it.

### Option chosen: recreate-transactionally, not refuse

I initially implemented a hard refuse (409 `last_workspace`) when deleting
the user's only workspace, mirroring the existing `LAST_OWNER` pattern. I
reverted that: it broke a legitimate flow already covered by
`e2e/v7-platform.test.ts` ("Bob deletes workspace") — a non-creator owner
(promoted via invite) cleaning up the last workspace in an app. Refusing
based on "is this the only workspace" doesn't distinguish that case from the
literal repro (a user's own auto-created default with no other workspaces),
and scoping the refusal correctly to preserve that flow got complicated fast.

Since the migration alone already makes the existing recreate-on-list path
(`listOrCreateDefaultWorkspace`) safe, recreate-transactionally is strictly
additive: it's the same recreate the client would trigger on its next list
call (the UI already invalidates the workspaces query on delete success), just
done synchronously so the user never even sees a zero-workspace flash. If the
replacement's provisioning fails, the delete response still succeeds (best
effort) — the next `GET /workspaces` recreate-on-list still repairs it.

## Tests

- `packages/core/src/server/db/__tests__/workspaceDefaultIndexExcludesDeleted.migration.test.ts` — migration SQL content test (mirrors the `0026` migration test's style).
- `packages/core/src/server/db/stores/__tests__/PostgresWorkspaceStore.test.ts` — store conformance case: delete the only default workspace, recreate one with the same `(created_by, app_id)`, prove no collision and exactly one active default remains.
- `packages/core/src/server/routes/__tests__/workspaces.test.ts` — route tests (bare mock store and with-provisioner mock) proving the synchronous recreate, and a case proving no recreate happens when the owner still has other active workspaces.
- `e2e/v7-platform.test.ts` and `LocalWorkspaceStore.test.ts` continue to pass unmodified.

## Proof

Ran against a dedicated local Postgres db (created for this task; never
touched other roles/dbs):

```
role_1463 / fullapp_1463  (owned by role_1463, created via sudo -u postgres psql)
```

- `pnpm --filter core typecheck` — clean except one pre-existing, unrelated
  error in `packages/workspace` (`WorkspaceAgentFront.tsx` `setArchived`),
  confirmed present on `origin/main` too.
- Targeted suite (migration test, `PostgresWorkspaceStore.test.ts`,
  `LocalWorkspaceStore.test.ts`, `workspaces.test.ts`,
  `workspaceAuthAudit.test.ts`, `e2e/v7-platform.test.ts`): **262/262 passed**.
- Full `packages/core` suite (backend only): **1087/1092 passed**; the 5
  failures (2 `PostgresModelBudgetStore` budget-double-counting tests, 3
  `createCoreWorkspaceAgentServer.provisioning.test.ts` lease/timeout tests)
  reproduce identically on unmodified `origin/main` against the same db —
  confirmed pre-existing and unrelated to this change.

No merge performed.

### DIFF
--- a/packages/core/drizzle/0027_workspace_default_index_excludes_deleted.sql
+++ b/packages/core/drizzle/0027_workspace_default_index_excludes_deleted.sql
@@ -1,0 +1,11 @@
+-- #1463: idx_workspaces_default_per_user_app only excluded non-default rows,
+-- not soft-deleted ones. Deleting a user's default workspace (deleted_at set,
+-- is_default stays true) left a tombstoned row that still satisfied the old
+-- partial index, so the auto-recreate insert for a fresh default collided
+-- with it (duplicate key) and left the user with zero active workspaces.
+-- Rebuild the index scoped to live rows only.
+DROP INDEX "idx_workspaces_default_per_user_app";
+--> statement-breakpoint
+CREATE UNIQUE INDEX "idx_workspaces_default_per_user_app"
+  ON "workspaces" USING btree ("created_by","app_id")
+  WHERE "workspaces"."is_default" = true AND "workspaces"."deleted_at" IS NULL;
--- a/packages/core/drizzle/meta/_journal.json
+++ b/packages/core/drizzle/meta/_journal.json
@@ -190,6 +190,13 @@
       "when": 1787961600000,
       "tag": "0026_workspace_agent_seats",
       "breakpoints": true
+    },
+    {
+      "idx": 27,
+      "version": "7",
+      "when": 1787990400000,
+      "tag": "0027_workspace_default_index_excludes_deleted",
+      "breakpoints": true
     }
   ]
 }
--- a/packages/core/e2e/v7-platform.test.ts
+++ b/packages/core/e2e/v7-platform.test.ts
@@ -459,4 +459,106 @@ describe('v7 platform E2E', () => {
       expect(res.json()).toEqual({ deleted: true })
     })
   })
+
+  // ===========================================================================
+  // Scenario C — #1463 destroy-succeeded-then-DB-failed recovery contract
+  // ===========================================================================
+  // Thermo-gate finding: the sandbox destroy in DELETE runs BEFORE the DB
+  // transaction and can't be rolled back with it. If the DB step then fails
+  // (transaction rolls back, workspace row survives), the workspace looks
+  // like an ordinary active workspace but its real sandbox is gone. This
+  // proves, end-to-end against real routes (workspace + settings) and a
+  // real LocalWorkspaceStore, that the surviving workspace is honestly
+  // marked recoverable and POST /runtime/retry actually brings it back —
+  // not just that DELETE returns 500.
+  describe('Scenario C: destroy-succeeded-then-DB-failed recovery', () => {
+    let app: FastifyInstance
+    let inject: ReturnType<typeof createInjector>
+    let workspaceStore: LocalWorkspaceStore
+    const provisionFn = vi.fn<WorkspaceProvisioner['provision']>()
+    const destroyFn = vi.fn<WorkspaceProvisioner['destroy']>()
+    let workspaceId: string
+
+    beforeAll(async () => {
+      const userStore = new LocalUserStore()
+      workspaceStore = new LocalWorkspaceStore(userStore)
+      userStore.seed({ id: 'carol', email: 'carol@test.dev', name: 'Carol', emailVerified: true, image: null })
+
+      provisionFn.mockResolvedValue({ volumePath: '/volumes/carol-ws' })
+      destroyFn.mockResolvedValue(undefined)
+
+      const provisioner: WorkspaceProvisioner = { provision: provisionFn, destroy: destroyFn }
+
+      app = Fastify({ logger: false })
+      app.decorate('config', { appId: 'test-app', defaultAgentTypeId: 'default', auth: { mail: null, url: 'http://localhost:3000' }, features: { inviteTtlDays: 7 } } as any)
+      app.decorate('workspaceStore', workspaceStore)
+      app.decorate('provisioner', provisioner)
+      registerErrorHandler(app)
+
+      app.addHook('onRequest', async (request) => {
+        const userId = request.headers['x-test-user'] as string | undefined
+        request.user = userId ? { id: userId, email: `${userId}@test.dev`, name: null } : null
+      })
+
+      await app.register(registerWorkspaceRoutes)
+      await app.register(registerSettingsRoutes)
+      await app.ready()
+
+      inject = createInjector(app)
+    })
+
+    afterAll(async () => {
+      await app.close()
+    })
+
+    it('Carol creates her only workspace', async () => {
+      const res = await inject('POST', '/api/v1/workspaces', 'carol', { name: 'Carol WS' })
+      expect(res.statusCode).toBe(201)
+      workspaceId = res.json().workspace.id
+    })
+
+    it('delete fails after the sandbox is already destroyed → 500, but the surviving workspace is marked recoverable', async () => {
+      // Force a genuine post-destroy DB failure: patch the real
+      // LocalWorkspaceStore's deleteAndRecreateDefaultIfEmpty to throw for
+      // exactly this one call (simulating, e.g., a real Postgres error mid-
+      // transaction), then restore it immediately after.
+      const original = workspaceStore.deleteAndRecreateDefaultIfEmpty.bind(workspaceStore)
+      workspaceStore.deleteAndRecreateDefaultIfEmpty = async () => {
+        throw new Error('simulated replacement insert failure')
+      }
+
+      try {
+        const res = await inject('DELETE', `/api/v1/workspaces/${workspaceId}`, 'carol')
+        expect(res.statusCode).toBe(500)
+      } finally {
+        workspaceStore.deleteAndRecreateDefaultIfEmpty = original
+      }
+
+      // Destroy genuinely ran (irreversible) before the DB step failed.
+      expect(destroyFn).toHaveBeenCalledWith(workspaceId)
+
+      // The workspace row survives — the account is not workspace-less.
+      const ws = await workspaceStore.get(workspaceId)
+      expect(ws).not.toBeNull()
+
+      // It's honestly marked recoverable via the same contract
+      // POST /runtime/retry already understands.
+      const runtime = await workspaceStore.getWorkspaceRuntime(workspaceId)
+      expect(runtime?.state).toBe('error')
+      expect(runtime?.lastErrorOp).toBe('provision')
+      expect(runtime?.lastError).toContain('simulated replacement insert failure')
+    })
+
+    it('POST /runtime/retry actually recovers it — the workspace becomes usable again, not just honestly-broken', async () => {
+      const res = await inject('POST', `/api/v1/workspaces/${workspaceId}/runtime/retry`, 'carol')
+      expect(res.statusCode).toBe(200)
+      expect(res.json().runtime.state).toBe('ready')
+      expect(res.json().runtime.volumePath).toBe('/volumes/carol-ws')
+      expect(provisionFn).toHaveBeenCalledWith(expect.objectContaining({ workspaceId }))
+
+      // Still exactly one workspace for Carol — the failed delete didn't
+      // leave her with zero, and the recovery didn't create a duplicate.
+      expect(await workspaceStore.list('carol', 'test-app')).toHaveLength(1)
+    })
+  })
 })
--- a/packages/core/e2e/v7-platform.test.ts
+++ b/packages/core/e2e/v7-platform.test.ts
@@ -64,10 +64,11 @@ describe('v7 platform E2E', () => {
     let workspaceId: string
     let rawToken: string
     let inviteId: string
+    let workspaceStore: LocalWorkspaceStore
 
     beforeAll(async () => {
       const userStore = new LocalUserStore()
-      const workspaceStore = new LocalWorkspaceStore(userStore)
+      workspaceStore = new LocalWorkspaceStore(userStore)
       const idempotencyStore = createInMemoryIdempotencyStore()
 
       // Seed alice and bob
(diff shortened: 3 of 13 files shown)


hachej:
REVISE

Blocking: the delete + replacement is not transactional, so the implementation does not satisfy the issue/PR claim that deleting the default cannot leave the account with zero active workspaces.

`packages/core/src/server/routes/workspaces.ts:249-270` does `store.delete()` → `store.list()` → `createWorkspaceForUser()` as separate operations. `PostgresWorkspaceStore.delete()` is a standalone update, while `PostgresWorkspaceStore.create()` opens a different transaction; `WorkspaceStore` exposes no compound transactional operation. Worse, the recreate error is swallowed and DELETE returns 200. Any replacement insert failure therefore leaves the exact zero-active-workspace state this PR is meant to eliminate. The new Map-backed test calls this “transactionally” but has no transaction or injected create failure, so it only proves the happy path.

Provisioning has a related partial-failure hole: workspace creation commits before provisioning. If provisioning fails, the catch still returns delete success with an active replacement in runtime `error`. The next GET does not repair that shape because `ensureDefaultWorkspaceProvisioned()` only provisions a missing runtime or `ready` without a volume, so the PR-body claim that GET repairs any replacement failure is not true.

Required revision: put the database delete + conditional default creation/member/seat insertion behind one canonical store transaction (with concurrency-safe zero-workspace/default enforcement), do not report success while a DB recreate failure has committed zero active workspaces, and add a real Postgres failure-path assertion proving rollback. Define and test the provisioning-failure contract separately; filesystem provisioning cannot share the DB transaction, so it needs honest compensation/retry semantics rather than being described as atomic.

Other adversarial checks:

- Migration 0027 itself is safe for normally migrated production data. The old broader unique index already rejected both live+tombstoned defaults and two live defaults for one `(created_by, app_id)`, so narrowing the predicate needs no data cleanup. Drizzle runs the DROP/CREATE and journal write transactionally; journal idx 27 is correctly ordered after seats migration idx 26. Out-of-band loss/corruption of the old index would make CREATE UNIQUE fail, but that is not a state the canonical migrations admit.
- Authorization remains owner-gated. A promoted non-creator owner deleting the last shared workspace creates a new default owned by the acting user; that matches the current public POST policy, but it is a material ownership transition and needs an explicit Bob-style test (including `createdBy`, membership/provisioner owner, and what happens to the original creator). The existing e2e only checks 200 + destroy.
- LocalWorkspaceStore parity is comments only; it does not model the one-live-default uniqueness invariant, so Local/e2e cannot catch the Postgres collision or duplicate-default races.
- No existing assertions were weakened (the test diff is additive), but all new assertions are happy-path and filter by `createdBy` rather than the user-visible membership invariant.

Verification on `65be87dccc84c36f8ee4bfcb77e92fc54ffa728c`: targeted migration, Postgres store, Local store, route, auth-audit, and v7 e2e suites passed (257/257 total). That green proof does not exercise the blocking partial-failure path.

hachej:
Addressed in 2e459d3ea. Per-finding disposition:

**Blocking — delete+recreate not transactional (real defect, fixed).** Replaced the separate delete()/list()/create() sequence with `WorkspaceStore.deleteAndRecreateDefaultIfEmpty(id, actingUserId, recreate)`: soft-delete, the "is the acting user now workspace-less" lock+count, and the replacement workspace+member+seat inserts now run in **one** Postgres transaction. The route no longer swallows the recreate error — a DB failure propagates as an honest 500 instead of a false `{ deleted: true }`. Proof: `PostgresWorkspaceStore.test.ts` has a new test that forces a *genuine* DB-level failure (a bogus `created_by` that violates the FK to `users`, not a mock) during the replacement insert and asserts the original workspace is exactly as it was afterward — not deleted, no orphan replacement.

While building the lock, I found and fixed a real deadlock in my own first draft: locking the acting user's active-workspace set *after* soft-deleting the target let two concurrent deletes of two different workspaces owned by the same user each hold their own target's row lock while blocked wanting the other's — a circular wait. Fixed by locking the full set in a stable `ORDER BY w.id`, *before* any mutation. Proof: a new concurrency test runs two `deleteAndRecreateDefaultIfEmpty` calls against real concurrent Postgres transactions via `Promise.all` and asserts no deadlock error and exactly one replacement created (never zero, never two).

**Provisioning partial-failure hole (real gap, fixed).** The route no longer swallows a provisioning failure on the replacement either — it throws `PROVISION_FAILED` (500), same contract as a normal `POST /api/v1/workspaces` failure: the DB row already exists and committed (so the account isn't workspace-less), runtime is left in `error` state, retryable via `POST /runtime/retry`. Proof: new route test asserts 500 + the replacement row's runtime state/lastError/lastErrorOp, exactly mirroring the existing create-failure contract.

**Migration safety on pre-existing data shapes — rebutted (already correct), reinforced with proof.** Your own analysis is right: the old broader index already rejected two live defaults and a live+tombstoned pair, so narrowing to also require `deleted_at IS NULL` needs no data cleanup, and journal idx 27 is correctly ordered after seats migration idx 26 (unchanged). Added a store test that proves the narrowing didn't accidentally loosen the still-live-rows invariant: two live defaults for the same `(created_by, app_id)` still collide with a `23505 unique_violation` on `idx_workspaces_default_per_user_app`.

**Non-creator owner recreate (real gap, fixed).** Added the explicit Bob-style test you asked for, at both layers: a mock-store route test (`workspaces.test.ts`) and a real-`LocalWorkspaceStore` e2e test extending `v7-platform.test.ts`'s "Bob deletes workspace". Both assert: the replacement's `createdBy` is the deleting/acting user (Bob), not the original creator (Alice); Bob gets `owner` membership on the replacement; the provisioner is called with Bob as `ownerId`; and Alice is left with zero workspaces in the app and is *not* a member of Bob's replacement (no silent carry-along).

**LocalWorkspaceStore parity (real gap, fixed).** `create()` now mirrors `idx_workspaces_default_per_user_app` — rejects a second live default for the same `(createdBy, appId)` — and `LocalWorkspaceStore` has its own `deleteAndRecreateDefaultIfEmpty` with the same all-or-nothing contract (rolls back the soft-delete in-memory if the replacement `create()` throws). Local/e2e-backed tests can now catch the same collisions Postgres does; proven with a dedicated forced-collision rollback test in `LocalWorkspaceStore.test.ts`.

**Weakened assertions — not actionable, confirmed by your own note ("the test diff is additive").** Took the adjacent critique seriously though: new/existing assertions in `workspaces.test.ts` now check the user-visible `list()`/membership invariant via an `activeForUser()` helper, not an internal `createdBy`-only filter.

**Proof:** dedicated Postgres db `fullapp_1463` (role `role_1463`, never touched other roles/dbs). Targeted suite (migration content, `PostgresWorkspaceStore` incl. the new rollback/concurrency/migration-safety tests, `LocalWorkspaceStore` incl. the new parity tests, workspaces route, `workspaceAuthAudit`, members, invites, `e2e/v7-platform.test.ts`): **340/340 passed**. Full core backend suite: **1117/1120 passed** — the 3 failures (2 `PostgresModelBudgetStore`, 1 `deleteUserCompletely`) reproduce identically on unmodified `origin/main` against the same db, confirmed pre-existing/unrelated. `tsc --noEmit` clean except the same pre-existing, unrelated error in `packages/workspace` (`WorkspaceAgentFront.tsx` `setArchived`).

Commit: `2e459d3ea`. No merge performed.

hachej:
REVISE

Blocking: the database transaction is now sound, but the route still performs the irreversible runtime destroy before entering it.

`packages/core/src/server/routes/workspaces.ts:229-247` awaits `provisioner.destroy(id)`; only afterward does `:257-260` call `deleteAndRecreateDefaultIfEmpty()`. If that store call fails after destroy succeeds (for example, a replacement/member/seat insert error), Postgres correctly rolls back and the global handler correctly returns 500 `internal_error` — but the original workspace row is restored after its filesystem/sandbox has already been destroyed. Its runtime metadata is not moved to an error/recovery state, and `POST /runtime/retry` only accepts `state=error,lastErrorOp=provision` (`packages/core/src/server/routes/settings.ts:65-81`). The result is a live-looking, potentially unusable workspace, contradicting the route claim that it is “left exactly as it was.”

Required revision: define and implement a recoverable cross-boundary contract for “destroy succeeded, DB delete/recreate failed,” and add a route test that forces that sequence and proves the surviving workspace is honestly marked/recoverable (or otherwise remains usable). Do not merely assert the 500.

Clean checks: the ordered pre-mutation workspace lock is deadlock-free for the requested same-user concurrent-delete case; workspace + owner member + #1442 Agent seat are inside one Drizzle transaction; DB failures map to 500; replacement provisioning failures map to `provision_failed` and retain the existing retry contract. Focused real-Postgres store tests passed 84/84 against `fullapp_1463`; route tests passed 49/49.

hachej:
Addressed in b4b1bd8ce.

**Blocking — destroy runs before the DB transaction, leaving a live-looking-but-broken workspace on post-destroy DB failure (real defect, fixed).** You're right: rollback restored the DB row, but its real sandbox was already irreversibly torn down by `provisioner.destroy(id)` above, and the runtime metadata was never updated to reflect that — `POST /runtime/retry` only accepts `state=error, lastErrorOp=provision`, and I hadn't set either.

Fix: `workspaces.ts` now catches a DB failure from `deleteAndRecreateDefaultIfEmpty` that occurs after a successful `destroy`, and marks the surviving workspace `state=error, lastErrorOp=provision`. I deliberately did not invent a new cross-boundary status — the actual remediation needed ("this workspace needs a fresh sandbox provisioned") is identical to an ordinary provisioning failure, so it reuses the already-tested `provision` retry contract rather than adding a new one. The route still rethrows the original DB error (still an honest 500); the runtime mark is best-effort compensation, logged separately if it itself fails, and never masks the real error.

Per "do not merely assert the 500," two tests:
- `workspaces.test.ts`: monkey-patches the mock store's `deleteAndRecreateDefaultIfEmpty` to throw for exactly one call, asserts `destroy` was called, the row survives undeleted, and the runtime carries `state=error/lastErrorOp=provision` with the real error message.
- `e2e/v7-platform.test.ts` (new Scenario C): registered against real workspace **and settings** routes with a real `LocalWorkspaceStore` — forces the same sequence, then proves recovery isn't just "honestly marked" but actually *works*: calls `POST /runtime/retry` and asserts the runtime reaches `state=ready` with a fresh `volumePath`, and the user still has exactly one workspace (no duplicate from the retry, no zero from the original failure).

**Clean checks acknowledged, no changes:** ordered pre-mutation lock deadlock-free for same-user concurrent deletes; workspace+member+seat in one Drizzle transaction; DB failures → 500; replacement-provisioning failures → `provision_failed` with the retry contract intact.

**Proof:** dedicated Postgres db `fullapp_1463` (role `role_1463`). Targeted suite: **368/368 passed** (PostgresWorkspaceStore 84/84, workspaces route 50/50, e2e v7-platform 17/17 including the new recovery scenario, plus migration/Local/authAudit/members/invites/settings/idempotency). Full core backend suite: **1121/1124 passed** — the 3 failures (2 `PostgresModelBudgetStore`, 1 `deleteUserCompletely`) reproduce identically on unmodified `origin/main` against the same db, confirmed pre-existing/unrelated. `tsc --noEmit` clean except the same pre-existing unrelated error in `packages/workspace`.

Commit: `b4b1bd8ce`. No merge performed.

hachej:
SHIP

Verified b4b1bd8ce. If recovery-state marking fails, workspace.delete.recovery_mark.failed is logged with workspaceId and markErr; the original post-destroy DB error is then separately logged and rethrown, preserving the honest 500. POST /runtime/retry only reprovisions the surviving workspace ID and updates its runtime; it never creates a workspace. Scenario C exercises the real routes with LocalWorkspaceStore and proves recovery reaches ready while the user still has exactly one workspace.

Local proof: workspaces routes 50/50, LocalWorkspaceStore 79/79, v7 Scenario suite 17/17, and PostgresWorkspaceStore 84/84 against fullapp_1463 (230/230 total). Worktree clean; diff check clean.

PR #1474, opened 2026-08-29
#1465 fix(core): auth card reachable on mobile viewports

### LINKED ISSUE #1465: Mobile (390x844): unauthenticated auth card renders with zero size — cannot sign in or sign up on mobile
HIGH. Round-3 hunt: load / unauthenticated at 390x844 → the auth card DOM (Sign in/Sign up, email, password, Continue) exists but every element reports visible:false / null bounding box; the floating bottom-left card that works on desktop doesn't render on mobile, and there's no alternative entry point (no hamburger/Sign-in affordance). Unauthenticated mobile users cannot sign in or sign up at all. Authenticated mobile chrome is fine. Fix: responsive layout for the landing/auth surface (stack the card below the hero or make it a full-width sheet), 44px targets, Playwright mobile spec asserting the sign-in form is visible + submittable at 390x844. Screenshots 7x-68, 7x-70 in the session scratchpad.

hachej:
Fixes #1465

## Root cause

`packages/core/src/app/front/chatFirst/chatFirstPublicShell.css` dropped the sign-in dock outright at narrow widths:

```css
@media (max-width: 1100px) {
  .public-arrow,
  .public-chat-first-shell > aside { display: none; }
}
```

So at 390x844 the Sign in / Sign up card was in the DOM with a null bounding box — exactly what the issue reports.

There was no fallback affordance either. full-app runs `workspaceLayout="plugin-tabs"`, where `WorkspaceAgentFront` routes `topBarRight` (the public "Sign in" button) into the **app-left pane's** `bottomSlot` — and the public shell sets `defaultAppLeftPaneCollapsed: true`. On a phone the pane is collapsed, so the only other sign-in entry point was invisible too. Unauthenticated mobile users could neither sign in nor sign up.

## Design

The dock becomes **responsive rather than absent**.

- **> 1100px — unchanged.** Floating bottom-left gutter card, pixel-identical geometry.
- **<= 1100px — bottom sheet.** Full-width, docked to the viewport edge, rounded top corners, upward shadow, `env(safe-area-inset-bottom)` padding so it clears the home indicator. The form stays a 420px centred column, so a tablet doesn't stretch inputs edge to edge.
- **44px touch targets** on the segmented Sign in / Sign up tabs, the name/email/password inputs, and the submit button (desktop keeps the compact 36-42px gutter card).
- **Keyboard/short-viewport safe.** The shell reserves the sheet's band with `height: 100dvh; padding-bottom: min(268px, 40dvh)`, so the hero and composer lay out *above* the sheet instead of behind it, and the reserve shrinks on short/landscape viewports instead of squeezing the composer out of existence. `dvh` follows the visual viewport when mobile browser chrome collapses.

Structural cleanups that come with it: the fragile `> aside` selector is replaced by an explicit `.public-auth-dock` / `.public-auth-card` contract with an `aria-label="Sign in"` complementary landmark, and `AuthModal`'s dangling `aria-labelledby="auth-modal-title"` (pointing at no element) becomes a real `aria-label`.

## Proof (full-app dev, live)

Measured on the running app, unauthenticated `/`:

| viewport | sign-in dock | composer rail | control heights |
| --- | --- | --- | --- |
| 390x844 (iPhone 15) | `(0, 579) 390x265` — docked to the bottom edge | `(12, 363) 366x56`, fully visible | tabs/inputs/submit all 44px |
| 834x1112 (iPad) | `(0, 847) 834x265`, 420px centred form column | `(77, 307) 680x56` | 44px |
| 1440x900 (desktop) | `(24, 630) 300x246` — **unchanged** | `(380, 311) 680x56` | 36/42/40px, unchanged |

Before the fix the 390x844 dock reported `display: none` / null box. Typing proof at 390x844: email and password fields focus and accept input (`mobile@example.com`, password field focused after fill).

Screenshots captured at 390x844 and 1440x900 during the live run (hero + composer + docked sheet all visible on the phone; desktop gutter card untouched).

## Tests

- New `apps/full-app/e2e/mobile-auth-card.spec.ts` (added to the config's `testMatch`): at 390x844 unauthenticated, asserts the dock has a non-zero box inside the viewport, both tabs are visible with >=44px targets, email/password focus and accept typed values, the submit button is enabled and 44px, and sign-up mode reveals the Name field and Create account button.
- `packages/core/src/app/front/__tests__/CoreWorkspaceAgentFront.test.tsx` pins the `.public-auth-dock` / `.public-auth-card` / `aria-label="Sign in"` contract the CSS hangs off, so a rename can't silently break the phone layout again. Suite: 30/30 passing.

### DIFF
--- a/apps/full-app/e2e/mobile-auth-card.spec.ts
+++ b/apps/full-app/e2e/mobile-auth-card.spec.ts
@@ -1,22 +1,27 @@
-import { expect, test } from '@playwright/test'
+import { expect, test, type Locator, type Page } from '@playwright/test'
 
 /**
  * #1465 — at 390x844 the public (no-auth) landing rendered its sign-in card
  * with a zero bounding box, and the workspace top bar's Sign in button lives in
  * the collapsed app-left pane in the plugin-tabs layout, so a phone had no way
- * to sign in or sign up at all. The dock must stay reachable at phone widths.
+ * to sign in or sign up at all.
+ *
+ * The dock is now a bottom sheet whose height IS the band the shell reserves,
+ * so it can never sit on top of the hero/composer. These tests pin that
+ * invariant at the viewport shapes that break naive bottom-sheet layouts:
+ * a tall phone, a short landscape phone, and a keyboard-reduced viewport.
  */
 
-const IPHONE_15 = { width: 390, height: 844 }
+const PHONE_PORTRAIT = { width: 390, height: 844 }
+const PHONE_LANDSCAPE = { width: 667, height: 375 }
 const MIN_TOUCH_TARGET = 44
+const KEYBOARD_INSET_PX = 336
 
 function json(body: unknown, status = 200) {
   return { status, contentType: 'application/json', body: JSON.stringify(body) }
 }
 
-test.use({ viewport: IPHONE_15 })
-
-test('mobile 390x844: public landing exposes a usable sign-in/sign-up form', async ({ page, baseURL }) => {
+async function mockPublicLanding(page: Page, baseURL: string | undefined) {
   await page.route('**/*', async (route) => {
     const request = route.request()
     const url = new URL(request.url())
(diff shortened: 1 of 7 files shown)


hachej:
Verification (this worktree, live full-app **production** build on 127.0.0.1:5493, Postgres `fullapp_1465`):

- `apps/full-app/e2e/mobile-auth-card.spec.ts` — **PASS (1) FAIL (0)** at 390x844 against the real built app: non-zero in-viewport dock box, 44px tabs/inputs/submit, email + password focus and accept typed input, sign-up mode reveals Name + Create account.
- `packages/core` front tests — **291 passed / 39 files**, including the new `.public-auth-dock` contract assertion in `CoreWorkspaceAgentFront.test.tsx`.
- `tsc --noEmit` — clean for `@hachej/boring-core` and `full-app`.
- Live geometry, unauthenticated `/`: 390x844 dock `(0, 579) 390x265` with the composer fully clear at `(12, 363)`; 834x1112 dock `(0, 847) 834x265` with a 420px centred form column; 1440x900 dock `(24, 630) 300x246` — identical to the pre-fix desktop geometry.

Note: the config's `webServer` (`e2e/google-auth-webserver.sh`) re-runs a full `build:deps` including DTS on every launch, which exceeds the 360s `webServer.timeout` on a cold VM. Unrelated to this change, but worth a follow-up if CI ever runs cold.

hachej:
REVISE

**P1 — the short-landscape layout traps the composer behind the fixed auth sheet.** In the current production build at `667x375`, the sheet is `265px` tall at `y=110..375`, but the `min(268px, 40dvh)` reserve is only `150px`. The composer textarea is at `y=242..282`; hit-testing its center returns the auth password input/form, not the composer. The document, body, and public shell all have `scrollHeight === clientHeight === 375`, with body overflow hidden, so there is no scroll escape. Sign-up expands the sheet to `293px` at `y=82..375` and worsens the overlap. This directly fails the short-landscape acceptance question and makes the keyboard claim unsafe: a keyboard-reduced visual viewport converges on the same geometry, while the fixed sheet has no bounded internal scrolling.

Please revise the short-height layout so the composer/primary CTA is genuinely unobscured or intentionally reachable by scrolling in both sign-in and sign-up modes. Add durable `667x375` proof that checks non-overlap/hit-testing or actual scroll reachability; also prove the intended keyboard-open policy on a mobile/WebKit/real-device visual viewport.

**The new Playwright spec catches the original `display:none` regression, but overclaims its proof.** `mobile-auth-card.spec.ts:52-59` says the sheet is inside the viewport but asserts only `x`/right bounds, never `y`/bottom. It does not measure the input heights despite the 44px-input claim, and the sign-up "stays inside the sheet" comment checks only visibility. It has no composer/hero overlap, landscape, tablet, scrollability, or visual-viewport/OSK assertion; `fill()` at a fixed `390x844` desktop-Chromium viewport does not test keyboard-open behavior. The reproduced `667x375` blocker therefore passes the test's current contract.

Non-blockers checked: portrait tablet CTA clearance is good (`834x1112`: composer bottom about `355`, sheet top `847`; `1024x768`: composer bottom about `332`, sheet top `503`); all visual changes are scoped to `max-width: 1100px`, so `>1100px` desktop positioning/utilities remain unchanged; the AuthModal `aria-label` replaces a dangling `aria-labelledby` with no in-repo consumer of `auth-modal-title`.

Validation: `pnpm --filter @hachej/boring-core exec vitest run src/front src/app/front --no-file-parallelism` — **39 files / 291 tests passed**.

reviewer: Codex thermo gate
target: `0056403196760995bf012e92b351bc08f757da39`
verdict: revise

hachej:
## Revision for the thermo gate — `03212361e`

All measurements below are from the **production build** of this branch (`node dist/server/main.js`, Postgres `fullapp_1465`), unauthenticated `/`.

### P1 — short-landscape traps the composer behind the sheet → **real defect, fixed**

Confirmed the diagnosis. The reserve and the sheet were two independent numbers: the shell reserved `min(268px, 40dvh)` while the sheet sized itself from its content (265px sign-in / 293px sign-up). At `667x375` the reserve was 150px and the sheet 265px, so 115px of sheet sat over the hero column's visible window — with the composer inside it.

They are now **one number**. `--public-auth-dock-height = min(320px, (100dvh - var(--keyboard-inset)) * 0.45)` is both the shell's bottom reserve and the sheet's `max-height`, so `reserve >= sheet height` holds by construction — sign-in and sign-up, every viewport height, keyboard open or closed. When the form outgrows the band it scrolls **inside** the sheet (`overflow-y: auto; overscroll-behavior: contain`) instead of growing over the composer.

| case | sheet | shell reserve | composer |
| --- | --- | --- | --- |
| 390x844 sign-in | `(0,579) 390x265` | 320px | bottom 419 — 160px clear |
| 390x844 sign-up | `(0,551) 390x293` | 320px | bottom 419 — 132px clear |
| 667x375 sign-in | `(0,206) 667x169` | 168.75px | column window ends at 206 |
| 667x375 sign-up | `(0,206) 667x169` | 168.75px | card scrolls 292→168 |
| 390x844, `--keyboard-inset: 336px` | `(0,279) 390x229`, bottom **508 = 844−336** | 564.6px | above the keyboard |
| 1440x900 | `(24,630) 300x246` | **0px** | unchanged |

### Keyboard policy → **now explicit, using the repo's existing contract**

The sheet is docked at `bottom: var(--keyboard-inset)` and the reserve adds the same inset. That variable is `useKeyboardInset()`'s published contract (`packages/workspace/src/front/layout/useKeyboardInset.ts`), computed from `visualViewport` — the only signal that works on iOS, where the layout viewport never shrinks and `dvh` alone is blind to the keyboard. `CommandPalette` already consumes it the same way. `AuthModal` gets the matching cap so the centred modal also stays inside the keyboard-reduced viewport. Proven in the spec by driving the real variable; headless Chromium has no software keyboard, and the test says so rather than pretending `fill()` at a fixed viewport is keyboard coverage.

### Test honesty → **rewritten, every criticism accepted**

`mobile-auth-card.spec.ts` is now 4 tests. Fixes for each point raised: full in-viewport bounds including **top and bottom** (not just x/right); `elementFromPoint` hit-testing at each element's centre; **measured** heights for every button and input rather than a 44px claim; sign-up fields asserted to sit **inside the sheet's own box**, not merely visible. New coverage: 390x844 sign-up, 667x375 landscape in both modes, and the keyboard-open viewport. The reproduced `667x375` blocker fails the new contract on the parent commit and passes here.

### One correction to the finding's wording

> there is no scroll escape

The document/body/shell are indeed unscrollable, but the hero column is its own `overflow-y: auto` scroller — that is where the escape lives. More importantly, at `667x375` it scrolls **with the sheet removed entirely**: measured `clientHeight 375, scrollHeight 451`. Scrolling on a 375px-tall viewport is inherent to the viewport, not caused by the sheet. So the guarantee I implemented and test is the one that was actually violated: the column's *visible* window ends where the sheet begins, so scrolling the composer into view leaves it unobscured and hit-testable. That was false on the parent commit and is asserted here.

### Non-blockers

Agreed and unchanged: tablet CTA clearance, desktop pixel-identity (all rules stay inside `max-width: 1100px`; desktop reserve measures `0px`), and the `auth-modal-title` `aria-labelledby` having no in-repo consumer.

### Validation

- `mobile-auth-card.spec.ts` — **PASS (4) FAIL (0)** against the production build.
- `pnpm --filter @hachej/boring-core exec vitest run src/front src/app/front --no-file-parallelism` — **39 files / 291 tests passed**.
- `tsc --noEmit` clean for `@hachej/boring-core` and `full-app`.

hachej:
SHIP

Rechecked 03212361e against the prior P1 and the revision reply.

- The short-landscape invariant now has one owner: `--public-auth-dock-height` caps the auth card and is the shell reserve. The dock is offset by `--keyboard-inset` and the reserve adds the same inset, so reserve ≥ sheet height remains true with the keyboard open.
- Sign-up uses the same capped card despite its taller content. At 667x375 the card cannot grow beyond the reserved band; `overflow-y: auto` keeps the taller form scrolling inside the sheet rather than over the composer.
- Every new `--keyboard-inset` consumer has a `0px` fallback. The workspace globals also declare `0px`, and `useKeyboardInset` resets to `0px` on unmount, so hosts or surfaces without the hook retain valid geometry.
- The rewritten spec covers portrait sign-in/sign-up, 667x375 in both modes, keyboard-open positioning, viewport bounds, hit-testing, and measured control sizes. The corrected hero-scroller explanation matches the implementation.
- Desktop rules remain outside the ≤1100px sheet branch except the bounded auth modal rule; the supplied measured desktop geometry is unchanged.

Local validation: `pnpm --filter @hachej/boring-core exec vitest run src/front src/app/front --no-file-parallelism` — 39 files, 291/291 tests passed.

reviewer: Codex final thermo gate
target: `03212361e87ad1b691a6befcc36d4ec6dc73c7bb`
verdict: ship

PR #1319, opened 2026-08-17
[wt-391-forward-9jxj] Composer shows the session's actual model

hachej:
## Summary
- Add addressed `currentModel` authority to session state, sourced from live Pi state or the latest persisted `model_change`/assistant model.
- Existing sessions ignore conflicting browser-local model history; local storage remains only a genuinely empty new-session default.
- Keep the current session model visible while labeling a differing viewer choice as a next-message override.
- Deduplicate and durably replay real addressed model transitions to connected viewers; preserve session authorization/isolation and leave provider/credential policy unchanged.

## Proof
- `node …/vitest.mjs run --exclude 'src/eval/**' --testTimeout 15000` ✅ **219 files passed; 2028 tests passed; 12 skipped; 0 failed; type errors: none** at `4dafa07e1dbe0fc9fc1edf271479826f6cc1bb1b`.
- `node …/typescript/bin/tsc -p packages/agent/tsconfig.json --noEmit` ✅ **0 errors**.
- Focused final `PiChatPanel` regression run ✅ **72/72** (including queued and optimistic non-new session isolation).
- GitHub Actions run `32022709726` ✅ **success**: E2E, changed unit tests, typecheck, lint, invariants, and reference/remote-worker smoke green at the exact SHA.
- Surface policy: two local browser attempts maximum were used while diagnosing CI; no origin cycling. The final authoritative E2E proof is the green GitHub run above.

## Review provenance
- Fresh-eyes — `openai-codex/gpt-5.6-sol`, adversarial acceptance/isolation review, SHA `a7d3adf020097ea96470d34dc66aa05a3eda9d87`: **revise**; connected viewers stale and persisted `model_change` omitted. Fixed.
- Thermo — `openai-codex/gpt-5.6-sol`, maintainability review, SHA `97ff572c2958dcedf5727e30b5491239aaf061d9`: **revise**; impossible event seam, duplicate events, implicit state machine, weak typing. Fixed.
- Fresh-eyes + thermo second passes — same required model, SHA `de6998acd6c5ff73025646b4a09b60ac39adebba`: **clean**.
- Final harness review — same model, SHA `e41709ff9878552bfcb0b90127151500a81cc896`: **clean**.
- New-session edge review — same model, SHA `2d7e0224df0d129e7382d9b8e99881202888640c`: **revise**; queued/optimistic emptiness coverage incomplete. Fixed.
[...]

PR #1475, opened 2026-08-29
Fix composer permanently disabled for sessions without a stored model

### LINKED ISSUE #1469: Composer permanently disabled for sessions not created via the primary new-chat flow (sessionAuthorityReady never satisfied without a model)
Playground round-3 hunt 2026-08-29 (origin/main 811911ec2, BORING_AGENT_E2E_SCRIPTED_PI=1, BORING_AGENT_MODE=direct). Pre-seeded sessions ('Scripted baseline' rows) and sessions created via direct POST /api/v1/agents/:id/sessions end up connected + idle but with the textarea permanently disabled. Root cause traced: useChatModelSelection never yields a currentModel for such sessions, and PiChatPanel.tsx's sessionAuthorityReady requires (sessionModel !== undefined || sessionIsNew) — never true for a resumed/pre-seeded session without a model, so the composer never enables. Reproduced 3x independently. It also stalls the committed e2e spec apps/workspace-playground/e2e/multi-agent-addressed-ui.spec.ts when run directly against main (hangs 3+ min) — check why CI doesn't catch it. Fix: resolve a model for existing sessions (fall back to the agent's default/current model when the session carries none) so authority readiness doesn't depend on a field only new sessions set; test with a pre-seeded session. Session-model resolution also matters for #1319's 'composer shows the session's actual model' behavior — keep that intact.

hachej:
## Summary
- Fixes #1469: sessions not created via the primary new-chat flow (pre-seeded scripted sessions, sessions created via a direct `POST /api/v1/agents/:id/sessions` call) hydrate with `sessionModel === undefined` and `sessionIsNew === false`. `useChatModelSelection`'s `sessionAuthorityReady` never became true for that combination, so `model` stayed `null` forever, `serverModelSelectionUnavailable`/`serverModelSelectionReady` in `PiChatPanel.tsx` never cleared, `policy` stayed `undefined`, and the composer textarea was permanently disabled.
- `sessionAuthorityReady` now also becomes ready once discovery has genuinely run (`enabled && currentDiscoveryLoaded`) for a hydrated, non-new session with no stored model — not just for sessions with a stored model or brand-new sessions.
- The discovery-success handler's existing-session branch no longer blanks `localDefault` to `null` when the session has no model; it now resolves the server's `defaultModel` (or first available model) the same way a new session does — this is what actually gives such a session a model.
- `useChatModelSelection`'s "does not fall back to local storage for a non-new session whose model authority is unavailable" test still passes unmodified: with `enabled: false` (discovery never runs), authority correctly stays unready, so a stale local-storage default is never trusted for an existing session — only a server-vetted default is now trusted.
- PR #1319's behavior (composer shows the session's actual model when one is stored) is untouched: `sessionModel !== undefined` still short-circuits `sessionAuthorityReady` immediately.

## Root cause confirmation
- `packages/agent/src/front/hooks/useChatModelSelection.ts:183` (old): `sessionAuthorityReady = !sessionId || (sessionHydrated && (sessionModel !== undefined || sessionIsNew))` — permanently `false` for a hydrated, non-new session with no stored model.
- `packages/agent/src/front/chat/PiChatPanel.tsx:788`: `policy` is `undefined` unless `serverModelSelectionReady`, which requires `selectedModelAuthorizedByDiscovery`, which requires a non-null `selectedModel` — so `model` staying `null` forever cascades into `policy === undefined` forever, and `disabled = !policy || ...` is permanently `true`.

## Test proof
- New unit test `resolves the agent default model for a pre-seeded/directly-created session with no stored model (#1469)` in `packages/agent/src/front/hooks/__tests__/useChatModelSelection.test.tsx`, verified failing against pre-fix code (`expected null to deeply equal { provider: 'anthropic', ... }`) and passing after the fix.
- `packages/agent/src/front/hooks/__tests__/useChatModelSelection.test.tsx`: 13/13 pass.
- Full agent front suite: 64 files / 662 tests pass, 0 type errors (`pnpm exec vitest run src/front`).
- `pnpm run typecheck` (packages/agent): clean.

## e2e finding
- Booted `apps/workspace-playground` with `BORING_AGENT_E2E_SCRIPTED_PI=1 BORING_AGENT_MODE=direct ANTHROPIC_API_KEY=scripted-e2e-not-a-real-key` and ran `apps/workspace-playground/e2e/multi-agent-addressed-ui.spec.ts` directly. With the fix applied, the composer resolves a model and both the pre-seeded "alpha" session (created via direct `POST .../sessions`) and the "beta" session become interactive — prompts submit and `PI_NATIVE_ASSISTANT_DONE` renders for both, confirming the composer-disabled bug is fixed.
- The spec still times out (180s) at a *different*, pre-existing step: `page.getByRole('menuitem', { name: 'Open in new chat pane' })` never resolves, because `AppSessionActionsMenu.tsx`'s "Chat actions for …" menu has no such item (only Pin/Copy/Rename/Archive/Delete). That menu item appears to have been removed by the #1169 left-pane redesign (`eac1f8e02`, 2026-08-08) without updating this spec — unrelated to #1469, out of scope here.

## CI-coverage finding
- CI's `e2e` job runs `pnpm e2e` → `pnpm --filter @hachej/boring-agent run test:e2e` → `playwright test -c e2e/playwright.config.ts`, whose config is `testDir: '.'` / `testMatch: '*.spec.ts'` scoped to `packages/agent/e2e/`.
- `apps/workspace-playground/e2e/multi-agent-addressed-ui.spec.ts` is only matched by `apps/workspace-playground/playwright.config.ts` (`testMatch: apps/workspace-playground/e2e/**/*.spec.ts`), which is never invoked by any script CI actually runs (`apps/workspace-playground/package.json`'s own `test:e2e` isn't called from the root `e2e` script or any workflow step). The `changes` job's path filter watches `apps/workspace-playground/**` to *gate whether the e2e job runs at all*, but the job body still only executes `boring-agent`'s narrower suite — so this spec (and its stale "Open in new chat pane" assertion) has never actually run in CI, which is why the hang was never caught.

## Test plan
- [x] `pnpm exec vitest run src/front/hooks/__tests__/useChatModelSelection.test.tsx` (packages/agent) — repro test fails pre-fix, passes post-fix
- [x] `pnpm exec vitest run src/front` (packages/agent) — 662/662 pass
- [x] `pnpm run typecheck` (packages/agent) — clean
- [x] `apps/workspace-playground/e2e/multi-agent-addressed-ui.spec.ts` run directly — composer-disabled symptom confirmed fixed; times out later on an unrelated pre-existing stale-selector defect (documented above, not fixed by this PR)

### DIFF
--- a/packages/agent/src/front/chat/__tests__/PiChatPanel.test.tsx
+++ b/packages/agent/src/front/chat/__tests__/PiChatPanel.test.tsx
@@ -1411,6 +1411,61 @@ describe('PiChatPanel sandbox shell', () => {
     })))
   })
 
+  test('unlocks the composer and submits with the server default model for a pre-seeded session with no stored model (#1469)', async () => {
+    // A session created outside the primary new-chat flow (pre-seeded scripted
+    // session, or one created via a direct POST .../sessions call) hydrates
+    // with no currentModel and is not "new" (remoteState()'s default already
+    // carries a committed message). Live server-side model discovery (not the
+    // availableModels/serverResourcesEnabled=false shortcut most tests use)
+    // must still resolve a model so the composer isn't disabled forever, and
+    // the resolved model must be the server's own default — never a leftover
+    // browser-storage choice.
+    const staleBrowserModel = { provider: 'anthropic', id: 'claude-haiku' } as const
+    const serverDefault = { provider: 'anthropic', id: 'claude-opus' } as const
+    const persisted = storage({
+      [scopedComposerStorageKey('workspace-a', 'model')]: JSON.stringify(staleBrowserModel),
+      [scopedComposerStorageKey('workspace-a', 'model:user-selected')]: '1',
+    })
+    const remote = new FakeRemotePiSession(remoteState())
+    const fetchMock = vi.fn(async (input: RequestInfo | URL) => {
+      const url = typeof input === 'string'
+        ? input
+        : input instanceof URL
+          ? input.toString()
+          : (input as Request).url
+      if (url.includes('/models')) {
+        return new Response(JSON.stringify({
+          models: [
+            { ...staleBrowserModel, label: 'Haiku', available: true },
+            { ...serverDefault, label: 'Opus', available: true },
+          ],
+          defaultModel: serverDefault,
+        }), { status: 200, headers: { 'Content-Type': 'application/json' } })
+      }
+      return jsonResponse([session('pi-1')])
+    })
+
+    render(
+      <PiChatPanel
+        storageScope="workspace-a"
+        storage={persisted}
+        fetch={fetchMock as unknown as typeof fetch}
+        createRemoteSession={remoteFactory(remote)}
+      />,
+    )
+
+    const textarea = await screen.findByLabelText('Agent prompt')
+    await waitFor(() => expect((textarea as HTMLTextAreaElement).disabled).toBe(false))
+    expect((screen.getByRole('button', { name: 'Submit' }) as HTMLButtonElement).disabled).toBe(false)
+
+    fireEvent.change(textarea, { target: { value: 'pre-seeded prompt' } })
+    fireEvent.click(screen.getByRole('button', { name: 'Submit' }))
+    await waitFor(() => expect(remote.prompt).toHaveBeenCalledWith(expect.objectContaining({
+      message: 'pre-seeded prompt',
+      model: serverDefault,
+    })))
+  })
+
   test('opens model and thinking pickers from slash commands', async () => {
     const remote = new FakeRemotePiSession(remoteState({ committedMessages: [], history: { mode: 'full', messageCount: 0 } }))
     const fetchMock = vi.fn().mockResolvedValue(jsonResponse([session('pi-1')]))
--- a/packages/agent/src/front/hooks/__tests__/useChatModelSelection.test.tsx
+++ b/packages/agent/src/front/hooks/__tests__/useChatModelSelection.test.tsx
@@ -147,6 +147,37 @@ describe('useChatModelSelection', () => {
     expect(result.current.model).toBeNull()
   })
 
+  it('resolves the agent default model for a pre-seeded/directly-created session with no stored model (#1469)', async () => {
+    // Sessions created outside the primary new-chat flow (pre-seeded scripted
+    // sessions, sessions created via a direct POST .../sessions call) hydrate
+    // with no session.currentModel and are never "new" (sessionIsNew is false
+    // once the session already carries history/queue state or was created
+    // out of band). Composer authority must still resolve a model from the
+    // agent's server-side default so the composer isn't disabled forever.
+    const defaultModel = { provider: 'anthropic', id: 'claude-sonnet' } as const
+    const fetchImpl = vi.fn(async () => new Response(JSON.stringify({
+      models: [
+        { provider: 'anthropic', id: 'claude-sonnet', label: 'Sonnet', available: true },
+      ],
+      defaultModel,
+    }))) as unknown as typeof fetch
+
+    const store = storage()
+    const { result } = renderHook(() => useChatModelSelection({
+      sessionId: 'pre-seeded-no-model',
+      sessionHydrated: true,
+      sessionIsNew: false,
+      storageScope: 'scope-a',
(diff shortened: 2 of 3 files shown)


hachej:
REVISE

Blocking behavior/test-honesty finding: packages/agent/src/front/hooks/useChatModelSelection.ts currently computes selected as pendingOverride ?? sessionModel ?? current.localDefault and returns current whenever that selection is discovered and available. For a hydrated, non-new existing session with no stored session model, an available browser-stored model therefore survives discovery. sessionAuthorityReady then publishes that browser choice instead of resolving the server defaultModel/first available model. This violates the stated #1469 behavior and can submit with the wrong model in both direct and remote deployments; the frontend path is shared, and remote filtering makes trusting an old browser choice especially unsafe.

The new test false-passes this branch because storage is empty and discovery exposes only the default model. Add a regression with available browser model A plus distinct server default B for a non-new/no-sessionModel session, and assert B is selected and submitted. A PiChatPanel-level case should prove the composer unlocks and the prompt payload carries B.

#1319 regression review: the implementation still gates readiness on sessionHydrated, canonical hydration assigns currentModel and hydrated:true atomically, and sessionModel outranks localDefault. Therefore discovery completing before hydration cannot expose/send the default when hydration later supplies a stored model. The merged #1319 hook and PiChatPanel tests passed, but they run with server discovery disabled/host-controlled models; add the deferred ordering case to lock this down: discovery completes while unhydrated => model null/no send, then atomic hydration with stored model => display and submit exactly that model.

Proof run at c448118ee8da995e4a50be44d7e05a8272ce9d04:
- useChatModelSelection: 13/13 passed, no type errors
- PiChatPanel: 75/75 passed, no type errors
- packages/agent typecheck: passed

hachej:
Addressed both findings at 099f568f3.

**1. Blocking finding (real defect, fixed).** Confirmed: the discovery-success branch checked `availableModel(models, selected)` before resolving a non-new/no-`sessionModel` session's default, so a leftover browser-stored model that happened to still be "available" per discovery survived instead of the server's own `defaultModel`. Fixed in `useChatModelSelection.ts` — for `sessionId && !sessionIsNew` with no `sessionModel`, we now always resolve exclusively from `payload.defaultModel`/first-available, regardless of whether the stale local default was independently available.
   - New test `useChatModelSelection.test.tsx`: browser storage has model A (available), discovery's `defaultModel` is distinct model B → asserts `model` resolves to B, never A. Verified failing on pre-fix code (`expected {id:'claude-opus'} to deeply equal {id:'claude-haiku'}`), passing after.
   - New test `PiChatPanel.test.tsx`: pre-seeded session (no stored model), *live* server model discovery via `fetch` (not the `availableModels`/`serverResourcesEnabled=false` shortcut every other test in the file uses), stale browser default present. Asserts the composer textarea/submit unlock and `remote.prompt` is called with the server default model. Verified failing on pre-fix code (composer stayed disabled — timed out on `textarea.disabled === false`), passing after.

**2. #1319 ordering finding: reviewed, confirmed not a live bug — rebutted with evidence, plus a regression lock added.** `sessionAuthorityReady` is `sessionHydrated && (sessionModel !== undefined || sessionIsNew || discoveryVetted)` — `sessionHydrated` gates all three disjuncts, so discovery completing before hydration cannot expose or submit a default for a session that will later report a stored model; canonical hydration assigns `currentModel` and `hydrated: true` atomically (there's no partially-hydrated state to race). Added `useChatModelSelection.test.tsx`: discovery resolves while `sessionHydrated: false` → asserts `model` is `null` (nothing to submit); then hydration lands atomically with a distinct stored model → asserts `model` equals exactly that stored model, never the discovery default that was already loaded. This test passes both before and after the review-fix commit (it was never broken) — it's a lock, not a repro.

**Direct/remote mode:** the hook and `PiChatPanel` gating are shared across both — no mode-specific branch exists in this code path, so the fix and both new tests cover both by construction. Not treated as a separate finding requiring a distinct test.

Proof run at 099f568f3:
- `useChatModelSelection.test.tsx`: 15/15 passed
- `PiChatPanel.test.tsx`: 76/76 passed
- full `src/front` suite: 665/665 passed
- `packages/agent` typecheck: passed

hachej:
SHIP

PR #1583, opened 2026-09-15
Add application-owned AgentCore shared-EFS runtime fixture

hachej:
Closes factory-plugin-owner-runtime-identity-prefix-8pdn.5

Adds host-owned ECS-local and AgentCore remote shared-EFS adapters, fenced handle lifecycle integration, local conformance tests, and an explicitly live-unverified AWS qualification runbook.

Proof: `pnpm --filter factory-playground exec vitest run src/server/agentcoreEfsRuntime.test.ts` (4 passed). Changed-file typecheck is clean; whole playground typecheck remains blocked by pre-existing unbuilt workspace package declarations.


hachej:
Proof update for delayed-delete/successor-publication race gap:

- Added deterministic real-Postgres regression in packages/core/src/server/runtime/__tests__/PostgresFencedSandboxHandleStore.test.ts: owner A holds a pending-validation unpublished handle, provider delete is deferred past lease expiry, owner B takes over, B cannot publish/resume A's pending handle, A's delayed delete is fenced false after takeover, and B can create/publish a successor without stale deletion tombstoning it.
- Fixed PostgresFencedSandboxHandleStore.claim to carry forward only published handles on expired-lease takeover; abandoned pending-validation handles are cleared for the successor while malformed encrypted rows without publication state still fail.

Validation:
- pnpm --filter @hachej/boring-core exec vitest run src/server/runtime/__tests__/PostgresFencedSandboxHandleStore.test.ts --no-file-parallelism — passed (8 tests)
- pnpm lint:invariants — passed (rerun after initial 180s timeout)
- pnpm check:core-bundle-size — passed

Pushed commit: abf56641f

